PR #1, opened 2017-11-17
Docker infrastructure.

clint-tseng:
wip


config/nginx/nginx.conf
```diff
@@ -0,0 +1,26 @@
...
+  listen 443 ssl;
+  ssl_certificate /etc/nginx/external/cert.pem;
+  ssl_certificate_key /etc/nginx/external/key.pem;
+
+  ssl_protocols TLSv1.2 TLSv1.1 TLSv1;
+  ssl_prefer_server_ciphers on;
+  ssl_ciphers "EECDH+ECDSA+AESGCM EECDH+aRSA+AESGCM EECDH+ECDSA+SHA384 EECDH+ECDSA+SHA256 EECDH+aRSA+SHA384 EECDH+aRSA+SHA256 EECDH+aRSA+RC4 EECDH EDH+aRSA RC4 !aNULL !eNULL !LOW !3DES !MD5 !EXP !PSK !SRP !DSS";
+  add_header Strict-Transport-Security "max-age=31536000";
+
+  location /api {
```
matthew-white: What do you think of a prefix of `/api/v1` in case we ever want to make big changes to the API? That would allow us to use a new set of routes under `/api/v2` without breaking existing scripts.

clint-tseng: Yes, though one wonders if that should maybe be at the service level rather than the url rewrite level. Depends on whether we think it's more likely we'd spin up an old binary for backwards compat or whether we'd have all the routes at once.

Either way, for now we can just rename it wherever and it'll transparently change later.

matthew-white: I think maybe it would be better to implement this at the service level. How about I go ahead and create a Jubilant Garbanzo issue along those lines?

clint-tseng: Sure!

matthew-white: Which sounds better to you? I'm happy with either option. :)

clint-tseng: I'll make a garbanzo issue

PR #10, opened 2018-06-06
dist: update client and server to latest masters

yanokwa:
I *think* this is how to update the submodules, but I'm probably wrong. The command I used was: 
`git submodule update --remote`


clint-tseng:
That works. The other way to do it is to just go into each directory and `git checkout` the exact SHA you want. Go back up and commit the change. Simple and powerful.

PR #12, opened 2018-06-27
Add GitHub Pages files for use for latest news panel of site root page

matthew-white:


lognaturel:
Looks like a fine start. Will be easier to see whether styling and structure need to be rethought once the embed is working.

matthew-white:
By the way, @lognaturel, I thought these files probably didn't need to include license/copyright information, so they don't have those headers. Let me know if I should add them!

lognaturel:
Nah, I think it's fine to exclude them.

PR #15, opened 2018-08-14
Nginx and node improvements

clint-tseng:
1. Enable gzip on nginx.
2. Always rebuild the nginx conf file so it picks up conf changes.
3. Start using multithreaded node.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -15,6 +15,12 @@ server {
 
   client_max_body_size 100m;
 
+  gzip on;
+  gzip_vary on;
+  gzip_min_length 1280;
+  gzip_http_version 1.1;
+  gzip_types text/plain text/html text/css application/json application/x-javascript text/xml;
+
   location ~ ^/v\d {
     proxy_set_header X-Forwarded-Proto $scheme;
     proxy_pass http://service:8383;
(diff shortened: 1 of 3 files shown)


files/nginx/odk.conf.template
```diff
@@ -15,6 +15,12 @@ server {
 
   client_max_body_size 100m;
 
+  gzip on;
+  gzip_vary on;
+  gzip_min_length 1280;
```
yanokwa: I've seen 1000 on the interwebs. Why did you choose 1280 bytes?

files/service/scripts/start-odk.sh
```diff
@@ -12,5 +12,6 @@ echo "starting cron.."
 cron -f &
 
 echo "starting server."
-node lib/bin/run-server.js
+mkdir -p /var/log/odk
+node node_modules/naught/lib/main.js start --remove-old-ipc true --worker-count 4 --daemon-mode false --log /var/log/odk/naught.log --stdout /var/log/odk/stdout.log --stderr /var/log/odk/stderr.log lib/bin/run-server.js
```
yanokwa: Why 4 workers?

files/nginx/odk.conf.template
clint-tseng: _shrug_. It's not a terribly sensitive value, and the thing I copied had 1280 and 2560 as suggestions. I'd comment on it, but the comment would simply read "why not?" Honestly some of this stuff is as much an incantation to me as it is to you. There is definite cargo-culting in this block of config.

yanokwa:
Thanks for the clarifications. I just like to have some record of where the magic numbers come from.

Merge whenever you are ready!

PR #30, opened 2018-11-02
Update server and client to latest versions

matthew-white:


yanokwa:
The diff helps. I can confirm that these are the tagged versions on the various repos.

PR #31, opened 2018-11-02
Update news for v0.3.0-beta.0

matthew-white:


lognaturel:
Perhaps we should move away from US-style dates and use month names in a future release? I had to pause and think about it because when I see the year first I usually expect international-style dates (that is, 2018/02/11).

clint-tseng:
International _is_ 2018/11/02 is it not?

lognaturel:
It's the ISO format so I guess international for computers but I don't think it is for humans. Europe and I think all the other continents use dd/mm and I believe that includes when the year is first.

Regardless, it's not terribly important exactly what the date is, but wanted to flag that it gave me pause. Maybe I had just been speaking French or something.

clint-tseng:
my understanding is that it's a little bit of a mess but very rarely does any country use yyyy/dd/mm. it's almost always either yyyy/mm/dd or dd/mm/yyyy.

okay, i looked it up. looks like YDM is a kazakh and latvian thing. not really anywhere else.

lognaturel:
Huh. Well, then, I stand corrected. 🤷🏻‍♀️

PR #50, opened 2019-02-12
make nginx decompress gzipped upload requests

clint-tseng:


lognaturel:
This was introduced so that Frontend can compress CSV form attachments.

Issue #22, opened 2018-09-10
Alternative deployments with Docker (Traefik.io)


poVoq:
I am trying to understand (sorry not a docker expert) how I can make this work with my Traefik.io based host (also using portainer so docker compose isn't preferred).

Looking at the docker-compose file the postgres database is easy to replicate, and I think the actual ODK-central service should also work. But from then on I am a bit lost...

However what is this "mail" service needed for (seems to be an smtp server?) and how can I avoid running an instance of Nginx? 

Traefik.io has me pretty much covered for all the reverse-proxy/https needs, and NodeJS usually comes with a build-in webserver, no? So which port is exposed in the "service" container that I could point Traefik.io to?

Thanks for the help!

poVoq:
Releated and answers some questions already:
https://github.com/opendatakit/central/issues/17

clint-tseng:
We are a small team and the project is young; our official support is for the whole stack as-is, with docker-compose managing the entire infrastructure. We cannot guarantee that these things will work well now or later once you break out of that packaging, nor that we won't make some change that might break your custom setup.

That said, I'm happy to offer answers where I can.

As you saw on the other thread, mail is just an SMTP relay. Any SMTP relay will work if you just point the `service` configuration at it by IP/DNS. Same goes for postgres, though the more you stray on postgres the less likely the built-in backup solution is to work for you, and the more likely you'll have to build your own.

The nginx instance is in charge of serving the static html/js/css/etc assets that comprise the frontend. So you can avoid running it somewhat, but you'll need to supply some alternative way to serve those files. The backend service does not do this.

poVoq:
Is it possible to pass the postgres service address:port also via an ENV variable instead of the command in the composer file? As stated above I am trying to integrate this in an existing infrastructure and this would be helpful in doing so.

clint-tseng:
hm. i am wary of passing more and more things through env vars, since eventually the whole thing will be a mess of env vars and it'll be really difficult to eg update config structures, etc, since everyone will just have a pile of customized variables. what's the issue with modifying the config file?

as for `start-odk.sh`, is it the `wait-for-it` that is the problem, or the location of the command within the compose file?

poVoq:
It's a defacto standard for docker containers to be configured through env variables and that they include autostart scripts to be fullyselfcontained.

Your docker compose setup works fine if you spin it up on a dedicated machine, but then the question comes up why use Docker in the first place?

[...]

PR #51, opened 2019-02-12
Added support for traefik

vertoforce:


vertoforce:
Apologies opened this by accident.  However if anyone wants to use traefik with ODK central, this is a setup that works with traefik.

clint-tseng:
hey @vertoforce, no worries about the accident. there is actually an open ticket (#22) asking about traefik-based deployment, so it seems like this is something that people do want.

i'm myself unfamiliar with traefik, and not especially experienced with docker, so if it _is_ of interest to you to help maintain such an alternative configuration, i'd be very happy to work with you on that. otherwise, at the very least you've contributed a much appreciated starting point for others.

thanks!

vertoforce:
Yeah sure thing I was setting it up for my own interest anyway.
Traefik is basically just a nice load balancer frontend with letsencrypt support as well.  However I was getting some bugs when nginx in ODK central was using https, so I set it up to use just http, then linked traefik to it.

PR #60, opened 2019-02-22
bug: sending SIGHUP too quickly fails nginx out.

clint-tseng:
* i don't know why.

### DIFF
--- a/files/nginx/run_certbot.sh
+++ b/files/nginx/run_certbot.sh
@@ -1,0 +1,37 @@
+#!/bin/sh
+
+# Source in util.sh so we can have our nice tools
+. $(cd $(dirname $0); pwd)/util.sh
+
+# We require an email to register the ssl certificate for
+if [ -z "$CERTBOT_EMAIL" ]; then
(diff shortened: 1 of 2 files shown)


clint-tseng:
the `run_certbot.sh` file is copied exactly from the source repository, but with the addition of the `sleep 5` line.

i don't know why this is failing now, i don't know why it doesn't work without the sleep, and i don't know why it works with it other than that delaying SIGHUP by as little as a half second seems to resolve the issue.

i like nothing about this. but at least this works.

files/nginx/run_certbot.sh
```diff
@@ -0,0 +1,37 @@
...
+
+exit_code=0
+set -x
+# Loop over every domain we can find
+for domain in $(parse_domains); do
+    if is_renewal_required $domain; then
+        # Renewal required for this doman.
+        # Last one happened over a week ago (or never)
+        if ! get_certificate $domain $CERTBOT_EMAIL; then
+            error "Cerbot failed for $domain. Check the logs for details."
```
yanokwa: Probably a typo that should be fixed upstream.

PR #84, opened 2019-08-01
repo: remove dev installation instructions.

issa-tseng:


lognaturel:
I think it'd be helpful to have some kind of guidance for someone looking to set up a dev box. They should just set up frontend and backend separately, right?

CC @smoyth

smoyth:
Yes I still wouldn't know how to set up a dev environment based on this readme. I've been told to use the production instructions but it's not clear from this readme that that's what one is supposed to do. I'm also curious now based on what @lognaturel has said if that is in fact the correct course of action.

issa-tseng:
sorry, yes, i am 10000% slammed right now and not able to write new instructions. i wanted mostly to eliminate the confusing ones based on the conversation we had. for now @smoyth please check out the dev instructions for the central-frontend repo and they should help you get there.

smoyth:
I certainly know that feeling! Thanks.

lognaturel:
Makes sense, thanks, @issa-tseng! I should have noted I knew you were slammed and wasn't expecting an answer right away. Let's merge this and follow up later!

PR #85, opened 2019-08-03
operations updates for release v0.6

issa-tseng:


issa-tseng:
okay, this is ready for review now.

service.dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM node:8.11.3
+FROM node:12.6.0
```
yanokwa: Just so we have a record. Why not 12.7?

nginx.dockerfile
```diff
@@ -5,7 +5,7 @@ RUN files/prebuild/write-version.sh
 RUN files/prebuild/build-frontend.sh
 
 
-FROM staticfloat/nginx-certbot:latest
+FROM staticfloat/nginx-certbot@sha256:a3fff8a1d75ae2b28d8e77d71468bc51bf98588c4762c6c7d7c55e0e548e6976
```
yanokwa: Is there an easy way to find what commit on Github this is?

service.dockerfile
issa-tseng: because node releases versions too quickly and i didn't know there was a 12.7. i looked at the changelog and it doesn't look too significant.

nginx.dockerfile
issa-tseng: nope. it's not a commit on github, it's the sha of the docker image, and it is notoriously hard to get lists of these shas or even the current one you are using. apparently this pain point is A Thing in the docker community.

service.dockerfile
```diff
@@ -1,11 +1,11 @@
...
 
 WORKDIR /usr/odk
 
-RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ jessie-pgdg main" > /etc/apt/sources.list.d/pgdg.list; \
```
matthew-white: Is the move from jessie to stretch related to the node upgrade?

service.dockerfile
```diff
@@ -1,11 +1,11 @@
...
-  wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | \
-  apt-key add -; \
-  apt-get update; apt-get install -y cron gettext postgresql-client-9.6
+RUN RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main" | tee /etc/apt/sources.list.d/pgdg.list; \
```
matthew-white: I'm probably missing something obvious, but where does the extra `RUN` come from?

service.dockerfile
issa-tseng: yes. i don't get a choice in the matter

service.dockerfile
issa-tseng: uhhh wow. how does this work.

PR #90, opened 2019-08-08
add a ci smoke test build

### LINKED ISSUE #88: Ensure master always builds cleanly
It'd be nice to run docker-compose on master to ensure the build works. 

Issa sez:
> if that is possible then yes. first we should have some process so master stays clean
> 
> i would have a release branch or something but i feel like it adds a step people could get wrong when cloning and people already have clones on master
> 
> so possibly we should merge to a prerelease and then merge that to master. though i have also had headaches in the past with long-running staging branches accumulating unintended deltas against master due to merge conflicts
> 
> https://circleci.com/docs/2.0/docker-compose/ looks like the CI thing may be possible

issa-tseng:
resolves #88

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,35 @@
+version: 2
+jobs:
+  build:
+    machine:
+      image: ubuntu-1604:201903-01
+      docker_layer_caching: true
+
+    steps:
+      - checkout
+
+      - run: git submodule update -i
+
+      - run: |
+          echo 'SSL_TYPE=selfsign
+          DOMAIN=local
+          SYSADMIN_EMAIL=no-reply@opendatakit.org' > .env
+
+      - run: docker-compose build
+
+      - run:
+          # we allow a long retry period for the first check because the first-run
+          # nginx setup could take several minutes due to key generation.
+          name: Verify frontend and backend load
+          command: |
+            set -x
+            docker-compose up -d
+            docker run --network container:nginx \
+              appropriate/curl -4 --insecure --retry 30 --retry-delay 10 --retry-connrefused https://localhost/ \
+              | tee /dev/tty \
+              | grep -q 'ODK Central'
+            docker run --network container:nginx \
+              appropriate/curl -4 --insecure --retry 20 --retry-delay 2 --retry-connrefused https://localhost/v1/projects \
+              | tee /dev/tty \
+              | grep -q '\[\]'
+
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,7 @@ services:
       POSTGRES_PASSWORD: odk
       POSTGRES_DATABASE: odk
   mail:
+    container_name: mail
     image: "itsissa/namshi-smtp:4.89-2.deb9u5"
     volumes:
       - ./files/dkim/config:/etc/exim4/_docker_additional_macros:ro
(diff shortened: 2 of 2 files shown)


issa-tseng:
this is now ready. i recommend not merging it to master until after v0.6 is out, and we try this branch out on some already-deployed systems to be sure it doesn't break anything, since i did touch the main docker-compose configuration a little bit.

it builds the entire platform and verifies that the frontend root returns the copyright text `Copyright 2017 ODK Central Developers` and that the backend projects resource returns `[]`.

issa-tseng:
okay, i ran this branch on antinodes and there was no data loss. would love to see a couple more attempts.

.circleci/config.yml
```diff
@@ -0,0 +1,33 @@
+version: 2
+jobs:
+  build:
+    machine:
+      image: ubuntu-1604:201903-01
+      docker_layer_caching: true
```
yanokwa: I think this might only work on premium CI plans, but let's keep it and see what happens after merge. We might be able to get a free upgrade.

docker-compose.yml
```diff
@@ -9,13 +9,15 @@ services:
       POSTGRES_PASSWORD: odk
       POSTGRES_DATABASE: odk
   mail:
+    container_name: mail
```
yanokwa: I'm guessing you added these names so you can refer to them with the docker run command?

.circleci/config.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - run:
+          name: Verify frontend and backend load
+          command: |
+            set -x
+            docker-compose up -d
+            docker run --network container:nginx \
+              appropriate/curl -4 --insecure --retry 30 --retry-delay 10 --retry-connrefused https://localhost/ \
+              | tee /dev/tty \
+              | grep -q 'Copyright 2017 ODK Central Developers'
```
yanokwa: `curl -4 --insecure --retry 30 --retry-delay 10 --retry-connrefused https://sandbox.central.opendatakit.org` doesn't return text that contains `Copyright 2017 ODK Central Developers`. Should it?

.circleci/config.yml
```diff
@@ -0,0 +1,33 @@
...
+            docker run --network container:nginx \
+              appropriate/curl -4 --insecure --retry 20 --retry-delay 2 --retry-connrefused https://localhost/v1/projects \
```
yanokwa: Curious about the difference in the retry and retry-delay in this vs the previous command. If it's done on purpose, it'd be nice to have a comment or something.

.circleci/config.yml
issa-tseng: i'm not fully convinced it works anyway

docker-compose.yml
issa-tseng: yeah. without explicit names they get semirandomly generated and it's hard to address them

.circleci/config.yml
issa-tseng: well, it used to. i'll adjust the search text.

.circleci/config.yml
issa-tseng: it is done on purpose. the initial bootup time is incredibly long because keys must be generated, etc. but once the first curl passes we expect the second to be ready right away.

how does one leave a comment in this sort of thing?

yanokwa: You can add a ` # comment` above command or run. If that fails, add to commit message. If that fails, then we have this convo.

PR #95, opened 2019-09-26
bug: set nginx proxy buffering settings to enable response streaming.

issa-tseng:


issa-tseng:
(the increase to a 2 minute timeout is probably not strictly necessary. it's a per-chunk timeout, so it will only trip if no new data is streamed back to the client within that window. i artificially inflated row-processing time by 50ms per row and the first chunk came back at 58 seconds so i figured 2 minutes wasn't a terrible idea for some diabolically complex forms, possibly. i don't see the harm in increasing by 60 seconds.)

PR #99, opened 2019-10-05
nginx: unzip uploads that were large enough to spool to disk.

issa-tseng:
* there is a filesize threshold above which nginx will still accept the
  file, but it will spool it to disk rather than store it in memory.
* in this case, the lua interface changes and the file must be read.
* so we do that here.
* a lua expert should probably review this...

### DIFF
--- a/files/nginx/inflate_body.lua
+++ b/files/nginx/inflate_body.lua
@@ -52,6 +52,12 @@ if content_encoding == "gzip" then
   ngx.req.read_body()
   local data = ngx.req.get_body_data()
 
+  if data == nil or data == '' then
+    local file = io.open(ngx.req.get_body_file(), "r")
+    data = file:read("*a")
+    file:close()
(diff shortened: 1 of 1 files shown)


lognaturel:
The logic appears sound and it doesn't seem like there would be many different ways to do this but we have confirmed in the identification of this very issue that I'm not very creative when it comes to imagining What Could Go Wrong. How about contacting the original author of the snippet for a quick review?

lognaturel:
In fact, I will attempt it! Dear @davidcaste, would you be willing to review this small addition to your very helpful post at http://www.pataliebre.net/howto-make-nginx-decompress-a-gzipped-request.html? Thank you so much! 🙏

issa-tseng:
yeah. it would. if people have problems we can then invent a further solution, i think. possibly handle it as middleware in nodejs directly.

PR #109, opened 2019-11-08
new: add xlsform conversion service; configure backend for it.

issa-tseng:


docker-compose.yml
```diff
@@ -45,6 +45,9 @@ services:
       - "443:443"
     healthcheck:
       test: [ "CMD-SHELL", "nc -z localhost 443 || exit 1" ]
+  xlsform:
+    container_name: xlsform
+    image: 'yanokwa/pyxform-http:v0.3.0'
```
yanokwa: I'm assuming this is trivial to change if we move the repo.

issa-tseng: yes

files/service/config.json.template
```diff
@@ -14,6 +14,10 @@
         "port": 25
       }
     },
+    "xlsform": {
```
yanokwa: I think you missed this one?

issa-tseng: nope. that is the key central-backend wants to read, and i feel like in that case it's fine for it to generically be "xlsform"

PR #119, opened 2019-12-04
Measure usage as proxy for number of deployments

yanokwa:
Analytics code is what Google provided. The UA key is owned by the Open Data Kit account and @issa-tseng has read access.

### DIFF
--- a/docs/news.html
+++ b/docs/news.html
@@ -4,6 +4,14 @@
     <meta charset="utf-8">
     <title>ODK Central</title>
     <link rel="stylesheet" href="style.css">
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-91951913-7"></script>
+    <script>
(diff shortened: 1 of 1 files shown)


issa-tseng:
does this have gdpr implications?

yanokwa:
@issa-tseng I don't know. What we do now across all the tools is to document what data we collect at https://docs.opendatakit.org/security-privacy and I'll add some language there about Central if this PR is approved. 

I've filed an issue at https://github.com/opendatakit/governance/issues/60 so can review our usage across the project and make the necessary changes. My bias is to merge this because this is no worse than current practice.

CC @lognaturel @matthew-white

issa-tseng:
well, i think if we do fall under gdpr then this is not a compliant solution, but it's not my problem to solve :)

PR #140, opened 2020-06-20
enketo: incorporate enketo into the docker setup.

issa-tseng:


yanokwa:
According to https://enketo.github.io/enketo-express/tutorial-10-configure.html, the support email is important to set. 

https://github.com/enketo/enketo-express/search?q=supportEmail&unscoped_q=supportEmail is how it's used. I'm not sure what the right approach here is. Maybe we explicitly set it to an empty string?

issa-tseng:
i can reuse the letsencrypt contact email?

yanokwa:
Martijn says, "I think the survey administrator is the one to use." @lognaturel agrees and thinks this is something that should be configurable at the project level (for Enketo and Collect).

For now, the LetsEncrypt email is sufficient, but we should plan on having configurable at the project level.

files/enketo/config.json.template
```diff
@@ -0,0 +1,32 @@
+{
+    "app name": "Enketo",
+    "base path": "enketo",
+    "encryption key": "${SECRET}",
+    "id length": 31,
```
lognaturel, line 5: Are we sure we want such a long default? It's on par with e.g. Google Drive links so maybe it will be somewhat familiar but I think we'll get pushback and requests to configure it pretty quickly. The Enketo default is 6. Given that we have the cookie auth option and it's only public surveys that will rely on this being hard to fake, could we make it shorter?

yanokwa: Agree with @lognaturel. I think the default is OK.

lognaturel: Also, it's lowercase, uppercase, digits so there are a lot of combinations even for 6.

issa-tseng: i dunno. @jnm set it to 31 so i just went with it.

issa-tseng:
> Martijn says, "I think the survey administrator is the one to use." @lognaturel agrees and thinks this is something that should be configurable at the project level (for Enketo and Collect).
> 
> For now, the LetsEncrypt email is sufficient, but we should plan on having configurable at the project level.

okay but that would be an _enketo_ change then, no? if it's configured on a server-by-server basis?

yanokwa:
You are right. If it's project by project, there would need to be an Enketo change to have a support email.

files/enketo/config.json.template
jnm: i'm sorry, i've totally fallen off of the :bus: with this, and i haven't caught up with how cookie auth is going to work. i set this to the maximum following the example of `fieldKeys` and long, random URLs being used as authentication with Collect—figuring that someday Enketo URLs could be distributed via QR code too, and that the URLs would be the only form of authentication. KoBo uses the default 6-character code but also prompts for a username and password when submitting (if the form administrator chooses to require that).

jnm: i caught up with the [slack discussion](https://getodk.slack.com/archives/CB2UJJ6SF/p1589667171115800?thread_ts=1589336453.045900&cid=CB2UJJ6SF) and see that you all have a long-term plan for a form-listing landing page that would continue to use cookie auth, so i understand now why only public forms would ever rely on the difficulty of guessing this identifier.

thanks @issa-tseng for seeing all this through.

issa-tseng: okay right so because draft forms via enketo are still only protected by the strength of the token i think we need to leave this long. draft forms do not ues cookie auth.

issa-tseng: hey @jnm while you're here does this config look okay for our purposes if you have time? i just went mostly with what you had on that kbt dev server.

jnm: i just checked against what KoBo uses in production and the only differences i see are that KoBo:
* adds some [extra map layers](https://github.com/kobotoolbox/kobo-docker/blob/1ea61dc6a0a536e9f95d8da0426e1bc5413a7e97/deployments/enketo_express/config.json#L59-L78)
* bumps of `payload limit` from its default of `100kb` up to `1mb`: i think this solved a problem where submissions whose XML was larger than 100kb could not be edited
* changes `text field character limit`, which was unlimited in older versions of Enketo, from its default of 2000 to 1,000,000: we had multiple complaints about the 2000-character ceiling when we deployed the first version with that limitation

not that any of these are necessarily desirable for your configuration, but at least you know ;)

PR #141, opened 2020-06-23
Add sponsor button

yanokwa:
Add sponsor button that links to https://github.com/sponsors/getodk

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+github: getodk


matthew-white:
Would it be useful to add something like this to the other repos as well? I feel like Backend especially will see independent usage.

yanokwa:
We'll eventually add to all repos, but since this is "the" Central repo, I think we should start here. As of today, it's only Collect and Central that have the sponsor badges.

PR #143, opened 2020-06-28
Support auto-restarting docker-compose stack on system reboot

ghostfreak3000:
The docs currently state in order to support auto-restart on system reboot, a systemd script must be installed.
This might ( and has in my case ) caused permission issues in the event the user running the docker commands isn't a root user

We can easily support auto-restart by adding the restart: always entry in the docker-compose service and enabling the docker service for systemd

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,7 @@ services:
       POSTGRES_USER: odk
       POSTGRES_PASSWORD: odk
       POSTGRES_DATABASE: odk
+    restart: always
   mail:
     container_name: mail
     image: "itsissa/namshi-smtp:4.89-2.deb9u5"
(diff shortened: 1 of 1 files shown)


issa-tseng:
oh, neat. i didn't know you could just do it that way.

i've done restart unless stopped instead; is there a strong reason to use always?

ghostfreak3000:
@issa-tseng 
For this usecase, unless_stopped and always would achieve the same result.

Though, the docker docs state that they wouldn't differentiate between a manual stop and an automated one. 
So a very small concern of mine would be if some rogue process somehow found a way to stop the containers ( but by the time this happens, you have a bigger problem on your hands than just being able to reboot the system )

issa-tseng:
okay, i've incorporated your changes (with you tagged as the author) into #140 which should be merged soon. and i've opened a PR against the documentation to adapt for this change as well. for people who already have the systemd configuration in place i think we'll just have them leave it there; it doesn't seem to cause any adverse effects.

thanks for the suggestion here!!

PR #145, opened 2020-07-13
Update submodules for v0.9

matthew-white:


yanokwa:
Confirmed that the hashes match the latest in backend/frontend repo

PR #153, opened 2020-07-28
enketo: change the base url/path prefix from enketo to _

issa-tseng:
* shorter, removes noncritical detail.
* depends on a corresponding frontend pr.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -21,7 +21,7 @@ server {
   gzip_http_version 1.1;
   gzip_types text/plain text/css application/json application/x-javascript text/xml text/csv;
 
-  location /enketo {
+  location /_ {
     proxy_pass http://enketo:8005/enketo;
     proxy_redirect off;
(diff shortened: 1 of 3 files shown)


files/nginx/odk.conf.template
```diff
@@ -21,7 +21,7 @@ server {
   gzip_http_version 1.1;
   gzip_types text/plain text/css application/json application/x-javascript text/xml text/csv;
 
-  location /enketo {
+  location /_ {
     proxy_pass http://enketo:8005/enketo;
```
lognaturel, line 25: I'm surprised to see everything else change to `_` but this `proxy_pass` stay the same. Wouldn't the Enketo base path have to remain `enketo` for this to work? (This is not super familiar to me and I think you've tried it so I'm probably misunderstanding.)

issa-tseng:
nope i did that wrong

lognaturel:
We’ll need another update one Enketo gets a new patch.

PR #163, opened 2020-08-28
enketo: update to 2.3.12 for lower memory reqs on build

lognaturel:
I've verified the diff from 2.3.11 to 2.3.12 and have confirmed there's nothing risky. I'm not sure the best way to try building this on a 1gb DO droplet and am hoping @yanokwa or @matthew-white can quickly do that.

### DIFF
--- a/enketo.dockerfile
+++ b/enketo.dockerfile
@@ -1,4 +1,4 @@
-FROM enketo/enketo-express:2.3.11
+FROM enketo/enketo-express:2.3.12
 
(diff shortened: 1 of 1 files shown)


seadowg:
I tried this on a 1GB Digital Ocean droplet and it was happy. Just for the record I checked out @lognaturel's branch and then did:

```
docker system prune
docker-compose build
docker-compose up -d
```

yanokwa (approved):
LGTM

PR #204, opened 2021-03-21
Update readme

matthew-white:
The main change is updating the section about Enketo now that Frontend can use a `Secure` cookie in development (getodk/central-frontend#418). The changes here follow those in getodk/central-frontend#421.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,19 @@ Our goal with Central is to create a modern server that is easy to install, easy
 This repository serves as an umbrella for the Central project as a whole:
 
 * Operations repository for packaging the server and client into a Docker Compose application.
-* Release repository for publishing binary artifacts.
+* Release repository for publishing binary artifacts. Release notes can be found in this repository.
 
 If you are looking for help, please take a look at the [Documentation Website](https://docs.getodk.org/central-intro/). If that doesn't solve your problem, please head over to the [ODK Forum](https://forum.getodk.org) and do a search to see if anybody else has had the same problem. If you've identified a new problem or have a feature request, please post on the forum. We prefer forum posts to Github issues because more of the community is on the forum.
 
 Contributing
 ============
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,16 +8,19 @@ Our goal with Central is to create a modern server that is easy to install, easy
 This repository serves as an umbrella for the Central project as a whole:
 
 * Operations repository for packaging the server and client into a Docker Compose application.
-* Release repository for publishing binary artifacts.
+* Release repository for publishing binary artifacts. Release notes can be found in this repository.
```
lognaturel, line 11: Link to https://github.com/getodk/central/releases?

lognaturel (approved):
Would be good to actually write up the instructions for doing dev with Enketo at some point.

I think you've seen @matthew-white that @ktuite wrote up [some instructions](https://docs.google.com/document/d/1L_o7n3lhPHAgHp2U73IpmOgiIkoOkpVCS8K_BzkRZ0c/edit#heading=h.cop79p36pjxj) for getting data into Central. I don't think it goes in any of these readmes because it's across many tools but perhaps we should think about adding to the development section of docs.getodk.org and linking to it.

matthew-white:
> Would be good to actually write up the instructions for doing dev with Enketo at some point.

I agree! I think development that touches Enketo would be much easier with Enketo set up locally. Having Enketo set up locally would also make it easier to create test submissions. I think setting up Enketo locally would require changes to `nginx.conf` (Frontend) and `config/default.json` (Backend).

> I think you've seen @matthew-white that @ktuite wrote up some instructions for getting data into Central. I don't think it goes in any of these readmes because it's across many tools but perhaps we should think about adding to the development section of docs.getodk.org and linking to it.

Ah very cool!  One possibility that the instructions mention is adding the file to `central-backend`. I could see that being a reasonable home, since a lot of the instructions have to do with the database and the API. `central-frontend` could then maybe link to the file in `central-backend`.

lognaturel:
> One possibility that the instructions mention is adding the file to central-backend

Could certainly start with that and maybe it's fine long-term too. It's useful not just for Central dev but also for Enketo dev, Collect dev, data integration dev.

PR #212, opened 2021-04-16
Update pyxform-http to 1.5.0

yanokwa:
Key improvements
* Upgrade to pyxform v1.5.0
* Switch to gunicorn to reduce memory leaks and use older binaries for smaller size
* Upgrade deps, support for percent-coded form ids

I have not actually tested this PR inside Central, but risk should be pretty low. It'd be nice if this could be put though it's paces on a dev box long before the next release.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,7 +59,7 @@ services:
 
   pyxform:
     container_name: pyxform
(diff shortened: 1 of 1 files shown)


matthew-white:
I'll put this on the QA server, probably Monday. :+1:

matthew-white (approved):

PR #214, opened 2021-04-20
ops: allow for multiple instances

yanokwa:
The use of [container_name](https://docs.docker.com/compose/compose-file/compose-file-v3/#container_name) means that you can't run multiple instances of Central on the same machine. 

It also means you can't ever have a container called mail or service or secrets. Removing the container name solves this problem because Docker automatically prepends the folder name to the container (e.g., service becomes central_service_1). This happens with our PostgreSQL container already since it doesn't have a container_name.

- [x]  I have confirmed that you can upgrade existing installs of Central with this change and it loads just fine. 
- [x] I have confirmed that after upgrade, Enketo (which uses host-name based addressing) works by launching a preview. This is because Docker uses the service name to reach containers. 
- [x] I have confirmed that I can run multiple copies of Central on the same machine and they both load.

This isn't a must have change, so if it feels high risk, I'm open to waiting until after v1.2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,12 +23,12 @@ jobs:
           command: |
             set -x
             docker-compose up -d
-            docker run --network container:nginx \
+            CONTAINER_NAME=$(docker inspect -f '{{.Name}}' $(docker-compose ps -q nginx) | cut -c2-)
+            docker run --network container:$CONTAINER_NAME \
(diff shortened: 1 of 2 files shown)


issa-tseng:
i think we might depend on these names in the docs?

issa-tseng (approved):

yanokwa:
@issa-tseng Good question. The docs seem to use `docker-compose name` and those are run in the folder where Central is installed, so you can still refer to the name (e.g., service, nginx).

PR #216, opened 2021-04-26
Add issue templates

matthew-white:
Following the instructions [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository) to create issue templates. This corresponds to what seems to be a newer approach to issue templates.

For the issue templates themselves, I reuse much of the language from the readme.


lognaturel:
If you want to use the multiple templates strategy, I would prefer only allowing new issues from the bug template. Why did you add the features one and allow non-templated issues? Do we have the option of just having the forum link (i.e. would that still allow repo admins to file arbitrary issues?)? If so, that would be my preference. If not, I would prefer using the old style template with a comment that directs everything to the forum.

matthew-white:
> If you want to use the multiple templates strategy, I would prefer only allowing new issues from the bug template. Why did you add the features one and allow non-templated issues?

Ah mostly I think I did that because those were the default templates when I went through the new issue template workflow:

add-template-drop-down-menu

Right now the two templates are worded a little differently. For example, the feature request template specifically suggests going to the features category on the forum.

> Do we have the option of just having the forum link (i.e. would that still allow repo admins to file arbitrary issues?)?

With this PR, anyone could file a blank issue, though the link to do so isn't prominent:

new-issue-page-with-multiple-templates

I don't think there's a way to only allow admins to file a blank issue.

I'm not totally sure how the link in `config.yml` is shown in the UI or what happens if there's a `config.yml` but no issue template. Maybe just the link is shown, which it sounds like might be what we want? We could try merging with just `config.yml` and seeing what happens?

I feel like the current multiple templates provide a little more guidance than just the link would, but I also don't have strong preferences — happy to merge with just `config.yml` or to use a single issue template.

lognaturel:
> We could try merging with just config.yml and seeing what happens?

Agreed it's not totally clear what will happen. I'm hoping there'll be a card for the link and then the "Don't see your issue here" text at the bottom. Let's try it and see! If it does seem too unclear we can either look at the wording for the link or add a bug template.

matthew-white:
Sounds good, I think this should be ready for another look!

lognaturel (approved):

matthew-white:
OK, looks like `config.yml` is used on its own:

Screen Shot 2021-04-26 at 5 47 24 PM

The UI seems maybe slightly confusing, but I also think it's OK, and I like how prominent the link to the forum is. @lognaturel, what do you think?

PR #218, opened 2021-04-27
Separate out forum links from issue menu

lognaturel:
Sorry about the back and forth for something that should be simple, @matthew-white! This stuff feels different when it's actually deployed. I think we should move all the ODK repos in this direction so ideally we can get it right on this one first.

Goals here are to direct people more towards the right place and generally be a little friendlier.

CC @yanokwa

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,11 @@
 blank_issues_enabled: true
-contact_links:
-  - name: Discuss Central on the ODK Forum
-    url: https://forum.getodk.org
-    about: Please describe issues you're running into in the Support category, ideas for improvements in the Features category and anything else in the Development category.
+contact_links: 
+  - name: "Report an issue"
+    about: "For when Central is behaving in an unexpected way"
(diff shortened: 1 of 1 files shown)


yanokwa:
Here's my take...

```
blank_issues_enabled: true
contact_links:
- name: Report an issue
  url: https://forum.getodk.org/c/support/6
  about: For when Central is behaving in an unexpected way
- name: Request a feature
  url: https://forum.getodk.org/c/features/9
  about: For when Central is missing functionality
- name: Everything else
  url: https://forum.getodk.org/c/support/6
  about: For everything else
```

I would drop the discuss the code, but from what I can see, people don't really want to discuss the code? They either want help (user or technical) or request a feature. I don't really want to put "Get help/support" because it sets the wrong expectations.  "Report an issue" and "Everything else" feels better to me.

lognaturel:
@yanokwa's changes work for me. I've updated based on that so the PR shows the latest. I'll squash and force push once anyone who wants to comment has had a chance to do so.

matthew-white (approved):

yanokwa (approved):

PR #220, opened 2021-05-10
Update Enketo to v2.7.3

matthew-white:


lognaturel:
Very unlikely that there will be an Enketo release in the next 10 days. As a reminder, Enketo's new XPath evaluator is not yet in production on any major service I know of. There have been lots of fixes so we think risk is low but we should be ready to act of something comes up.

lognaturel (approved):

Issue #219, opened 2021-04-27
Add instructions for running Enketo for development


lognaturel:
From @matthew-white at https://github.com/getodk/central/pull/204#issuecomment-827590232:

`I think setting up Enketo locally would require changes to nginx.conf (Frontend) and config/default.json (Backend).`

lognaturel:
Enketo config:
- "server url": localhost:8989
- "allow insecure transport": true
- "base path": "-"
- "port": "6379" for redis cache

nginx.conf:
    location /- {
      proxy_pass http://localhost:8005/-;
      proxy_redirect off;
      proxy_set_header Host $host;
    }

config/default.json:
    "enketo": {
      "url": "http://localhost:8005/-",
      "apiKey": "enketorules"
    }

ktuite:
I saw this issue come across my email (or slack or something) and from these clues, was able to get it running and working with Central in development 👍 (on my mac).

Some other clues/sticking points to collect in some documentation later:
1. Which enketo?? `enketo-express` (vs. enketo-core or something)
2. Installed redis via homebrew, running in the default place, tested that it was working with `redis-cli ping`
3. My stripped down enketo config.json looks like:

```
{
    "port": "8005",
    "linked form and data server": {
        "server url": "localhost:8989",
        "api key": "enketorules",
        "authentication": {
            "allow insecure transport": "true"
        }
    },
    "base path": "-",
    "redis": {
        "cache": {
            "port": "6379"
        }
    }
}
```
4. central-backend config as described above
5. nginx config as described above, nginx restarted
6. Initially, attempting to edit a submission didn't work and gave this error: 
`{"message":"The form you tried to access is not ready for web use yet. Please wait some time and try again.","code":409.11}` 
Looking into the code, I can see this is because the Form doesn't yet have an enketo ID. This generation of this ID is something that would happen automatically if the form was created/updated after enketo was configured. Since I was using pre-existing forms, there was nothing to trigger it. I was able to re-publish an existing form with the same form definition copied over, give it a new version, and that was enough to trigger the form enketo setup.

matthew-white:
Super excited to see this! ✨ I'd love to get this into one of the readmes. Maybe `central-frontend`, since it houses the development `nginx.conf` and has existing language around Central services? 🤔 And then `central-backend` could link there?

PR #226, opened 2021-06-09
Changed main readme to mention Enketo in development

ktuite:
In response to to https://github.com/getodk/central-frontend/pull/467#issuecomment-857769473, further addresses issue #219.

I took out the bit about testing enketo by checking out a specific fork in production because it's not necessary anymore and seems like a challenging way to do enketo-related development.

But one argument for keeping it is that maybe that nugget of information isn't well-captured somewhere else? Ehh.. what nugget am I talking about? The nugget of checking out a specific commit and following the [upgrade instructions](https://docs.getodk.org/central-upgrade/) while not running `docker-compose down`.


matthew-white (approved):
This looks great! I agree with taking out the previous language, since the new setup seems much more convenient. We also document the perils of `docker-compose down` [elsewhere](https://docs.getodk.org/central-install/). Perhaps we'll add back some of this language if/when we add a new service, but I think it makes sense to replace it for now.

PR #234, opened 2021-09-01
Enable local driver for log rotation

### LINKED ISSUE #231: Container logs can very large and should be compressed
Containers logs (particularly nginx and service) can be very large and take up a lot of space. Also, it isn't fast to `docker logs container | grep foo` them for issues.

We can't use logrotate inside the containers because they all report out on stdout/err, but it looks like we can configure the “[local](https://docs.docker.com/config/containers/logging/local/)” logging driver and it does log-rotation by default and still works with `docker logs`, so we can do `docker logs --since 10m <container_id>`. 

My bias is to do this for every container in the docker-compose.yml file.

yanokwa:
Closes #231. For this PR, I verified that on the QA server...
* `docker-compose logs service` outputs logs
* `sudo docker ps -qa | sudo xargs docker inspect --format='{{.LogPath}}' | sudo xargs ls -hl` shows a reduction in log size
* `docker inspect <container>` shows max-file as a config

As detailed at https://docs.docker.com/config/containers/logging/local/#options, this defaults to a 20M file for each log and 5 maximum files (each compressed). 

In a 3 month period, we [expect 15MB for service and 640MB for nginx](https://github.com/getodk/central/issues/231#issuecomment-910864302) for a very active server, so the defaults for 100MB for service (because it's less stuff in docker-compose) and 600MB for nginx seem like good conservative defaults. We can dial the latter down if necessary.

This risk of this PR is that we are possibly throwing away old logs and someone might find that objectionable.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,8 @@ services:
       - SYSADMIN_EMAIL=${SYSADMIN_EMAIL}
     command: [ "./wait-for-it.sh", "postgres:5432", "--", "./start-odk.sh" ]
     restart: always
+    logging:
+      driver: local
   nginx:
     build:
       context: .
@@ -52,6 +54,10 @@ services:
     healthcheck:
       test: [ "CMD-SHELL", "nc -z localhost 80 || exit 1" ]
     restart: always
+    logging:
+      driver: local
+      options:
+        max-file: "30"
   pyxform:
     image: 'getodk/pyxform-http:v1.5.1'
     restart: always


yanokwa:
@issa-tseng, per your suggestion, I looked into what version we would need...

Support for the `local` driver was added to docker-compose in v1.28.3 (https://github.com/docker/compose/pull/8082).

We recommend this DigitalOcean [droplet](https://marketplace.digitalocean.com/apps/docker) which runs docker-compose v1.27.0,  so we should add docs for v1.3 that people need to upgrade their docker-compose. 

Instructions are at https://docs.docker.com/compose/install/. 

```
sudo curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
sudo chmod +x /usr/local/bin/docker-compose
```

We use v1.29.2 on ODK Cloud and it works great.

issa-tseng (approved):

Issue #223, opened 2021-05-18
Central v1.3 release


lognaturel:
This release focuses on providing more information about submission edits, adding access information to logs, and performing regular maintenance. It is intended to be a shorter release.

Criteria will be shared soon!

matthew-white:
Release criteria: https://docs.google.com/document/d/1dIoPkYDI8AChbJiicce6Jc3eF1NYBdcEcnauSpYjLYA/edit?usp=sharing

PR #237, opened 2021-09-22
Changes for v1.3

matthew-white:
This PR prepares the release of v1.3 (closing #223). It only contains commits from other PRs that have already been approved and merged:

#230
#236
#234
#232
#235
#238
#239
#240

@yanokwa, I've tagged you for review. Would be great if you could ✅ before release next week!

### DIFF
--- a/files/service/crontab
+++ b/files/service/crontab
@@ -1,2 +1,3 @@
+0 3 * * * root /usr/odk/run-analytics.sh
 0 2 * * * root /usr/odk/run-backup.sh
 0 1 * * 0 root /usr/odk/reap-sessions.sh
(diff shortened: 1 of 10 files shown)


yanokwa:
I can confirm my commits are here. I've added @issa-tseng so she can do the same.

files/service/crontab
```diff
@@ -1,2 +1,3 @@
+0 3 * * * root /usr/odk/run-analytics.sh
```
yanokwa, line 1: We could call this run-usage-reporting or something similar to match the UI. Obviously a nitpick, so feel free to ignore.

matthew-white: We actually use the word "analytics" not just here, but throughout the Backend/Frontend codebases, as well as in the API. I think that's OK for more of an internal/technical name.

issa-tseng (approved):

PR #238, opened 2021-09-25
Prevent exploitable TLS v1.0 and cross-frame scripting

yanokwa:
Per @issa-tseng's guidance, I tried to do the most narrow change from our pen test. 

The slightly controversial change is increasing the Strict-Transport-Security age and adding always. The age matches the [Moz guide](https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&guideline=5.60) instead of the [Nginx guide](https://www.nginx.com/blog/http-strict-transport-security-hsts-and-nginx/), but I don't think we've done anything dangerous.

Things I tried to verify.
* https://www.ssllabs.com/ssltest/analyze.html?d=dev.getodk.cloud reports the server has TLS 1.3
* https://gf.dev/x-frame-options-test says we prevent click-jacking.
* Enketo logged in submissions and edits work
* Enketo public access links work
* Collect submissions work

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -4,13 +4,17 @@ server {
   ssl_certificate_key /etc/${SSL_TYPE}/live/${CNAME}/privkey.pem;
   ssl_trusted_certificate /etc/${SSL_TYPE}/live/${CNAME}/fullchain.pem;
 
-  ssl_protocols TLSv1.2 TLSv1.1 TLSv1;
-  ssl_prefer_server_ciphers on;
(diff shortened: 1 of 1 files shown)


issa-tseng:
i wonder if the frame policy may need to be configurable someday, if anybody is trying to do cheap integration via iframe. security hole to one shop, vital pipeline for another.

issa-tseng (approved):

yanokwa:
@issa-tseng Good point. Let's see if anyone complains and then we can make it customizable after the fact. We should bias to secure by default.

PR #239, opened 2021-09-27
Add Cache-Control header for /version.txt and /index.html

matthew-white:
This commit aims to prevent excessive caching of /version.txt and /index.html.

Prior to this commit, I noticed the following behavior:

  - Open Frontend in a browser.
  - Change the source, then rebuild. This will change /version.txt.
  - However, if you click on the Version link in the help dropdown, the old /version.txt is shown in a new tab.
  - Only if you refresh the new tab is the new /version.txt shown.

getodk/central-frontend#496 only works if requesting /version.txt returns the current version, reflecting any change.

Adding the `Cache-Control` header for /version.txt seems to resolve this issue. I also added it for /index.html, which should make it unnecessary to ever force-refresh /index.html in a browser.

I think `Cache-Control` is useful for these two files, but I don't think that it is useful for other files:

  - CSS and JS files from Vue will have a hash in their filenames that will change as the associated source changes.
  - We use a query string for cache-busting when requesting icon files.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -49,6 +49,13 @@ server {
 
   location / {
     root /usr/share/nginx/html;
+
+    location /version.txt {
+      add_header Cache-Control no-cache;
+    }
+    location /index.html {
+      add_header Cache-Control no-cache;
+    }
   }
 }
 


files/nginx/odk.conf.template
```diff
@@ -49,6 +49,13 @@ server {
 
   location / {
     root /usr/share/nginx/html;
+
+    location /version.txt {
+      add_header Cache-Control no-cache;
+    }
+    location /index.html {
+      add_header Cache-Control no-cache;
+    }
```
matthew-white, line 58: I could also use a regular expression instead of specifying two `location` blocks.

files/nginx/odk.conf.template
```diff
@@ -49,6 +49,13 @@ server {
 
   location / {
     root /usr/share/nginx/html;
+
+    location /version.txt {
+      add_header Cache-Control no-cache;
```
matthew-white, line 54: This value of `Cache-Control` seems right for the browsers we're targeting. [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#requiring_revalidation) says:

> `no-cache` and `max-age=0, must-revalidate` have the same meaning. Clients can cache a resource but must revalidate each time before using it. This means HTTP request occurs each time though, it can skip downloading HTTP body if the content is valid.

I tried this on a QA server, and things seem to work as expected.

matthew-white, line 54: One question I have is whether I need to specify any other header in addition to `Cache-Control`. From a first look, I don't think I need to, but there are a number of headers related to caching, so I'm not quite sure.

issa-tseng: it may be more hygienic to suppress some of the others if they exist, like ETag, but i don't think it's critical at all. and this should probably always have had this flag on it.

issa-tseng (approved):

PR #243, opened 2021-10-04
Use stock run_certbot.sh and add required server_name to conf

yanokwa:
We started using our own `run_certbot.sh` at https://github.com/getodk/central/commit/74f3a4f48525410827e033d882c0e890b985df1c#diff-89a626365199d57de80dce5b8e5a4a7790909efcb20f31749038ff43278b8f38 to fix an nginx reloading issue. That meant that when we upgrade to `jonasal/nginx-certbot:2.4` we were using an out-of-date script. I have removed our copy of that script so we use whatever comes with the container.

That script uses `server_name` in the conf to build the [certificate request](https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/master/src/scripts/run_certbot.sh#L64) to LetsEncrypt, so I've added it. The call stack...
- https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/master/src/scripts/run_certbot.sh#L99
- https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/master/src/scripts/run_certbot.sh#L93
- https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/master/src/scripts/util.sh#L63

You can also see the `server_name` used at https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/master/examples/example_server.conf.

My assumptions:
* The original nginx reloading issue likely isn't present in jonasal's container. 
* Adding ${CNAME} wouldn't break non-LetsEncrypt setup. I looked at `odk-setup.sh` and I think we are fine.
* We don't need something else in `odk.conf`. I looked at [run_certbot.sh](https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/master/src/scripts/run_certbot.sh) and I think we are fine.

I have verified that:
- [x] Fresh v1.3.0 install loads with a working cert
- [x] Upgrade from v1.2.2 loads with a working cert
- [x] upstream, selfsign, customssl produce odk.conf files that look reasonable to my eye


issa-tseng:
yes. (facepalm)

issa-tseng (approved):

PR #261, opened 2021-11-18
Downgrade node to fix stream closing

yanokwa:
Tried this PR on a few servers and those servers don't report the premature close errors we've seen.

I changed node in the intermediate nginx build as well to reduce the number of node containers on the host machine.

### DIFF
--- a/nginx.dockerfile
+++ b/nginx.dockerfile
@@ -1,4 +1,4 @@
-FROM node:14.17.6 as intermediate
+FROM node:12.22.7 as intermediate
 
 COPY ./ ./
 RUN files/prebuild/write-version.sh
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -1,4 +1,4 @@
-FROM node:14.17.6
+FROM node:12.22.7
 
 WORKDIR /usr/odk
 


issa-tseng (approved):
that's not the one we had before though eh?

ktuite:
I was on node 12 locally last week and ran into an error (triggered by a migration) here with the `?.` 
https://github.com/getodk/central-backend/blob/master/lib/data/schema.js#L408

yanokwa:
@issa-tseng, this is the one latest v12. I'm glad to go back to the version we had previously if you think it's safer (it's probably safer). 

@ktuite Do you remember the specific node 12 you were on?

ktuite:
> @ktuite Do you remember the specific node 12 you were on?

looks like `v12.21.0`

matthew-white:
In addition to removing `?.`, I guess we'll also need to revert getodk/central-backend@42cfc8d9ec89ee49dbeee51c09c4396d25d62c59. Would probably be a good idea to bump the Backend CircleCI Node version back to 12 just to make sure that nothing else fails.

issa-tseng:
with that all on the table i'll ask whether it's worth putting everybody into an unknown new configuration to solve a problem we don't know is widespread?

yanokwa:
I'm seeing the same problems now with node 12, so I'm closing this PR. 😭

PR #262, opened 2022-01-24
Daily cron job for purging deleted froms

ktuite:
Cron job for a daily script that purges forms deleted over 30 days prior. The task and script are captured in this backend PR: getodk/central-backend#445


ktuite (commented):
could this script be part of the odk cli?

issa-tseng (approved):

PR #274, opened 2022-03-03
Switch from GA to Fathom

yanokwa:
Austria and France have both ruled that GA is not GDPR compliant. Fathom is.

I've deployed it on getodk.org, forum.getodk.org, docs.getodk.org, and xlsform.getodk.org so far and it's working great. 

No cookies are required so I think cross-site issues. That said, it'd be good if someone could confirm that this won't throw up any cross-site errors in Central.

### DIFF
--- a/docs/news.html
+++ b/docs/news.html
@@ -4,14 +4,7 @@
     <meta charset="utf-8">
     <title>ODK Central</title>
     <link rel="stylesheet" href="style.css">
-    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-91951913-7"></script>
-    <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-
-      gtag('config', 'UA-91951913-7', { cookie_flags: 'SameSite=None;Secure' });
-    </script>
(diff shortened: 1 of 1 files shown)


matthew-white (approved):
Looks good to me! I don't foresee there being any issues here, though we should double-check once the PR is merged.

I'm wondering though whether it'd be simpler in some ways to move this into Frontend. I wouldn't have wanted to load GA within Frontend, but a one-line `<script>` tag that asynchronously loads a Fathom instance hosted on getodk.org seems fine to me. I think there would be benefits to disconnecting news from analytics, since it'd allow us to make changes to the news section without worrying about how it affects our analytics.

If that sounds like the right move, feel free to just remove GA here and add Fathom to index.html in Frontend. But also if you just want to merge this PR and wait until later to consider moving things to Frontend, that also sounds fine to me.

matthew-white:
Hmm though could it be slightly more secure from an XSS perspective loading it in an iframe, in case somehow script.js is compromised? Not sure how much of a risk that really is though.

yanokwa:
When we load it in an iframe as we do now, all the data is bound to the Github domain. I believe moving it to frontend will mean we know the domain of every instance and I'd rather not do that.

matthew-white:
Ah right, that makes sense. I'll go ahead and merge now!

PR #287, opened 2022-05-31
Enable security improvements

yanokwa:
This PR should be merged at the beginning of Block 16. 

QA should verify that the security changes (frameguard, no sniff, hsts, csp) don't break existing functionality.

https://enketo.github.io/enketo-express/tutorial-10-configure.html has more on the changes.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -13,10 +13,11 @@ server {
   ssl_dhparam /etc/dh/nginx.pem;
 
   server_tokens off;
-  add_header Strict-Transport-Security "max-age=63072000" always;
 
-  add_header X-Frame-Options "SAMEORIGIN";
+  add_header Content-Security-Policy "default-src 'self';";
+  add_header Strict-Transport-Security "max-age=63072000" always;
   add_header X-Content-Type-Options nosniff;
+  add_header X-Frame-Options "SAMEORIGIN";
 
   client_max_body_size 100m;
 
@@ -59,4 +60,3 @@ server {
     }
   }
 }
-


files/enketo/config.json.template
```diff
@@ -34,5 +34,14 @@
...
+    "text field character limit": 1000000,
+    "frameguard deny": true,
+    "no sniff": true,
+    "hsts": {
+        "seconds": 63072000
+    },
+    "csp": {
+        "enabled": true
+    },
+    "exclude non-relevant": true
```
matthew-white, line 46: Based on what @lognaturel [posted](https://forum.getodk.org/t/feedback-needed-enketo-relevance-change-to-align-with-odk-collect/37719) in the forum, I thought we were going to wait until v1.6 to specify `true` here. Did it end up seeming alright to make that change in a v1.5 patch release?

lognaturel: No, all of this is for release with Block 16. I think Yaw just mixed up processes. It should be merged for release with 1.6.

matthew-white:
@lognaturel, are you familiar with these config options already? If so, would you be up for a quick review?

files/enketo/config.json.template
```diff
@@ -34,5 +34,13 @@
     "support": {
         "email": "support@getodk.org"
     },
-    "text field character limit": 1000000
-}
+    "text field character limit": 1000000,
+    "frameguard deny": true,
+    "no sniff": true,
+    "hsts": {
+        "seconds": 63072000
```
lognaturel, line 41: This is 2 years which seems like the recommended value according to https://hstspreload.org/. But it looks like `includeSubDomains` and `preload` should be specified. Shouldn't this be done at the nginx level in our case?

lognaturel:
They were added to the Enketo config in April: https://github.com/enketo/enketo-express/pull/402

My sense is that in Central's case they should be configured for nginx but I'm not sure. Maybe Trevor is a good person to pull in to review?

lognaturel:
Going to https://geekflare.com/tools/hsts-test, I can confirm that HSTS is configured for /v1/ and /-/ via nginx [here](https://github.com/getodk/central/blob/master/files/nginx/odk.conf.template#L16). That seems good. If I put in the raw domain of a Central server it says it doesn't get the header. Is that expected?

https://github.com/getodk/central/blob/master/files/nginx/odk.conf.template#L18 takes care of  "frameguard deny" and "no sniff". I think all we're missing is CSP.

files/nginx/odk.conf.template
```diff
@@ -13,10 +13,11 @@ server {
   ssl_dhparam /etc/dh/nginx.pem;
 
   server_tokens off;
-  add_header Strict-Transport-Security "max-age=63072000" always;
 
-  add_header X-Frame-Options "SAMEORIGIN";
+  add_header Content-Security-Policy "default-src 'self';";
```
yanokwa, line 17: This is the only new change. The rest is from sorting.

alxndrsn: do we need to add `from-src https://thirtyfive-england.getodk.org/` here?

matthew-white:
Closing in favor of #308 and #309.

files/enketo/config.json.template
matthew-white: @alxndrsn, just wanted to make sure you saw this comment! Do you think we need either of these for our HSTS?

alxndrsn: Thanks @matthew-white.

* `includeSubDomains`: I'm not sure central has subdomains configured by default.  Maybe it's good to play it safe, but if 3rd-party deployments set up their own non-http sub-domains then this could be quite confusing.
* `preload`: getodk.cloud subdomains are not currently eligible for this due to configuration of the root domain (https://getodk.cloud).  I suspect `preload` should not be included in public `central` config anyway due to https://hstspreload.org/#opt-in

PR #292, opened 2022-06-10
Add support for SSL running on different ports upstream

yanokwa:
I don't think this should be merged, but I want to show my preferred alternative to https://github.com/getodk/central/pull/291/files. 

I think perhaps what we should do is add docs for people to edit their .env and docker-compose to match what is in this PR, and warn them that it's subject to change. This way, most users don't have to change anything.

I've confirmed that with an upstream server, web forms with previews (including with images) works. I've also confirmed that  web submissions work and password resets show the correct port.

Truth be told, I regret allowing port customization and upstream servers to begin with. It's too fragile.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,7 +32,7 @@ services:
       - /data/transfer:/data/transfer
     environment:
       - DOMAIN=${DOMAIN}
-      - HTTPS_PORT=${HTTPS_PORT:-443}
+      - HTTPS_PORT=${UPSTREAM_HTTPS_PORT:-443}
       - SYSADMIN_EMAIL=${SYSADMIN_EMAIL}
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -83,7 +83,7 @@ services:
     environment:
       - DOMAIN=${DOMAIN}
       - SUPPORT_EMAIL=${SYSADMIN_EMAIL}
-      - HTTPS_PORT=${HTTPS_PORT:-443}
+      - HTTPS_PORT=${UPSTREAM_HTTPS_PORT:-443}
```
yanokwa, line 86: I don't think this is actually required. The port information seems to be used as a key to media, and I think it's nice to have that key be like the URL.

yanokwa:
Closing in favor of https://github.com/getodk/docs/pull/1540.

Issue #265, opened 2022-02-10
Requests logged twice for service


matthew-white:
I have noticed on a couple of different servers that requests seem to be logged twice for `service`. Interestingly, the requests don't seem to be logged twice for `nginx`. Requests do seem to be logged twice for `service`; I am less sure whether errors and warnings are logged twice, but I think that that's also possible. Even though requests are logged twice, I doubt that two requests are actually being sent, because that would almost certainly lead to user-facing errors that we would have noticed.

lognaturel:
@alxndrsn and I chatted briefly about this and he asked for confirmation that it's always happening and that there's no funny business based on request type or anything like that. From what I can tell, every request is always double logged and it looks like a configuration issue of some sort. 

`docker-compose logs --tail 12 service` from the QA server:
```
service_1 | <ip> - - [24/Jun/2022:18:57:14 +0000] "POST /v1/sessions HTTP/1.0" 401 80
service_1 | <ip> - - [24/Jun/2022:18:57:14 +0000] "POST /v1/sessions HTTP/1.0" 401 80

service_1 | <ip> - - [24/Jun/2022:18:59:14 +0000] "POST /v1/sessions HTTP/1.0" 401 80
service_1 | <ip> - - [24/Jun/2022:18:59:14 +0000] "POST /v1/sessions HTTP/1.0" 401 80

service_1 | <ip> - - [24/Jun/2022:19:00:04 +0000] "GET /v1/test/<key>/projects/149/forms/regex/draft/formList HTTP/1.0" 404 458
service_1 | <ip> - - [24/Jun/2022:19:00:04 +0000] "GET /v1/test/<key>/projects/149/forms/regex/draft/formList HTTP/1.0" 404 458

service_1 | <ip> - - [24/Jun/2022:19:00:04 +0000] "GET /v1/test/<key>/projects/149/forms/random-no-dupes/draft/formList HTTP/1.0" 200 465
service_1 | <ip> - - [24/Jun/2022:19:00:04 +0000] "GET /v1/test/<key>/projects/149/forms/random-no-dupes/draft/formList HTTP/1.0" 200 465

service_1 | <ip> - - [24/Jun/2022:19:00:04 +0000] "GET /v1/test/<key>/projects/149/forms/filteringgameRPG/draft/formList HTTP/1.0" 200 468
[...]

PR #293, opened 2022-06-26
pm2: remove duplicate logging

alxndrsn:
Closes https://github.com/getodk/central/issues/265?

This change seems to remove the duplicates from `central-backend` logging when run from `docker-compose`.

I haven't had a chance to test this on a real deployment, but _it works on my machine_ :tm: :grin:

### DIFF
--- a/files/service/pm2.config.js
+++ b/files/service/pm2.config.js
@@ -6,10 +6,6 @@ module.exports = {
     // the default is 1600ms; we aren't that impatient:
     kill_timeout: 30000,
 
-    // log to stdout/stderr:
-    out_file: '/proc/1/fd/1',
-    error_file: '/proc/1/fd/2',
-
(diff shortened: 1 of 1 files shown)


matthew-white (commented):
I'd love to better understand why this works! If `out_file` and `error_file` aren't specified, then how is any logging getting through?

> I haven't had a chance to test this on a real deployment

I think you'll have access to a real deployment soon (if you don't already). Would be great to verify that it works there as well!

matthew-white:
We've decided that it's not super important that we understand exactly why this works. @alxndrsn, just to check (and you might have said this already), but were you able to verify this change on a real deployment?

alxndrsn:
@matthew-white this seems to work (tested on dev.getodk.cloud)

PR #301, opened 2022-07-04
Added Helm Chart for ODK Central

ntandomng:
Helm chart for odk-central deployment.

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -1,0 +1,23 @@
+apiVersion: v2
+name: odk-central
+description: An ODK Central Helm chart for Kubernetes
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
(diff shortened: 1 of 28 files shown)


lognaturel:
Thanks so much for this nice contribution to the ODK ecosystem! A number of folks have asked about Kubernetes deployment. 

We encourage you to continue maintaining this as its own repository. You can let people know about it on the forum showcase (https://forum.getodk.org/c/showcase/10) and see if anyone else might want to collaborate with you. We also find that there is a big ODK community on Twitter.

The core team has limited capacity so we focus on adding functionality to Central and providing one well-tested deployment path. Things like this are great areas for community members to get involved!

ntandomng:
CC'ing colleagues

@avikganguly01 and @fynmanoj

lognaturel:
Great! If you’re part of a broader organization, consider publishing this as part of that organization. 

A helpful thing to include in the readme would be details on the environment you’ve deployed to. Then you can ask other users and contributors to also report how they use it.

Issue #311, opened 2022-08-25
Update Sentry ingestion URLs (deadline: Nov 1 2022)


alxndrsn:
> On November 1 [2022] ... event ingestion will be disabled on legacy ingestion endpoints - `https://sentry.io` and `https://app.getsentry.com`.
>
> -- _https://blog.sentry.io/2022/08/22/deprecating-tls-1-0-and-1-1_

This URL is currently set in `central-backend` at https://github.com/getodk/central-backend/blob/4e41e719ca40a122a14bf6a366dab7d37d6cc1d6/lib/external/sentry.js#L120.

As per https://blog.sentry.io/2022/08/22/deprecating-tls-1-0-and-1-1#deprecating-ingestion-endpoints, it looks like the "organization subdomain" will need to be added as a config option at: https://github.com/getodk/central/blob/63cfab894cd619a703a1a38a00756f864e58b42b/files/service/config.json.template#L30-L33

A corresponding central-backend change will also be required (filed at https://github.com/getodk/central-backend/issues/584).

matthew-white:
Thanks for filing this, @alxndrsn!

I guess this means that we'll stop receiving events from old versions of Central (getodk/central-backend#547), at least for one release!

matthew-white:
Closed by #312 and getodk/central-backend#585.

PR #313, opened 2022-08-25
Enable Content-Security-Policy reporting

alxndrsn:
This will _report_ content security policy violations to Sentry, but _will not_ block content.  This might be a useful step toward discovering if the rules configured here are safe.

Enketo Express requires a much more permissive policy than central-frontend, so its policy is configured separately.

It might be useful to pull the sentry config (organization subdomain, project ID, API key) into e.g. `docker-compose.yml`, as after https://github.com/getodk/central/issues/311 is fixed, they will all have to be configured twice (once for nginx, once for central-backend).


files/nginx/odk.conf.template
```diff
@@ -28,6 +28,15 @@ server {
     proxy_pass http://enketo:8005/-;
     proxy_redirect off;
     proxy_set_header Host $host;
+
+    # More lax CSP for enketo-express:
+    # Google Maps API: https://developers.google.com/maps/documentation/javascript/content-security-policy
+    add_header Content-Security-Policy-Report-Only "default-src 'none'; connect-src 'self' blob: https://maps.googleapis.com/maps/api/mapsjs/; font-src 'self' https://fonts.gstatic.com/; frame-src 'none'; img-src data: blob: 'self' https://maps.google.com/maps/ https://maps.gstatic.com/mapfiles/ https://maps.googleapis.com/maps/api/js/StaticMapService.GetMapImage; manifest-src 'none'; media-src 'none'; script-src 'unsafe-inline' 'self' https://maps.googleapis.com/maps/api/js/ https://maps.google.com/maps/ https://maps.google.com/maps-api-v3/api/js/; style-src 'unsafe-inline' 'self' https://fonts.googleapis.com/css; object-src 'none'; report-uri https://o130137.ingest.sentry.io/api/5724763/security/?sentry_key=25f147a612f3464ca2dd46735434bddb";
```
matthew-white, line 34: I'm noticing that there's a default Enketo CSP that can be enabled: https://github.com/enketo/enketo-express/blob/35e0f2d43d3ae90bbe8957bcdaf35cd82550137d/config/express.js#L148-L149. It sounds like that won't work for us, since we need to account for Google Maps. But I thought I'd mention it in case there's anything in there that's relevant for us.

matthew-white:
> This will _report_ content security policy violations to Sentry, but _will not_ block content.

That sounds like a reasonable next step to me. So we would ship Content-Security-Policy-Report-Only with the next release (v2022.3), then maybe ship Content-Security-Policy in the release after that?

Once we feel reasonably confident about our Content-Security-Policy-Report-Only setting, maybe we could put that on some ODK Cloud servers even before v2022.3.

matthew-white:
I just noticed that a report of a CSP violation will include query parameters, including the sensitive `st` query parameter for public links. In other cases, we remove the `st` query parameter before sending the error to Sentry: https://github.com/getodk/central-backend/blob/87aedb0b58973d7454ddd3a543fad7547646bcdd/lib/external/sentry.js#L19-L21. For a CSP violation, is there a way that we could filter out the `st` query parameter (or all query parameters)? Just a thought: could we set the `report-uri` to a Backend endpoint that does that filtering before sending on the violation to Sentry? (If we did that, we also wouldn't need to plumb the Sentry configuration to nginx because Backend would be the one communicating with Sentry.)

matthew-white:
I'm also noticing that the settings page in Sentry for Content Security Policy says:

> We recommend setting this up to only run on a percentage of requests, as otherwise you may find that you've quickly exhausted your quota.

Is this something worth doing for us? Maybe we wouldn't have enough CSP violations to risk exhausting our quota?

alxndrsn:
> Maybe we wouldn't have enough CSP violations to risk exhausting our quota?

Currently the number is very low, and I'm working to decrease.

> For a CSP violation, is there a way that we could filter out the st query parameter (or all query parameters)?

It looks like this could be done on the Sentry side with https://docs.sentry.io/product/data-management-settings/scrubbing/server-side-scrubbing/.  Would that be sufficient?

matthew-white:
Just wanted to note that it looks like there's now a conflict.

> It looks like this could be done on the Sentry side with https://docs.sentry.io/product/data-management-settings/scrubbing/server-side-scrubbing/. Would that be sufficient?

I feel like it'd be preferable to do that scrubbing before the data even reaches Sentry. That'd also make it clear to anyone interested that that scrubbing is happening: folks can't look at our server-side Sentry setup, but they can look at our source. But that's just my two cents / it might be worth discussing further as a team. Maybe it'd depend on how hard it'd be to add a Backend endpoint to modify the violation report before forwarding it on to Sentry?

matthew-white:
Do we not want to see IP addresses in Sentry? It looks like those are coming through the CSP reports as well.

alxndrsn:
To avoid collecting end user IP addresses, proxying the CSP report requests through nginx might be a simple option, and the query param scrubbing could be done there too - I'll look into this.

alxndrsn:
@matthew-white I've updated this PR to scrub IPs by proxying the reports through nginx.

It's less clear how easy it is changing values inside the forwarded request body.  Would it be acceptable to do this in Sentry?

matthew-white:
> I've updated this PR to scrub IPs by proxying the reports through nginx.

Just to make sure I understand what's happening here, will the IP address of the server still be shown in Sentry? I think we probably want to avoid that as well. If that's what's happening, maybe we could keep the new /csp-report endpoint, but proxy all CSP reports through data.getodk.cloud/csp-report specifically?

> It's less clear how easy it is changing values inside the forwarded request body. Would it be acceptable to do this in Sentry?

I could see how that might be the easiest thing to do. CC @lognaturel for a second opinion on whether it's important to scrub data before it reaches Sentry.

alxndrsn:
> Just to make sure I understand what's happening here, will the **IP address of the server** still be shown in Sentry?

Yes.  But at this point I'd be interested to know what proxying through data.getodk.cloud is trying to protect against - if a user wants to protect their server's IP address being known to Sentry, then they have little reason to trust data.getodk.cloud vs sentry.io, and should probably just disable Sentry completely.

Also, _end users_ are already sending their IP address to https://getodk.github.io/central/news.html every time they load the frontend homepage.  This should probably be considered alongside CSP-related IP address sharing.

files/nginx/common-headers.nginx.conf
```diff
@@ -2,6 +2,7 @@
 # which has a call to add_header.
 # See: https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header
 
+add_header Content-Security-Policy-Report-Only "default-src 'none'; connect-src 'self'; font-src 'self'; frame-src 'self' https://getodk.github.io/central/news.html; img-src data: *; manifest-src 'none'; media-src 'none'; script-src 'self'; style-src 'self'; style-src-attr 'unsafe-inline'; object-src 'none'; report-uri /csp-report";
```
lognaturel, line 5: My understanding is that `default-src` isthe default so here the following are redundant:
- `manifest-src 'none'`
- `media-src 'none'`

`img-src data: *` - is the `data` here necessary? It seems it is possible to directly embed data URIs in markdown but it seems fine to block it.

The rest looks good to me!

files/nginx/odk.conf.template
```diff
@@ -28,6 +28,15 @@ server {
     proxy_pass http://enketo:8005/-;
     proxy_redirect off;
     proxy_set_header Host $host;
+
+    # More lax CSP for enketo-express:
+    # Google Maps API: https://developers.google.com/maps/documentation/javascript/content-security-policy
+    add_header Content-Security-Policy-Report-Only "default-src 'none'; connect-src 'self' blob: https://maps.googleapis.com/maps/ https://maps.google.com/ https://maps.gstatic.com/mapfiles/ https://fonts.gstatic.com/ https://fonts.googleapis.com/; font-src 'self' https://fonts.gstatic.com/; frame-src 'none'; img-src data: blob: jr: 'self' https://maps.google.com/maps/ https://maps.gstatic.com/mapfiles/ https://maps.googleapis.com/maps/; manifest-src 'none'; media-src blob: jr: 'self'; script-src 'unsafe-inline' 'self' https://maps.googleapis.com/maps/api/js/ https://maps.google.com/maps/ https://maps.google.com/maps-api-v3/api/js/; style-src 'unsafe-inline' 'self' https://fonts.googleapis.com/css; object-src 'none'; report-uri /csp-report;
+    # Other security headers are identical to those in common-headers.nginx.conf;
```
lognaturel, line 35: Are `blob` `jr` still needed now that we've changed the way media is served from Enketo? I know that `data` is still required for images because the Enketo logo itself is served that way but I'm not sure about the others.

lognaturel:
I think that we shouldn't expect many violations to be reported by the time we release this and that as @alxndrsn states there are ways for users who really don't trust us to opt out of all of this.

I'd rather we focus on making sure we feel really good about the rules and not worry too much about the reporting, especially since it's short-term.

files/nginx/odk.conf.template
```diff
@@ -28,6 +28,15 @@ server {
...
+    # We can't just include that file here though, as it will set two Content-Security-Policy* headers
+    add_header Strict-Transport-Security "max-age=63072000" always;
+    add_header X-Frame-Options "SAMEORIGIN";
```
lognaturel, line 38: Some folks do want to embed their Enketo forms in iframes (https://apidocs.enketo.org/v2#/post-survey-iframe). Was that already not possible unless the parent is on the same origin with prior versions of Central?

lognaturel: Seems like maybe to properly/securely support that we'd want to make `frame-ancestors` configurable?

lognaturel: Evidence that someone uses Central and Enketo with iframes today: https://github.com/enketo/enketo-express/pull/454

yanokwa:
I did some research to try to understand what we should do about IPs.

Any service you use will store IP addresses as part of maintaining the security and integrity of that service. There's nothing wrong or illegal with that. Under GDPR, that's considered a "legitimate interest" and only requires opt-out. 

Personally, what I'd like us to avoid is storing IPs of non-Cloud users so we can't identify them. Ideally, we'd anonymize client side, but it's not terrible if Sentry can do it server side for us.

Reducing client IP and query param exposure via the local nginx server is straightforward and helpful and so let's keep doing that. Proxing everything through data.getodk.cloud doesn't seem helpful because we are just shifting the storage of IPs from Sentry's servers to our servers.

Because our goal is increase the security of Central and reduce the number of violations over time, I think the above approach is reasonable, defensible, and legal.

matthew-white:
OK cool, that's helpful! It sounds like @alxndrsn's approach should work well for IP addresses.

It's looking like it might be hard to scrub the session token query param within nginx. If I understand correctly, the query param isn't on the URL itself, but instead is in the request body: the query param is part of a value within a JSON report. @yanokwa, is it OK to scrub the `st` query param within Sentry? (Typically, the value of that query parameter is a session token that provides access to a form.) The alternative would probably be to do that scrubbing within a new Backend endpoint.

@lognaturel, that makes sense about focusing more on the rules than the reporting. I do think that we're planning to keep the reporting indefinitely: even once we change the header from `Content-Security-Policy-Report-Only` to `Content-Security-Policy`, we'll still specify `report-uri` so that we continue to hear about violations.

files/nginx/odk.conf.template
```diff
@@ -58,5 +67,9 @@ server {
       add_header Cache-Control no-cache;
     }
   }
+
+  location /csp-report {
+    proxy_pass https://o130137.ingest.sentry.io/api/5724763/security/?sentry_key=25f147a612f3464ca2dd46735434bddb;
```
yanokwa, line 72: @alxndrsn Seems like we should generate a new key just for this, no?

files/nginx/odk.conf.template
matthew-white, line 35: I think it'd be useful to add a parallel comment in common-headers.nginx.conf noting that the headers there are mostly duplicated here. That way if we add a new header to common-headers.nginx.conf, we won't forget to add it here.

files/nginx/common-headers.nginx.conf
alxndrsn: IIRC `data` is required by a particular cat gif...

files/nginx/odk.conf.template
alxndrsn: How would configuration normally occur?  Presumably anyone running their own server _could_ edit these files directly, but perhaps there are more friendly approaches.

files/nginx/odk.conf.template
alxndrsn: :+1:  added

alxndrsn: Reverted.

files/nginx/common-headers.nginx.conf
alxndrsn: Reverted.

alxndrsn: I'm going to re-add all of these "no-op" rules - they don't affect the policy, but they _do_ affect the content of the reports - reports contain the rule which was violated, and knowing it was e.g. `object-src` vs `default-src` seems useful.

files/nginx/odk.conf.template
lognaturel: > How would configuration normally occur?

Not sure. I guess the most important question in my mind is whether this is a breaking change for folks who use Enketo in iframes.

files/nginx/common-headers.nginx.conf
lognaturel: > IIRC data is required by a particular cat gif...

I don't think so. That would be `*`. I think we need `image-src: *` but not `data`.

> I'm going to re-add all of these "no-op" rules

Got it, that probably makes sense? Or would we know from the URL of the thing that triggered the violation what it is? If you do want to add them back, I would find grouping them together and adding a comment helpful.

lognaturel:
I'm eager to get it back on the QA server so we can see how it does! Looks like https://github.com/getodk/central/pull/313#discussion_r981573982 and https://github.com/getodk/central/pull/313#discussion_r981554605 need resolution.

files/nginx/common-headers.nginx.conf
alxndrsn: > I don't think so. That would be `*`. I think we need `image-src: *` but not `data`.

I'm pretty sure `*` will include `data` :slightly_smiling_face:

> I would find grouping them together and adding a comment helpful.

I'll add a comment.  Currently the rules are alphabetical.  I'd rather maintain this, as it's otherwise hard to keep track of which rules are set and which are not in such a long string...  If you know options for multiline strings in nginx config, that could help a lot!

alxndrsn: > would we know from the URL of the thing that triggered the violation what it is?

I don't think so: e.g. if a script was hidden at http://example.com/bad-script.js.gif, we have the option of seeing a violation of `script-src` or of `default-src`.  If it's reported as a violation of `default-src`, I don't think we can tell whether the client tried to load it via `<script>` or ``.

files/nginx/odk.conf.template
alxndrsn: `X-Frame-Options` isn't being introduced in this PR - it just needs to be repeated here in this PR because any call to `add_header` inside a `location` block overrides _all_ previous calls to `add_header`.

files/nginx/common-headers.nginx.conf
lognaturel: > I'm pretty sure * will include data 🙂

It does not. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources#sources

> If it's reported as a violation of default-src, I don't think we can tell whether the client tried to load it via <script> or .

Ah, tricky. Ok, that makes sense.

files/nginx/odk.conf.template
lognaturel: That makes sense. Very interesting. I wonder whether people using iframes are modifying that policy or just embedding in a page on the same domain. That can be a question for another time.

files/nginx/common-headers.nginx.conf
```diff
@@ -2,6 +2,14 @@
 # which has a call to add_header.
 # See: https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header
 
+# Rules set to 'none' here would fallback to default-src if excluded.
+# They are included here to ease interpretation of violation reports.
+#
+# N.B. a separate CSP is defined for Enketo in odk.conf.template
+add_header Content-Security-Policy-Report-Only "default-src 'none'; connect-src 'self'; font-src 'self'; frame-src 'self' https://getodk.github.io/central/news.html; img-src *; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'self'; style-src 'self'; style-src-attr 'unsafe-inline'; report-uri /csp-report";
```
lognaturel, line 9: One could spend one's entire life studying CSP headers. I found https://csplite.com/csp113/ helpful for understanding `style-src-attr 'unsafe-inline'` and its implications. Seems reasonable to me.

files/nginx/odk.conf.template
lognaturel: This got unreverted, I think. I don't see a reason for `blob` and `jr`.

lognaturel (changes requested):
One last issue that I can see.

files/nginx/odk.conf.template
lognaturel: Ok, just kidding, I see that `form-cache.js` uses `blob` in offline mode. Still not sure about `jr` though.

I'm not sure what the right place for this is but it would be helpful to document the reasoning for anything non-obvious. Maybe it's just in the PR description. That way if we make changes to the related code we can more easily reconsider the related CSP aspect.

files/nginx/common-headers.nginx.conf
lognaturel: Firefox: `Content Security Policy: Couldn’t process unknown directive ‘style-src-attr:’`

files/nginx/odk.conf.template
alxndrsn: I don't know all the reasons for the exceptions that enketo requires, but happy to add a mention for any that are known.

alxndrsn: @lognaturel

* example of `jr:` at https://sentry.io/organizations/getodk/issues/3650820395/
* another `blob` example at https://sentry.io/organizations/getodk/issues/3663159410/

files/nginx/common-headers.nginx.conf
lognaturel: I see that too when accessing test and online-only Enketo forms. I see the variant with the colon when I request an offline-capable form view: https://test.getodk.cloud/-/x/dVpkm8cSxy3AoHukfm0byhRXj1yxViN

lognaturel (approved):

PR #314, opened 2022-08-31
Referrer policy header

alxndrsn:
Worth considering while reviewing privacy- & security-related headers.  Ref https://developer.mozilla.org/en-US/docs/Web/Security/Referer_header:_privacy_and_security_concerns

E.g. this will marginally decrease the option of opportunistic credential stealing via basic auth dialog.

> When an external resource(e.g. http://foo.com/image.jpeg) is used as the source of an image tag, if the external resource returns a 401 response code and sets a WWW-Authenticate header then the browsers standard 'Basic authentication' dialogue will pop up.. on the [ODK] page.
> While this is standard (and expected) browser behavior it could confuse users and be used in phishing attacks.
> -- _https://jira.atlassian.com/browse/CONFSERVER-28932_

### DIFF
--- a/files/nginx/common-headers.nginx.conf
+++ b/files/nginx/common-headers.nginx.conf
@@ -2,6 +2,7 @@
 # which has a call to add_header.
 # See: https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header
 
+add_header Referrer-Policy same-origin;
 add_header Strict-Transport-Security "max-age=63072000" always;
 add_header X-Frame-Options "SAMEORIGIN";
 add_header X-Content-Type-Options nosniff;


files/nginx/common-headers.nginx.conf
```diff
@@ -2,6 +2,7 @@
 # which has a call to add_header.
 # See: https://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header
 
+add_header Referrer-Policy same-origin;
```
matthew-white, line 5: Adding this header sounds like a useful change to me! `same-origin` also seems reasonable. Out of curiosity, do you see an advantage to `same-origin` over `no-referrer`? It looks like the only way we use the Referer header in Backend is to redact it before sending information about the request to Sentry: https://github.com/getodk/central-backend/blob/87aedb0b58973d7454ddd3a543fad7547646bcdd/lib/external/sentry.js#L69-L71

alxndrsn: One thing I can think of is that `no-referrer` might making tracing certain errors more difficult for us.

Ref https://github.com/getodk/central-backend/blob/87aedb0b58973d7454ddd3a543fad7547646bcdd/lib/external/sentry.js#L69-L71, I think this is also scrubbing _third party_ URLs from Sentry reports, so if you link from https://nsfw.example.com -> https://something-nice.getodk.cloud, Sentry will never see `Referer: https://nsfw.example.com`

matthew-white (approved):

PR #316, opened 2022-09-01
sentry: set environment

alxndrsn:
see: https://docs.sentry.io/platforms/javascript/configuration/environments/

Related to: https://github.com/getodk/central-backend/issues/521, https://github.com/getodk/central-backend/issues/547

This is one axis on which we can filter Sentry issue reports.  Suggested values:

* development: dev.getodk.cloud
* staging: test.getodk.cloud
* production: everything else

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,6 +34,7 @@ services:
       - DOMAIN=${DOMAIN}
       - HTTPS_PORT=${HTTPS_PORT:-443}
       - SYSADMIN_EMAIL=${SYSADMIN_EMAIL}
+      - SENTRY_ENVIRONMENT=production
     command: [ "./wait-for-it.sh", "postgres:5432", "--", "./start-odk.sh" ]
     restart: always
     logging:


yanokwa:
I think this is a reasonable idea, but I don't think we should put it in the public docker-compose.yml because people will change it. It seems like we should just add it on dev/staging for our own use.

alxndrsn:
These values have been set on dev & test envs.

yanokwa:
Maybe I'm being dense or unreasonable here, but I'd like to not have this in the public docker-compose. Is that weird? I don't want self-hosters to change this and then we have weird environments in our Sentry. Basically everything but dev/test should be production, but not explicitly labeled as such.

matthew-white:
If we're already setting the username (getodk/central-backend#593), I'm wondering to what extent setting the environment is additionally helpful. Could we just filter for dev and test using their usernames? Or do things become easier by setting the environment?

alxndrsn:
> If we're already setting the username, I'm wondering to what extent setting the environment is additionally helpful.

Would it be helpful for controlling alert levels / automated triaging of issues?  E.g. you may want to be "paged" (in whatever form) for a _production_ incident, but never for one triggered in _staging_ or _dev_.

alxndrsn:
> instead of filtering for a production environment, we could filter out the usernames `test` and `dev`

@matthew-white that's possible, but for both browsing issues and configuring alerts, "environment" is a top-level filter, whereas configuring an exclusive username filter is more involved and error-prone, and would need to be done twice each time.

Would this PR be acceptable if the change was somewhere less user-accessible than `docker-compose.xml`?

matthew-white:
I could see how it might make some things easier to have the environment set. If there's a way to set it while addressing @yanokwa's concern, that sounds good!

alxndrsn:
@matthew-white Seems like `docker-compose.yml` has other changes in the non-prod envs, so would the following addition be suitable at https://github.com/getodk/central-backend/blob/master/lib/external/sentry.js#L120-L121:

```js
  environment: process.env.SENTRY_ENVIRONMENT || 'production',
```

?

matthew-white:
That sounds reasonable to me!

matthew-white:
Now that getodk/central-backend#601 is merged, I think we can close this PR. (Please reopen if I got that wrong!)

alxndrsn:
AFAICT, `SENTRY_ENVIRONMENT` has been unset in both dev & test.  Ref #547

yanokwa:
I broke this in the upgrade to v1.5.3 in Cloud. It's now fixed. The values I used:
```
dev.getodk.cloud: development
test.getodk.cloud: testing (not staging because maybe we'll setup a staging server?)
```

PR #318, opened 2022-09-07
new: updated node version to 16.17.0

sadiqkhoja:
new: updated node version to 16.17.0

### DIFF
--- a/nginx.dockerfile
+++ b/nginx.dockerfile
@@ -1,4 +1,4 @@
-FROM node:14.19.3 as intermediate
+FROM node:16.17.0 as intermediate
 
 COPY ./ ./
 RUN files/prebuild/write-version.sh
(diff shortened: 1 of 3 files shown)


nginx.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14.19.3 as intermediate
+FROM node:16.17.0 as intermediate
```
matthew-white, line 1: It sounds good to upgrade Node for `nginx` as well as `service`. Would you be up for upgrading Node in the `central-frontend` repository as well? (I'd also be happy to do that!)

matthew-white:
I think this PR intersects with #317, since that PR uses Node in a new place. If we ship #317 in the upcoming patch release, then we could rebase this PR on the `master` branch then.

PR #321, opened 2022-09-16
Update news item about Google backups

matthew-white:


yanokwa:
This has been a fun little design challenge. What if we put a future date? 

Screen Shot 2022-09-16 at 05 23 56

matthew-white:
I like the word "sunsetted"!

I feel unclear about whether the date of the news item should be the date of the announcement or the date of the event itself. But I don't have strong feelings, so I'd be happy to run with 2022/10/03.

Maybe a related question is whether we'll want to put the news item for the release of v1.5.3 above or below the news item about Google backups. If the news item about Google backups is dated 2022/10/03, then I think the news item about the release should go below it. But then is it easy to miss?

yanokwa:
You know, putting date of the announcement (2022/09/15), plus the ⚠️, plus "sunsetted" is pretty clear. Like if you have Google backups, it's pretty obvious you should click the link to find out more. 

Then we don't have to think about what the date means.

yanokwa (approved):
Ship it 🛳

PR #324, opened 2022-09-21
Update submodules for v1.5.3

matthew-white:
`server` is set to the tip of getodk/central-backend#594, which I'll merge into the `master` branch once I've resolved conflicts.


yanokwa (approved):

yanokwa:
Not sure how best to verify the client part of this PR. The server looks good.

matthew-white:
The client has been set to the tip of getodk/central-frontend#635, which has been merged into the `master` branch. That commit intentionally excludes https://github.com/getodk/central-frontend/commit/2b07bbf14f5f50764b5a00215b64630425001da2, which is the only Frontend commit that shouldn't go out with the patch release.

matthew-white:
I've updated the server to match the latest commit on getodk/central-backend#594.

yanokwa:
I believe you and I also believe @lognaturel  when she says it's low risk. 

For the record, it's hard to verify because the client PR you link to has 16 files and this PR has 34. My understanding is that those extra files are the files when the first PR's branch was made.

matthew-white:
Ah good point, I can see how that's confusing. I'm not quite sure what the answer is, but I'll think about that for future patch releases!

PR #327, opened 2022-09-23
openssl can fail, leaving an empty file and causing a reboot loop

yanokwa:
openssl can fail, leaving an empty file and causing a reboot loop since `/etc/dh/nginx.pem` cannot be found.

During an upgrade to Central v1.5.3, this is what the user saw when the server did not come up.
image

Logs showed it was an issue with `/etc/dh/nginx.pem`.
image

Not sure what happened, but it seems that for whatever reason, the openssl command failed and resulted in an empty file.

This PR checks for an empty file in all places where openssl is called. I confirmed the syntax is correct with shellcheck. I also wrap the mkdir command in case we ever use a white space in the name.

I confirmed this fix works on the server that had this issue.

### DIFF
--- a/files/nginx/odk-setup.sh
+++ b/files/nginx/odk-setup.sh
@@ -1,15 +1,17 @@
+#!/bin/bash
+
 DHPATH=/etc/dh/nginx.pem
-if [ ! -e "$DHPATH" ] && [ "$SSL_TYPE" != "upstream" ]
+if [ ! -s "$DHPATH" ] && [ "$SSL_TYPE" != "upstream" ]
 then
   echo "diffie hellman private key does not exist; creating.."
   openssl dhparam -out "$DHPATH" 2048
 fi
 
 SELFSIGN_BASEPATH=/etc/selfsign/live/$DOMAIN
-if [ "$SSL_TYPE" = "selfsign" ] && [ ! -e "$SELFSIGN_BASEPATH/privkey.pem" ]
+if [ "$SSL_TYPE" = "selfsign" ] && [ ! -s "$SELFSIGN_BASEPATH/privkey.pem" ]
 then
   echo "self-signed cert requested but does not exist; creating.. (this could take a while)"
-  mkdir -p $SELFSIGN_BASEPATH
+  mkdir -p "$SELFSIGN_BASEPATH"
   openssl req -x509 -newkey rsa:4086 \
     -subj "/C=XX/ST=XXXX/L=XXXX/O=XXXX/CN=localhost" \
     -keyout "$SELFSIGN_BASEPATH/privkey.pem" \


matthew-white:
@alxndrsn, if you have time, would you mind taking a look?

matthew-white:
I'm going to target the `release` branch with the thought that this change would go out with v2022.3.

files/nginx/odk-setup.sh
```diff
@@ -1,12 +1,12 @@
 DHPATH=/etc/dh/nginx.pem
-if [ ! -e "$DHPATH" ] && [ "$SSL_TYPE" != "upstream" ]
+if [[ (! -e "$DHPATH" || ! -s "$DHPATH" ) && "$SSL_TYPE" != "upstream" ]]
```
alxndrsn, line 2: is `-e` made redundant by `-s`?

From the bash manpage:
```
       -s file
              True if file exists and has a size greater than zero.
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
...
+
+#### Why is this the best possible solution? Were any other approaches considered?
+
+#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?
+
+#### Does this change require updates to documentation? If so, please file an issue [here](https://github.com/getodk/docs/issues/new) and include the link below.
+
+#### Before submitting this PR, please make sure you have:
+
+- [ ] verified that any code or assets from external sources are properly credited in comments.
```
alxndrsn, line 13: Changes seem unrelated to the PR title.

docs/news.html
```diff
@@ -8,15 +8,15 @@
...
+        ODK Central v1.5.3
       </a>
     </div>
     <div class="news-item">
-      <span class="item-date">2022/07/12</span>
-      <a href="https://forum.getodk.org/t/odk-central-v1-5/37882/3" target="_blank">
-        ODK Central v1.5.2
+      <span class="item-date">2022/09/15</span>
+      <a href="https://forum.getodk.org/t/backups-to-google-drive-from-central-will-stop-working-after-jan-31st/38895" target="_blank">
+        ⚠️ Google backups sunsetted ⚠️
```
alxndrsn, line 19: Changes seem unrelated to the PR title.

alxndrsn (commented):
> I confirmed the syntax is correct with shellcheck. I also wrap the mkdir command in case we ever use a white space in the name.

I get two shellcheck violations:

```
In odk-setup.sh line 1:
DHPATH=/etc/dh/nginx.pem
^-- SC2148: Tips depend on target shell and yours is unknown. Add a shebang.


In odk-setup.sh line 12:
  mkdir -p $SELFSIGN_BASEPATH
           ^----------------^ SC2086: Double quote to prevent globbing and word splitting.
```

Is there a reason not to use bash's `-e` and `-o pipefail` options?  This might give more robust handling of unforeseen failures of individual commands in this file.

yanokwa:
Sorry about the extra commits, I've cleaned that up and also fixed the redundant flag. The script should pass shellcheck.

As to `-e` and `-o pipefail`, maybe those can come as part of https://github.com/getodk/central/pull/328 so we can do that work at once on all scripts.

yanokwa:
I went for the narrow change.

I don't know offhand how to fail immediately here. In what I've seen, the container will restart endlessly when something goes wrong.

PR #329, opened 2022-09-30
fix: start correct number of node instances

sadiqkhoja:
Closes #

Tested locally

We can wait for pm2 team to fix the issue at their end. Ref: https://github.com/Unitech/pm2/issues/5450

NA

NA

- [X] verified that any code or assets from external sources are properly credited in comments.

### DIFF
--- a/files/service/pm2.config.js
+++ b/files/service/pm2.config.js
@@ -11,7 +11,10 @@ module.exports = {
     error_file: '/proc/1/fd/2',
 
     // per Unitech/pm2#2045 this resolves a conflict w node-config:
-    instance_var: 'INSTANCE_ID'
+    instance_var: 'INSTANCE_ID',
+
+    // --instances option from cli is not working in pm2 v5.2. Ref: Unitech/pm2/issues/5450
+    instances: process.env.WORKER_COUNT
   }]
 };
 
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -27,5 +27,5 @@ fi
 echo "using $WORKER_COUNT worker(s) based on available memory ($MEMTOT).."
 
 echo "starting server."
-pm2-runtime ./pm2.config.js --instances $WORKER_COUNT
+pm2-runtime ./pm2.config.js
 
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -11,7 +11,7 @@ COPY files/service/crontab /etc/cron.d/odk
 
 COPY server/package*.json ./
 RUN npm install --production
-RUN npm install pm2 -g
+RUN npm install pm2@5.2.0 -g
 
 COPY server/ ./
 COPY files/service/scripts/ ./


lognaturel:
> Does this mean that when we updated to the LTS OS as part of Central v1.5.2 on 7/12/2022 we ended up updating to pm2 5.2.0 and losing multiple process functionality?

Note from conversation with @sadiqkhoja: he tried `pm2` 5.1.2 and it doesn't work either. How far back did you go, @sadiqkhoja? We don't know for sure that multiple processes ever worked.

matthew-white:
It looks like we always install the latest version of `pm2`:

https://github.com/getodk/central/blob/da62679fe138eb89b42fcc2d9f61ae9beca9d67b/service.dockerfile#L14

How about we specify a particular version like we do with our other npm dependencies?

lognaturel:
@matthew-white Do you remember ever knowing for sure that multiple processes were running? I have a vague memory of running `htop` when I was doing the 2021-03 benchmarking and seeing node 4 times but I could very well be imagining things. I have no recollection of knowing one way or the other since then.

matthew-white:
I don't remember myself! One thing that I think is relevant is that I think we only moved to `pm2` with v1.3: b105743365e628303644aa4a0a9f2a986baea1e0

matthew-white:
Also, I was able to reproduce getodk/central-backend#482 on the QA server. It only took 10 concurrent requests to cause an issue, not 40.

lognaturel:
> able to reproduce https://github.com/getodk/central-backend/issues/482 on the QA server

That was in May so Central v1.4.2. Hmm! Will be good to have addressed. This might explain some things.

matthew-white:
I'm thinking that whenever we rebuild `service`, we bump to the latest version of `pm2` (regardless of the version of Central, as long as it's v1.3 or later). Does that sound right?

lognaturel:
> whenever we rebuild service, we bump to the latest version of pm2

Right, yes. So Central version doesn't matter, sorry.

sadiqkhoja:
Since we added pm2 in August 2021, I tried latest version of PM2 that was available then i.e. v5.0.3 and multiple instances are not created with it too

lognaturel:
@sadiqkhoja will pin the `pm2` version in this PR and then @matthew-white will do final review.

matthew-white:
It looks like there's a conflict with the `release` branch (I think with #334 specifically).

I'm also wondering whether we should move everything related to `pm2` to the `central-backend` repo. If `pm2` is our only npm dependency in `central`, I think it'll be easy to forget to upgrade it. We also won't get automated security alerts related to it. I don't think we should make that move as part of this PR, but I'm thinking of filing an issue.

matthew-white:
@sadiqkhoja, I was also wondering whether you've verified this change on a server already. That could also happen after the PR is merged, when it'll automatically be deployed to the staging server.

matthew-white:
> I'm also wondering whether we should move everything related to `pm2` to the `central-backend` repo.

I've filed an issue for this: #336

matthew-white (approved):

sadiqkhoja:
I'm thinking about any implications of this on session authentication. Do processes share memory or we have sticky sessions?

matthew-white:
Sessions are persisted in the database, not in app memory. Each session is associated with a token, which Frontend sends to Backend in either a header or a cookie. Does that sound like it should work alright?

sadiqkhoja:
then we are good

Issue #330, opened 2022-10-07
Add shebangs and shell options


matthew-white:
shellcheck seems to recommend that we add shebangs to shell scripts. Doing so would make it clear what shell we're targeting.

@alxndrsn has also [brought up](https://github.com/getodk/central/pull/327#pullrequestreview-1119630281) the idea of using bash's `-e` and `-o pipefail` options. When we add shebangs, we could consider adding shell options as well.

matthew-white:
Closed by #331.

PR #331, opened 2022-10-07
Add shebangs to all shell scripts

alxndrsn:
Closes https://github.com/getodk/central/issues/330

Different shebangs are used because these scripts are currently run in different contexts, specifically:

* `files/enketo/start-enketo.sh`: triggered in `enketo.dockerfile` using [Docker's `CMD`](https://docs.docker.com/engine/reference/builder/#cmd) in _shell_ form; "If you use the shell form of the `CMD`, then the `<command>` will execute in `/bin/sh -c`"
* `files/nginx/odk-setup.sh`: triggered in `nginx.dockerfile`, explicitly using `/bin/bash` there
* `files/service/scripts/start-odk.sh`: triggered via `exec` in `./wait-for-it.sh`, so inherits the shebang from `./wait-for-it.sh`
* `files/enketo/generate-secrets.sh`: triggered by [`docker-compose.yaml`'s `command`](https://docs.docker.com/compose/compose-file/#command), which "overrides...Dockerfile’s `CMD`". so presumably follows the same rules

### DIFF
--- a/files/enketo/generate-secrets.sh
+++ b/files/enketo/generate-secrets.sh
@@ -1,3 +1,4 @@
+#!/bin/sh
 
 if [ ! -f /etc/secrets/enketo-secret ]; then
   LC_ALL=C tr -dc '[:alnum:]' < /dev/urandom | head -c64 > /etc/secrets/enketo-secret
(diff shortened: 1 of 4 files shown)


yanokwa:
Why use `#!/usr/bin/env bash` for `start-odk.sh` and `wait-for-it.sh`. Or rather, seems like we should use /bin/bash and be consistent. I really like consistency.

alxndrsn:
@yanokwa I agree, happy to follow up with that later. This PR is aiming to add explicit shebangs which declare what is already happening.

yanokwa (approved):
LGTM

PR #332, opened 2022-10-07
start-enketo: simplify config generation

alxndrsn:
The original code looks like it's doing something clever (e.g. protecting secrets from leaking into the `pm2` process?), but I can't work out what it is...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
       - checkout
 
       - run: sudo apt install shellcheck
-      - run: cat <(git grep -El '^#!.*sh\b') <(git ls-files | grep -E '.sh$') | sort -u | grep -v '/wait-for-it.sh$' | xargs shellcheck
+      - run: cat <(git grep -El '^#!.*sh\b') <(git ls-files | grep -E '.sh$') | sort -u | grep -v '/wait-for-it.sh$' | xargs shellcheck --exclude=SC2016
 
       - run: git submodule update -i
 
--- a/files/enketo/start-enketo.sh
+++ b/files/enketo/start-enketo.sh
@@ -2,7 +2,13 @@
 
 CONFIG_PATH=${ENKETO_SRC_DIR}/config/config.json
 echo "generating enketo configuration.."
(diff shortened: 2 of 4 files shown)


lognaturel:
@issa-tseng doesn't think there was a reason for the original structure.

matthew-white:
@yanokwa, do you have time to take a look at this? Also happy to pick a different reviewer if that'd be better.

yanokwa (approved):
LGTM. I confirmed that envsubst does the right substitution (at least on my Mac). I also confirmed that variables with spaces work.

yanokwa (changes requested):
I think the reason for `bash -c` is probably to guarantee that we don't clobber any existing env variables (because the names we use are generic) or leave those variables behind for some rogue process to exfiltrate.

We should leave the code as-is, but add a comment explaining why we have this.

alxndrsn:
> guarantee that we don't clobber any existing env variables (because the names we use are generic) or leave those variables behind for some rogue process to exfiltrate.

I think that is already covered... E.g.

```bash
KEEP_ME=1
echo "KEEP_ME (original):    $KEEP_ME"
echo "DONT_LEAK:             $DONT_LEAK"

echo '---'

KEEP_ME=2 \
DONT_LEAK=X \
envsubst '$KEEP_ME$DONT_LEAK' <<<'KEEP_ME (substituted): $KEEP_ME
DONT_LEAK:             $DONT_LEAK'

echo '---'

echo "KEEP_ME (final):       $KEEP_ME"
echo "DONT_LEAK:             $DONT_LEAK"
```

will print:

```shell
KEEP_ME (original):    1
DONT_LEAK:             
---
KEEP_ME (substituted): 2
DONT_LEAK:             X
---
KEEP_ME (final):       1
DONT_LEAK:             
```

yanokwa:
I stand corrected. 

I propose we apply this change to the following and merge at once.
files/service/scripts/start-odk.sh
files/nginx/odk-setup.sh

lognaturel:
@yanokwa @alxndrsn the todo here is to apply the change more broadly?

.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ jobs:
       - checkout
 
       - run: sudo apt install shellcheck
-      - run: cat <(git grep -El '^#!.*sh\b') <(git ls-files | grep -E '.sh$') | sort -u | grep -v '/wait-for-it.sh$' | xargs shellcheck
+      - run: cat <(git grep -El '^#!.*sh\b') <(git ls-files | grep -E '.sh$') | sort -u | grep -v '/wait-for-it.sh$' | xargs shellcheck --exclude=SC2016
```
alxndrsn, line 11: It might be nice to have a `scripts` directory for storing things like this.  This would also make running the command locally more convenient.

alxndrsn:
@lognaturel @yanokwa I've applied the change to the requested files.

lognaturel (approved):

lognaturel:
Verified on dev server. This will make adding more env vars a bit more palatable.

PR #333, opened 2022-10-07
service: use npm ci

alxndrsn:
> This command is similar to npm install, except it's meant to be used
> in automated environments such as test platforms, continuous
> integration, and deployment -- or any situation where you want to
> make sure you're doing a clean install of your dependencies.
>
> - https://docs.npmjs.com/cli/v8/commands/npm-ci/


matthew-white (commented):
Seems like a reasonable change! It looks like there's a conflict with #334, which I just merged.

matthew-white (approved):

Issue #298, opened 2022-06-27
Central v2022.3 release


matthew-white:
The v2022.3 release will add datasets that let users flow data between forms. Our initial goal is to automate baseline/follow-up style workflows. For example, going from registration to a dataset for use in a follow-up as described in the [encounters with known entities](https://docs.getodk.org/data-collector-workflows/#encounters-with-known-entities) section of the data collector workflows doc.

Please take a look at the details in [this document](https://docs.google.com/document/d/1naxyzFU8EjcWKfYzE_mEF5cY38l1QRNUZ_0VW2Gp-r0/edit?usp=sharing). Comments and feedback may be left there, or here.

PR #337, opened 2022-10-12
Release v2022.3

matthew-white:
This PR prepares the release of v2022.3 (closing #298). It only contains changes from other PRs that have already been approved and merged.


files/nginx/odk.conf.template
```diff
@@ -58,5 +72,9 @@ server {
       add_header Cache-Control no-cache;
     }
   }
+
+  location /csp-report {
+    proxy_pass https://o130137.ingest.sentry.io/api/1298632/security/?sentry_key=3cf75f54983e473da6bd07daddf0d2ee;
```
matthew-white, line 77: We should remember to update our [docs about Sentry](https://docs.getodk.org/central-install-digital-ocean/#disabling-or-customizing-sentry) to mention how to disable or customize this reporting.

matthew-white:
With the release of the patch v1.5.4, I've merged in the `master` branch so that CircleCI will continue to pass on this branch.

matthew-white:
@lognaturel, I've tagged you for review since I think this PR is complete other than when we update the submodules right before release. The PR shouldn't need a deep review, as it should only include changes from other PRs and merge commits. Let me know if it'd be better for me to find someone else to tag for review!

lognaturel:
I should be able to do it! I can’t remember the sequence. You’ll update sub modules on here, right? So I should wait until that’s done?

matthew-white:
I think it's OK to ✅ now. On release day, we'll update the submodules on this branch, and there may be other changes if something comes up during regression testing. But it's nice to have the release PR reviewed before release day so there are fewer things to do/coordinate on the day itself. The ✅ on this PR mostly feels pro forma to me, since everything in here should already have been approved elsewhere.

lognaturel (approved):
Quickly verified that I remember all commits and that nothing seems to be missing.

PR #338, opened 2022-10-12
Fix sentry version copy

### LINKED ISSUE #335: Error accessing sentry-versions/
As of #317, `central` is supposed to output version information to a directory named sentry-versions. It then passes that information to Backend, which sends it to Sentry. However, I deployed #317 to the QA server along with getodk/central-backend#637, and it looks like something isn't working about the current setup. You can see the issue in this Sentry error: https://sentry.io/organizations/getodk/issues/3663423953/. There, the `version.central` and `version.client` tags are marked invalid. When you hover over the warning icon, it says, "Expected a non-empty value". The release also appears not to be set (`SENTRY_RELEASE`).

When I looked at the `service` logs, I saw:

```
generating local service configuration..
cat: sentry-versions/server: No such file or directory
cat: sentry-versions/central: No such file or directory
cat: sentry-versions/client: No such file or directory
```

@alxndrsn, would you mind taking a look at this?

alxndrsn:
Closes #335

### DIFF
--- a/files/enketo/config.json.template
+++ b/files/enketo/config.json.template
@@ -34,5 +34,6 @@
     "support": {
         "email": "support@getodk.org"
     },
-    "text field character limit": 1000000
+    "text field character limit": 1000000,
+    "exclude non-relevant": true
 }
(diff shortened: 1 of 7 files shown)


matthew-white (approved):
I deployed this to the QA server, and version information is now being passed to Sentry! 🎉 Example: https://sentry.io/organizations/getodk/issues/3663423953/

PR #339, opened 2022-10-21
Wait for it licence

alxndrsn:
The `wait-for-it.sh` script was introduced in fcad46fcd38fb67ecb0cb2631fb66b2904f582ed.

It looks like the version at https://github.com/vishnubob/wait-for-it/blob/8ed92e8cab83cfed76ff012ed4a36cef74b28096/wait-for-it.sh, whose licence is MIT and includes the clause:

> The above copyright notice and this permission notice shall be included in all
> copies or substantial portions of the Software.

See: https://github.com/vishnubob/wait-for-it/blob/8ed92e8cab83cfed76ff012ed4a36cef74b28096/LICENSE.

### DIFF
--- a/files/service/scripts/wait-for-it.sh
+++ b/files/service/scripts/wait-for-it.sh
@@ -1,5 +1,28 @@
 #!/usr/bin/env bash
 #   Use this script to test if a given TCP host/port are available
+#
+# Source: https://github.com/vishnubob/wait-for-it
+#
+# The MIT License (MIT)
+# Copyright (c) 2016 Giles Hall
+#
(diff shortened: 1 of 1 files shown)


yanokwa (approved):
MIT License is fine. If we have more of these, maybe we add a LICENSE.md file like [Collect](https://github.com/getodk/collect/blob/master/LICENSE.md) and put it there.

alxndrsn:
> MIT License is fine.

I don't think this is a choice...

yanokwa:
There's no choice as far as whether we have to include the MIT license, we do. The choice is whether we can/should include MIT code in an Apache project. We can and we do.

PR #340, opened 2022-10-21
scripts: fail if cd fails

alxndrsn:
These changes resolve the shellcheck violation:

    cd /usr/odk
    ^---------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
    
An alternative would be to follow `shellcheck`'s suggestion to add `|| exit` to the `cd` invocations, but adding `-e` will also protect against future changes and seems like a reasonable default for scripts:

```
-e errexit       If not interactive, exit immediately if any untested command fails.  The exit
                 status of a command is considered to be explicitly tested if the command is
                 used to control an if, elif, while, or until; or if the command is the left
                 hand operand of an “&&” or “||” operator.
```

`-u` () is a bonus, which also seems like a sensible default for shell scripts in this repo:

```
-u nounset       Write a message to standard error when attempting to expand a variable that
                 is not set, and if the shell is not interactive, exit immediately.
```

### DIFF
--- a/files/service/scripts/purge-forms.sh
+++ b/files/service/scripts/purge-forms.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -eu
 
 cd /usr/odk
 /usr/local/bin/node lib/bin/purge-forms.js >/proc/1/fd/1 2>/proc/1/fd/2
--- a/files/service/scripts/reap-sessions.sh
+++ b/files/service/scripts/reap-sessions.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -eu
 
 cd /usr/odk
 /usr/local/bin/node lib/bin/reap-sessions.js >/proc/1/fd/1 2>/proc/1/fd/2
--- a/files/service/scripts/run-analytics.sh
+++ b/files/service/scripts/run-analytics.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -eu
 
 cd /usr/odk
 /usr/local/bin/node lib/bin/run-analytics.js >/proc/1/fd/1 2>/proc/1/fd/2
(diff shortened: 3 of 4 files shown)


files/service/scripts/purge-forms.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -eu
```
matthew-white, line 1: Do `-e` and `-u` definitely work with /bin/sh as well as /bin/bash? Also, we want to move to /bin/bash at some point, right? Would it be worth it to move to /bin/bash as part of this PR?

matthew-white:
<blockquote>

```
-e errexit       If not interactive, exit immediately if any untested command fails.  The exit
                 ...
```
</blockquote>

Where is this text from out of curiosity?

alxndrsn:
@matthew-white 

> Do `-e` and `-u` definitely work with /bin/sh as well as /bin/bash?

> Where is this text from out of curiosity?

I think the default shell on the `service` image is `dash`:

```shell
$ docker exec -it central_service_1 sh -c 'readlink $(which sh)'
dash
```

The text is from the `dash` man pages, so I think the options are likely to work.

There are no man pages available on `service`, but you may have them locally (`man dash`), or check e.g. https://linux.die.net/man/1/dash.

> Also, we want to move to /bin/bash at some point, right? Would it be worth it to move to /bin/bash as part of this PR?

If that's the plan, can we keep that discussion separate?

matthew-white (approved):

matthew-white:
> > Also, we want to move to /bin/bash at some point, right? Would it be worth it to move to /bin/bash as part of this PR?
> 
> If that's the plan, can we keep that discussion separate?

Definitely, and I'm not sure that that's even necessarily the plan.

matthew-white:
@alxndrsn, is this ready to be merged?

PR #343, opened 2022-10-24
Upgrade postgres

alxndrsn:
Related PRs:
* https://github.com/getodk/docs/pull/1573

There are automated tests for this upgrade at https://github.com/alxndrsn/odk-central-postgres-migration-tester/.  These may find a permanent home in this repo.


docker-compose.yml
```diff
@@ -1,9 +1,13 @@
 version: "3"
 services:
   postgres:
-    image: "postgres:9.6"
+    build:
+      context: .
+      dockerfile: postgres.dockerfile
     volumes:
       - /var/lib/postgresql/data
+      # TODO here's an opportunity to make this a non-anonymouse volume...
```
alxndrsn, line 9: :mouse:

matthew-white:
While merging #344, I noticed this line referencing 9.6, which I'm thinking we'll want to change in this PR:

https://github.com/getodk/central/blob/2eea364384e3f978a04c89ed2d8b262e2d0bdba0/service.dockerfile#L8

Let me know if it'd be useful for me to pull #344 into the `next` branch! Right now that commit is only in `master`.

alxndrsn:
@matthew-white 

> Let me know if it'd be useful for me to pull #344 into the `next` branch! Right now that commit is only in `master`.

That would be very helpful, thanks!  Will #344 be needed in `next` anyway?

postgres-migrate.dockerfile
```diff
@@ -0,0 +1,14 @@
+# see: https://github.com/tianon/docker-postgres-upgrade/blob/master/9.6-to-14/Dockerfile
+FROM tianon/postgres-upgrade:9.6-to-14
+
+# This file is required to encourage human validation of the process.
+# It's expected it will be provided by the sysadmin performing the upgrade.
```
yanokwa, line 5: > And if it's not provided, this container will not build and thus the upgrade cannot proceed.

Or something...

matthew-white (commented):
I've taken a very high-level look at the PR. I can't say I understand with any detail how the flow works here (how the different files relate to each other). However, nothing stood out to me as being an issue. @alxndrsn and I are meeting later today, so I may gain a better understanding then. It's clear that a lot of thought and work has been put into this! 🌟

I see that we're checking disk space. Are there any limitations on memory? Is a 1 GB machine capable of completing the upgrade?

Can we update the central-backend CircleCI so that it uses Postgres 14? If that's what users will generally be using going forward, I think we should use that in CircleCI. It'd also be good to validate that all tests continue to pass. Though maybe that validation was already done in getodk/central-backend#644?

We have a QA server running on DigitalOcean that doesn't use an external database. Should we try upgrading that? And then should we ask the QA team to do a bit of regression testing on it?

We're removing Google Drive backups, but we will continue to support direct backups (/v1/backup). Will those work in Postgres 14?

I'm noticing that the `server` and `client` submodules were changed in this PR, but I don't think we want to merge those changes into the `next` branch.

Presumably this won't be our last Postgres upgrade. In some hopefully quite distant future, we'll be having users upgrade from Postgres 14 to some other version of Postgres. Is there anything that we could do now to ease that upgrade in the future? The next Postgres upgrade will be more straightforward, right?

check-available-space
```diff
@@ -0,0 +1,49 @@
...
+if [[ "$volume" = "" ]]; then
+  log "!!!"
+  log "!!! Volume not found for container: $containerName !!!"
+  log "!!!"
+  log "!!! Cannot continue !!!"
+  log "!!!"
+  exit 1
+fi
+mountpoint="$(docker volume inspect --format='{{.Mountpoint}}' "$volume")"
+used="$(du -s "$mountpoint" | cut -f1)"
```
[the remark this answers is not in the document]
matthew-white: From chatting with @alxndrsn: it's unclear whether Postgres 14 would require more space than Postgres 9. It seems within the realm of possibility. It's hard to know how to account for that though. 1.05 is not exactly a scientific choice.

We may have discussed this today, but what happens if a user attempts the upgrade when they don't have enough space? Nothing so problematic, I assume?

matthew-white:
Notes from meeting with @alxndrsn:

> Are there any limitations on memory? Is a 1 GB machine capable of completing the upgrade?

I believe that there are still 1 GB machines in the wild running Central, though possibly now we recommend at least 2 GB. We're going to try the upgrade on a QA server. Before we try the upgrade, we will downsize the machine to 1 GB.

> Can we update the central-backend CircleCI so that it uses Postgres 14?

I think @alxndrsn is going to make this change at some point? Though it's more likely that something would pass on 14 and fail on 9.6 than vice versa (for example, using new syntax available in 14 but not 9.6).

Are we requiring users on 9.6 to upgrade? Or are we more so providing a path forward to 14 and encouraging an upgrade?

> We have a QA server running on DigitalOcean that doesn't use an external database. Should we try upgrading that? And then should we ask the QA team to do a bit of regression testing on it?

We're planning to do this. I'm not sure how much regression testing we would need to feel confident. I'm thinking probably not a full pass.

> We're removing Google Drive backups, but we will continue to support direct backups (/v1/backup). Will those work in Postgres 14?

This hasn't been verified, but it seems like it should continue working. We use `pg_dump` for that, calling it in Backend like this:

```
pg_dump -j 4 -F d -f ${directory} -h ${dbConfig.host} -U ${dbConfig.user} ${dbConfig.database}
```

It looks like those flags are still all supported.

This is something that we should verify during regression testing.

> Presumably this won't be our last Postgres upgrade. In some hopefully quite distant future, we'll be having users upgrade from Postgres 14 to some other version of Postgres. Is there anything that we could do now to ease that upgrade in the future? The next Postgres upgrade will be more straightforward, right?

The next Postgres upgrade will be well into the future: 3 years from now. We'll wait until close to when Postgres 14 sunsets in order to minimize how often users need to upgrade their Postgres. By then, it will hopefully be safe to assume that no one is on 9.6 anymore. That means that we probably won't need three Docker services (for 9.6 + 14 + future latest), just two (for 14 + future latest). In other words, we don't anticipate a future in which a user will need to leapfrog from 9.6 to a version past 14.

One possibility that was brought up was using the name `latest-postgres` in some places rather than `postgres14` in case that means that fewer things will need to be renamed during the next upgrade. But all in all, it's more important to get this merged than to think through the next upgrade. We know that the next upgrade will be easier than this one because we won't have to do so much wrangling around the anonymous volume.

matthew-white:
Another thought I had was whether we should consider doing some heavy-handled prompting. Something like:

>  Before proceeding, please confirm that you have created a backup by typing the words: I have a backup

Maybe that's too much, and we should trust folks to be responsible. But I really don't want someone to try upgrading without a backup!

lognaturel:
> Before proceeding, please confirm that you have created a backup by typing the words: I have a backup

That's what the `allow-postgres14-upgrade` file is for! As far as we could figure out there's no way to do the kind of interactive prompting you're suggesting.

docker-compose.yml
```diff
@@ -1,14 +1,32 @@
 version: "3"
 services:
-  postgres:
-    image: "postgres:9.6"
+  postgres14:
+    build:
+      context: .
+      dockerfile: postgres14.dockerfile
```
lognaturel, line 6: Should we now consider 14.7? 14.7-bullseye? @yanokwa

docker-compose.yml
```diff
@@ -1,14 +1,32 @@
...
+  postgres:
+    # This service upgrades from postgres 9.6 to 14.
+    # The legacy name must be maintained to allow access to the anonymous volume.
+    build:
+      context: .
+      dockerfile: postgres_upgrade.dockerfile
+    volumes:
+      - /var/lib/postgresql/data
+      - postgres14:/var/lib/postgresql/14
+      # TODO this can potentially run out of space too!!
```
lognaturel, line 23: Is there something left to do here?

postgres_upgrade.dockerfile
```diff
@@ -0,0 +1,17 @@
+# see: https://github.com/tianon/docker-postgres-upgrade/blob/master/9.6-to-14/Dockerfile
+FROM tianon/postgres-upgrade:9.6-to-14
+
+# This file is required to encourage human validation of the process.
+# It's expected it will be provided by the sysadmin performing the upgrade.
+# Docker build will fail if this file is missing.
+# TODO this file may stick around for a while; maybe we should call this simply `use-postgres14`?
+COPY allow-postgres14-upgrade .
```
lognaturel, line 8: Can we make this `postgres14-upgrade/allow-postgres14-upgrade` instead to keep things neat? We expect people will copy from the instructions anyway. We should also add postgres14-upgrade to `.gitignore`

files/postgres/upgrade-postgres.sh
```diff
@@ -0,0 +1,130 @@
...
+log() {
+  echo "$(TZ=GMT date) [$logPrefix] $*"
+}
+
+log "Checking for existing upgrade marker file..."
+if [[ -f "$flag_upgradeCompletedOk" ]]; then
+  log "Upgrade has been run previously."
+
+  if [[ -f "$flag_deleteOldData_internal" ]]; then
+    # FIXME see if we can easily do a data sanity check of old vs new db, e.g. count the number of submissions, blobs, projects and check that new numbers are >= old numbers
```
lognaturel, line 21: What's the conclusion on this?

postgres14.dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM postgres:14
```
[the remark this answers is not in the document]
lognaturel: I deleted a comment I'd made about using 14.7 or 14.7-bullseye. I think using 14 and letting that pick the latest point release and Ubuntu is safe and most meets our goals of getting folks on latest versions.

postgres_upgrade.dockerfile
```diff
@@ -0,0 +1,17 @@
+# see: https://github.com/tianon/docker-postgres-upgrade/blob/master/9.6-to-14/Dockerfile
+FROM tianon/postgres-upgrade:9.6-to-14
+
+# This file is required to encourage human validation of the process.
+# It's expected it will be provided by the sysadmin performing the upgrade.
+# Docker build will fail if this file is missing.
+# TODO this file may stick around for a while; maybe we should call this simply `use-postgres14`?
```
lognaturel, line 7: Can we make this postgres14-upgrade/allow-postgres14-upgrade instead to keep things neat? We expect people will copy from the instructions anyway. We should also add postgres14-upgrade to .gitignore I expect these will stay there forever.

docker-compose.yml
alxndrsn: TODO remove this TODO

postgres_upgrade.dockerfile
lognaturel: No, the `postgres14-upgrade` directory is created by the new container. But instead, `files/postgres14/` is a good place for it.

Both need to go into `.gitignore`

lognaturel:
TODO:
- Space script isn't consistently working, what's up with that?
- How risky is the postgres upgrade? Is it atomic-ish? Probably we don't need to do an extra sanity check.
- Add new files to .gitignore
- Move the allow file

files/postgres/upgrade-postgres.sh
alxndrsn: Will remove FIXME

lognaturel:
TODO from @matthew-white:
- Is a 1 GB machine capable of completing the upgrade?
- Update the central-backend CircleCI so that it uses Postgres 14
- Verify direct backups

lognaturel (approved):

PR #353, opened 2022-11-16
service.dockerfile: don't use apt-key

### LINKED ISSUE #352: Bake postgres GPG key into images?
When building `service.dockerfile`, the following message appears on stderr:
```
Warning: apt-key output should not be parsed (stdout is not a terminal)
```
It looks like it's triggered by:
https://github.com/getodk/central/blob/2eea364384e3f978a04c89ed2d8b262e2d0bdba0/service.dockerfile#L5-L8
The man page for `apt-key` reads:
```
add filename
    Add a new key to the list of trusted keys. The key is read from the 
    filename given with the parameter filename or if the filename is - from
    standard input.

    It is critical that keys added manually via apt-key are verified to
    belong to the owner of the repositories they claim to be for otherwise
    the apt-secure(8) infrastructure is completely undermined.

    Note: Instead of using this command a keyring should be placed directly
    in the /etc/apt/trusted.gpg.d/ directory with a descriptive name and 
    either "gpg" or "asc" as file extension.
```
Perhaps the `apt-key add` call can be replaced with:

1. copying https://www.postgresql.org/media/keys/ACCC4CF8.asc into e.g. `files/service/postgres-apt-key-ACCC4CF8.asc`
2. `COPY files/service/postgres-apt-key-ACCC4CF8.asc /etc/apt/trusted.gpg.d/`

alxndrsn:
Closes #352

### DIFF
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -13,7 +13,7 @@ FROM node:16.17.0
 WORKDIR /usr/odk
 
 RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODENAME=\K\w+' /etc/os-release)-pgdg main" | tee /etc/apt/sources.list.d/pgdg.list; \
-  wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -; \
+  curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | gpg --dearmor > /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg; \
   apt-get update; \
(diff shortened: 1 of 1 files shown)


alxndrsn:
PR will need updating if https://github.com/getodk/central/pull/354 is merged first.

matthew-white:
@yanokwa, do you have time to take a look at this one? Seems like this would be a nice one to get into v2022.3.

yanokwa (approved):
LGTM. 

I was convinced by the fact that these instructions are in quickstart wiki as noted at https://github.com/getodk/central/issues/352#issuecomment-1316569825

PR #354, opened 2022-11-16
Run linked commands with &&

alxndrsn:
It seems risky to have multiple statements in a single `RUN` instruction.

In this particular example, it's likely that the final step (`apt-get install`) will fail if one or more earlier steps failed.  But replacing `;` with ` && ` means that:

* build will fail faster if an early step fails
* if this pattern is copied elsewhere, we might avoid obscure bugs


yanokwa:
Seems to happen in enketo.dockerfile and nginx.dockerfile. Can you make those changes there too?

I also checked *.sh and didn't find any, but double-check.

matthew-white:
We didn't get the chance to review this one before regression testing, so I'd lean toward waiting until v2023.1 to merge this.

lognaturel:
Ready to merge after resolving conflicts?

Issue #349, opened 2022-11-16
Update pyxform-http for v2023.1


matthew-white:

matthew-white:
Closed by #377.

PR #355, opened 2022-11-21
Upgrade pyxform to v1.11.1 with entity creation support

lognaturel:
Addresses part of #349

I've upgraded `pyxform` on the staging and dev servers and verified that it works as intended. I believe that without getting this in, it will be reverted on staging if anything else gets merged to `next`.

Given the review and verification we've done, I am fairly confident this is the version we'll release with.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -60,7 +60,7 @@ services:
       options:
         max-file: "30"
   pyxform:
-    image: 'ghcr.io/getodk/pyxform-http:v1.10.1.1'
+    image: 'ghcr.io/getodk/pyxform-http:v1.11.1'
     restart: always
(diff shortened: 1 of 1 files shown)


matthew-white (approved):

matthew-white:
> Addresses part of #349

That one's actually for the next release. (For the next release, I've filed issues for the sort of maintenance that we've previously tracked in the release criteria.)

matthew-white:
Did this new version of pyxform make any changes not related to entity creation support? Is there anything to mention for the forum announcement?

lognaturel:
One small change that I don’t think needs to be mentioned: https://github.com/XLSForm/pyxform/pull/621

PR #359, opened 2022-11-29
service: npm ci: suppress noise

alxndrsn:
Follow on from #358 ~~(please wait for that to be resolved before considering this)~~.


matthew-white:
This seems like a nice change to me. At some point, could you make a similar change for the Frontend `npm install` (in files/prebuild/build-frontend.sh I believe)?

matthew-white (approved):

PR #360, opened 2022-11-30
Update Enketo

lognaturel:

EE 4.1.4 is running on the QA server. 

We tried to get an additional improvement to number inputs in but it’s not complete after all. 4.1.4 is the safest version to release. 

This release mostly has bug fixes and should be relatively safe. 

No. 

### DIFF
--- a/enketo.dockerfile
+++ b/enketo.dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/enketo/enketo-express:4.1.2
+FROM ghcr.io/enketo/enketo-express:4.1.4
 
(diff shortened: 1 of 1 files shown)


matthew-white (approved):

matthew-white:
@lognaturel, are there any changes since 4.1.2 that we should mention in the forum announcement?

lognaturel:
I don’t think any really need to be mentioned. A few targeted bug fixes: https://github.com/enketo/enketo-express/releases/tag/4.1.3

PR #365, opened 2022-12-06
Update news for v2022.3

matthew-white:


docs/news.html
```diff
@@ -8,9 +8,9 @@
   </head>
   <body>
     <div class="news-item">
-      <span class="item-date">2022/11/08</span>
-      <a href="https://forum.getodk.org/t/odk-central-v1-5/37882/5" target="_blank">
-        ODK Central v1.5.4
```
matthew-white, line 13: I replaced the news item about v1.5.4 instead of the one about Google backups sunsetting, since I think it's a good idea to continue displaying the announcement about backups (until v2023.1).

PR #371, opened 2022-12-16
Update fathom domain name

yanokwa:


lognaturel (approved):

lognaturel:
I think this should get merged into `next`? Feels like it wouldn't be a big deal to go to master but it's weird to have different installs on different hashes.

matthew-white:
As things are set up now, I think it's inevitable that different installs will have different hashes. Any change to the news or to docs like the readme will result in a different hash. So I don't think it'd cause issues to merge this PR into `master`. (I have a small change to the readme coming up that I similarly plan to merge to `master`.) I'd also love to continue brainstorming how to improve this setup in general, which I don't think is ideal, but as things are now, I think it's OK to merge to `master`.

PR #372, opened 2022-12-20
Release v2023.1

matthew-white:
This PR prepares the release of v2023.1. It only contains changes from other PRs that have already been approved and merged.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -60,7 +60,7 @@ services:
       options:
         max-file: "30"
   pyxform:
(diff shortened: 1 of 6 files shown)


lognaturel:
@matthew-white What do you think about keeping this as a draft until it's complete?

matthew-white:
I think this PR now has all the commits that we plan to ship, with the exception of updating submodules. Given that submodules will be updated here only after being approved in their own PR, I think we could mark this PR as ready for review and approve it. @lognaturel, what do you think about that process?

yanokwa (approved):
Submodules match and files look like the ones we need to update.

PR #374, opened 2022-12-21
fix: Name the volume used in service postgres in docker-compose for d…

### LINKED ISSUE #373: Bug: docker-compose.yml is using anonymous volumes which are not mounted on new runs
More on this: https://github.com/docker/compose/issues/7444

Each time we do a `docker-compose up` new anonymous docker volume is generated and mounted, the older volume is not deleted but left as a dangling volume in the system. 

https://github.com/tushar5526/central/blob/master/docker-compose.yml#L6

Fix: To name the anonymous volume

tushar5526:
…ata to persist between compose up and downs

Closes #373 

Tested locally

This is a severe issue as this could lead to data loss between subsequent docker-compose ups and downs. 

Users who have their data already mounted on a anonymous volume will have to migrate their data from anonymous volume to new named volume or to a mount folder. 

no

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ services:
   postgres:
     image: "postgres:9.6"
     volumes:
-      - /var/lib/postgresql/data
+      - db:/var/lib/postgresql/data
     environment:
       POSTGRES_USER: odk
       POSTGRES_PASSWORD: odk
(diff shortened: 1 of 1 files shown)


lognaturel:
Closing because it doesn't address folks who are upgrading installs with existing anonymous volumes. We are working to address this in the postgres version update and will tag @tushar5526 on that work when it's close to ready.

PR #379, opened 2023-02-02
Update news for v2023.1

matthew-white:
@ktuite will merge this PR after merging #372. (Hence this PR branches off and targets the `master` branch, not `next`.) If we don't release on Friday, then @ktuite will update the date of the news item.

### DIFF
--- a/docs/news.html
+++ b/docs/news.html
@@ -8,15 +8,15 @@
   </head>
   <body>
     <div class="news-item">
-      <span class="item-date">2022/12/12</span>
-      <a href="https://forum.getodk.org/t/odk-central-v2022-3/39787/2" target="_blank">
-        ODK Central v2022.3.1
+      <span class="item-date">2023/02/03</span>
+      <a href="https://forum.getodk.org/t/odk-central-v2023-1/40267" target="_blank">
+        ODK Central v2023.1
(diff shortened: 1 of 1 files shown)


docs/news.html
```diff
@@ -8,15 +8,15 @@
...
-        ODK Central v2022.3.1
+      <span class="item-date">2023/02/03</span>
+      <a href="https://forum.getodk.org/t/odk-central-v2023-1/40267" target="_blank">
+        ODK Central v2023.1
       </a>
     </div>
     <div class="news-item">
-      <span class="item-date">2022/09/15</span>
-      <a href="https://forum.getodk.org/t/backups-to-google-drive-from-central-will-stop-working-after-jan-31st/38895" target="_blank">
-        ⚠️ Google backups sunsetted ⚠️
```
matthew-white, line 19: Now that we are fully sunsetting Google Drive backups, I think we should remove this item. However, I could also see the argument for keeping it for a little longer. Once the Postgres upgrade is ready, we might want to keep its associated news item pinned in the news for a little bit.

yanokwa (approved):
Confirmed the links go to the right things. I'm OK dropping the Google Drive backups link.

PR #386, opened 2023-02-15
Update pm2 to latest version

matthew-white:
This PR updates pm2 to the latest version.

I haven't done much to verify this change to the patch version. If there is an issue, then CircleCI or regression testing should help catch it.

No change needed.

### DIFF
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -22,7 +22,7 @@ COPY files/service/crontab /etc/cron.d/odk
 COPY server/package*.json ./
 
 RUN npm clean-install --omit=dev --legacy-peer-deps --no-audit --fund=false --update-notifier=false
-RUN npm install pm2@5.2.0 -g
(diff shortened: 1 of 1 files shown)


sadiqkhoja (approved):
It would be nice to verify that multiple instances of pm2 are running on dev/test environment

matthew-white:
> It would be nice to verify that multiple instances of pm2 are running on dev/test environment

How have you done that in the past? Could we check for that in the `central` CircleCI?

matthew-white:
I think I'm going to go ahead and merge. If we're going to verify this, I think it'd be best to do so in CircleCI. (This won't be the last time that we update pm2!) @sadiqkhoja, let me know if you have thoughts about how to do that!

matthew-white:
From @sadiqkhoja: Go into the service container. There is a command related to pm2 that can be used to check this.

PR #390, opened 2023-02-22
Upgrade from node 16.17.0 to 16.19.1

yanokwa:
This should be a safe upgrade. I've looked through the [v16](https://raw.githubusercontent.com/nodejs/node/main/doc/changelogs/CHANGELOG_V16.md) and I didn't see anything concerning.

### DIFF
--- a/nginx.dockerfile
+++ b/nginx.dockerfile
@@ -1,4 +1,4 @@
-FROM node:16.17.0 as intermediate
+FROM node:16.19.1 as intermediate
 
 COPY ./ ./
 RUN files/prebuild/write-version.sh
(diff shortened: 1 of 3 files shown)


matthew-white:
@sadiqkhoja, I think you did this update last time! Would you mind taking a look at this PR?

It'd also be nice to bump Node in the CircleCI config in Backend and Frontend. That'd be a good way to confirm that tests are still passing. We should also update the package.json `volta` config in Backend and Frontend in order to use 16.19.1 in development.

sadiqkhoja (approved):
Updated node version in backend and frontend repositories as well. All tests are passing.

PR #391, opened 2023-02-23
Upgrade to nginx 1.23.3

yanokwa:
I wanted to explain some of the changes I've made...

docker-compose.yml
* Adding letsencrypt volume so rebuilds don't throw away certs and risk hitting LetsEncrypt limits.
* Upgrade requires removal of old container or mount won't work. Steps: ```docker-compose stop && docker-compose rm nginx && docker-compose build && docker-compose stop && docker-compose up -d```
* I chose not to persist dh, selfsign, /etc/nginx/conf.d because those are transitory. Yes, it saves time when you recreate the container after a docker-compose down and it's not great to leave disconnected volumes around on the disk, but it's also work to move.

inflate_body.lua
* The lua infrastructure that the script relies on aren’t easy to get in this current build of nginx.
* Deflating compressed request bodies is not usually supported in web servers due to possibility of zip bombs.
* This feature is used by front-end to zip up CSV attachments to forms

redirector.conf
* I store a local copy because we rewrite it when turning certbot off. Copying the file each time ensures we have known good copy. I don't want to trust that the image or some script has changed that file.
* I also store a local copy because it's more deterministic when that redirector file is in the right place during the upstream build and our own builds given the use of anonymous volumes
* It's probably safe to keep lines 7-14, but I delete then to guarantee certbot's response protocol isn't triggered. 

certbot.conf
* Was removed in [nginx-certbot v4](https://github.com/JonasAlfredsson/docker-nginx-certbot/releases/tag/v4.0.0) in favor of putting everything in redirector.conf, so I'm removing it here too.

openssl
* nginx already has openssl. Installing it here forces an upgrade. Said upgrade could be dangerous because that update hasn't been tested with our known copy of nginx.

netcat
* There's now only netcat-traditional and netcat-openbsd. The latter has the `-z` flag we need and is up-to-date. netcat-traditional is unmaintained.

You might be wondering why we don't use [docker-nginx-certbot's local CA](https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/master/docs/advanced_usage.md#local-ca). I've explored integrating that, but the reality is that because we support customssl, we'll always have to have a mechanism to use those custom certs. 

The built in local CA uses the same locations as certbot and so it gets messy. Having three known locations `/etc/letsencrypt`, `/etc/customssl`, `/etc/selfsign` is clearer and easier to reason about.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,7 @@ services:
       POSTGRES_PASSWORD: odk
       POSTGRES_DATABASE: odk
   mail:
-    image: "itsissa/namshi-smtp:4.92-8.deb10u6"
+    image: "ixdotai/smtp:v0.2.0"
     volumes:
       - ./files/dkim/config:/etc/exim4/_docker_additional_macros:ro
       - ./files/dkim/rsa.private:/etc/exim4/domain.key:ro
--- a/files/service/config.json.template
+++ b/files/service/config.json.template
@@ -11,7 +11,8 @@
       "transport": "smtp",
       "transportOpts": {
         "host": "mail",
-        "port": 25
+        "port": 25,
+        "ignoreTLS": true
       }
     },
     "xlsform": {


matthew-white:
I just changed the branch to `next`, and it looks like there's now a conflict.

PR #393, opened 2023-02-24
Upgrade to redis 7

yanokwa:
https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/VersionManagementConsiderations.html suggest upgrades to 6.0 are very safe. Kobo runs [6.2](https://github.com/kobotoolbox/kobo-docker/blob/HEAD/docker-compose.backend.template.yml#L42) so that’s also likely safe. I wanted to push the boundaries and try the latest.

I tested this works by starting with a v2023.1.0 install...
* Confirm form preview, form submission, and form edit work before upgrade.
* Confirm form preview, form submission, form edit, and web link submission, work on same form after upgrade.
* Confirm form preview, form submission, and form edit work a new form after upgrade.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -85,7 +85,7 @@ services:
       - SUPPORT_EMAIL=${SYSADMIN_EMAIL}
       - HTTPS_PORT=${HTTPS_PORT:-443}
   enketo_redis_main:
-    image: redis:5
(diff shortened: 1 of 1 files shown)


lognaturel:
I have seen this working on the dev server and I agree that it is low risk.

PR #395, opened 2023-02-27
Renaming for increased clarity

yanokwa:


files/postgres/upgrade-postgres.sh
```diff
@@ -61,7 +59,7 @@ else
     log "!!!"
     log "!!! Please remove file and restart container to continue:"
     log "!!!"
-    log "!!!   rm $flag_deleteOldData_forUsers && docker-compose up -d"
+    log "!!! Please email support@getodk.org for assistance.  "
```
lognaturel, line 62: The line above no longer makes sense.

lognaturel:
This is a series of cosmetic changes. @yanokwa merged it into `next` accidentally and then fixed the build. I think there's one more sentence that needs adjusting to be coherent with the new changes.

alxndrsn:
> This is a series of cosmetic changes.

Also one functional change: data cleanup is no longer automated, but instead documented at https://docs.getodk.org/central-upgrade/#upgrading-to-central-v2023-2.

PR #401, opened 2023-03-07
Extract configurable values from service json template to .env

lognaturel:
Continuation of #389. The goals are for `docker-compose.yml` and the service `config.json.template` to very rarely be customized by users to minimize dirty files.

I tried the mail commit on a server using the default mail container (caribou) and using an external mail service (dev). In particular, I confirmed that it's acceptable to have empty username and password values.

I'm not sure how to verify the Sentry or Node changes but I think they're simple enough to verify by reviewing the code.

It's extending a pattern we've already established. I think it's a good balance of providing configurability in a flexible, secure way and not making dramatic changes to how things are done now.


.env.template
```diff
@@ -11,8 +11,27 @@ SSL_TYPE=letsencrypt
...
+# Optional: configure a custom mail server
+# EMAIL_NO_REPLY=
+# EMAIL_HOST=
+# EMAIL_PORT=
+# EMAIL_SECURE=
+# EMAIL_IGNORE_TLS=
+# EMAIL_USER=
+# EMAIL_PASSWORD=
+
+# Optional: configure a custom Sentry server for error reporting.
```
yanokwa, line 32: > Optional: configure a custom error reporting server
> \# SENTRY_ORG_SUBDOMAIN=

Drop the please leave sentence. I like keeping the Sentry name out of the description to make it less appealing

.env.template
```diff
@@ -11,8 +11,27 @@ SSL_TYPE=letsencrypt
 HTTP_PORT=80
 HTTPS_PORT=443
 
+# Optional: configure Node
```
yanokwa, line 14: >Optional: configure service container
> SERVICE_NODE_OPTIONS

We might use node somewhere else and it seems wise to add a prefix

files/enketo/start-enketo.sh
```diff
@@ -3,13 +3,13 @@
 CONFIG_PATH=${ENKETO_SRC_DIR}/config/config.json
 echo "generating enketo configuration.."
 
+BASE_URL=$( [ "${HTTPS_PORT}" = 443 ] && echo https://"${DOMAIN}" || echo https://"${DOMAIN}":"${HTTPS_PORT}" ) \
```
yanokwa, line 6: Nice!

files/service/scripts/start-odk.sh
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
 
 echo "generating local service configuration.."
+
 ENKETO_API_KEY=$(cat /etc/secrets/enketo-api-key) \
-envsubst '$DOMAIN:$HTTPS_PORT:$SYSADMIN_EMAIL:$ENKETO_API_KEY:$DB_HOST:$DB_USER:$DB_PASSWORD:$DB_NAME' \
+BASE_URL=$( [ "${HTTPS_PORT}" = 443 ] && echo https://"${DOMAIN}" || echo https://"${DOMAIN}":"${HTTPS_PORT}" ) \
+envsubst '$BASE_URL:$SYSADMIN_EMAIL:$ENKETO_API_KEY:$DB_HOST:$DB_USER:$DB_PASSWORD:$DB_NAME:$EMAIL_NO_REPLY:$EMAIL_HOST:$EMAIL_PORT:$EMAIL_SECURE:$EMAIL_IGNORE_TLS:$EMAIL_USER:$EMAIL_PASSWORD:$SENTRY_ORG_SUBDOMAIN:$SENTRY_KEY:$SENTRY_PROJECT' \
```
yanokwa, line 7: Replace : with whitespace so it's consistent.

yanokwa (changes requested):

yanokwa (approved):

PR #403, opened 2023-03-08
Only use named volumes for generated files to prevent compose copy fa…

yanokwa:
Move static files from `/etc/odk` to `/usr/share/odk`. This move is necessary because dockerfile copy commands don't seem to copy files to mounted volumes.

Other changes
* Remove unnecessary rm's to clean up directories. We copy just what we need, so upgrades should be fine.
* Rename letsencrypt volume to associate it with the container.
* Remove unnecessary echos since openssl writes its own helpful output

What I tested
* Fresh install: confirmed with `docker volume ls` that nothing exists. 
* Fresh install with letsecnrupt gets new certs. Also remove nginx, rebuild and confirmed certs are still used.
* Fresh install with selfsignSSL gets correct certs.
* Can toggle between letsencrypt and selfsign with a build and an up.
* Changing odk.conf.template gets the correct file to show up in the container after a build

I did not test
* customssl

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -61,8 +61,8 @@ services:
       driver: local
   nginx:
     volumes:
-      - letsencrypt:/etc/letsencrypt
       - nginx:/etc/odk/nginx
+      - nginx_letsencrypt:/etc/letsencrypt
     build:
       context: .
       dockerfile: nginx.dockerfile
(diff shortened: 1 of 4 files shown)


lognaturel:
We managed to think that the prior dockerfile worked because it's actually really hard to completely clear out an existing Central install (all volumes, etc). We believed we were trying a clean install but were actually relying on some previously-written configs.

PR #411, opened 2023-03-09
Enable database SSL

lognaturel:
Enables database SSL.

Tried on dev with `true` and without the variable in .env. I have not added this one to the `.env.template` because adding it when an external db does not use SSL may result in bad behavior.

Only one considered.

This is reestablishing functionality that worked before. The worst case is that it might not work.

Yes.


matthew-white (approved):
At some point, we should probably change Backend to permit `ssl` to be `false`. It's probably a funny pattern that only `true` or `null`/not-specified are allowed. `null` only happens to work because we generally don't distinguish between `null` and `undefined` in Backend. I wouldn't want to document `null` as a possible value for end users, but in the user docs, we only mention `true` as a possible value. 💯 I think it's fine for us to use `null` here for now given that we know it will work under the hood.

lognaturel:
And now we'll be able to make the change to ``docker-compose.yml`` without merge conflicts!!

PR #414, opened 2023-03-14
Use pm2-runtime with Enketo to ensure quick stop

lognaturel:
Mirror the change in https://github.com/enketo/enketo-express/pull/533 and the `pm2-runtime` usage in https://github.com/getodk/central/blob/0502d5397faf42b641a16487dbcdea10de418dea/files/service/scripts/start-odk.sh#L43


lognaturel:
Not sure why the build is failing.

lognaturel:
The build was failing because I only half converted to pm2-runtime. 🙈 Let's get this in the point release!

PR #415, opened 2023-03-16
Upgrade node

yanokwa:
Follow on from https://github.com/getodk/central/pull/390


lognaturel:
I think we could do 

```
ARG node_version=16.19.1
FROM node:${node_version}
```

would need to be verified. Alternately, a comment above the intermediary container node version might be sufficient.

lognaturel:
I pulled this branch on dev, built, then `docker exec central-service-1 node --version`.  I believe this usage of `ARG` is common.

lognaturel (approved):

PR #417, opened 2023-03-22
Speed up stopping containers

### LINKED ISSUE #416: enketo, service and nginx containers are slow to stop
https://github.com/getodk/central/pull/414 attempted to speed up stopping the enketo container. I thought I had verified the change but it's still taking 10s to stop which means sigterm is failing and sigkill is sent after 10s.

I see that the service and nginx containers have the same issues.

This hasn't been a big deal because we don't tend to do a lot of maintenance synchronously and don't tend to do a lot of development with the whole Docker setup. But still, it would be a nice improvement for those containers to stop immediately, especially when iterating on ops.

lognaturel:
Closes #416

Did a `docker compose build` `docker compose up -d` `docker compose stop` after each commit to make sure each was minimal and effective. It's still on there if someone wants to do a sanity check.

The underlying issue is that signals weren't being proxied to the intended processes. More at https://stackoverflow.com/a/32261019/137744. This was addressed by running all the various processes with `exec`.

Additionally, the Enketo dockerfile needed to be updated to use the JSON-array syntax so it wouldn't be called with `/bin/sh -c` which is immune to SIGTERM

This is an ops change. Worst case would be things don't stop or come up. I think if it works on dev it's likely to work anywhere.

I'm not entirely sure about risk beyond things not stopping or coming up. ~I keep seeing things about the [PID 1 zombie reaping problem](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/) but I can't tell if this changes anything with respect to that. [This post](https://petermalmgren.com/signal-handling-docker/) suggests that it doesn't change anything.~ [This `tini` post has more about zombie processes](https://github.com/krallin/tini/issues/8#issuecomment-146135930). Given my understanding of the processes we launch, I believe it's unlikely that any secondary processes likely to become zombies are being launched.

No.

### DIFF
--- a/enketo.dockerfile
+++ b/enketo.dockerfile
@@ -19,4 +19,4 @@ RUN apt-get update && \
 
 EXPOSE 8005
 
-CMD ./start-enketo.sh
+CMD ["./start-enketo.sh"]
--- a/files/enketo/start-enketo.sh
+++ b/files/enketo/start-enketo.sh
@@ -12,4 +12,4 @@ envsubst '$DOMAIN $BASE_URL $SECRET $LESS_SECRET $API_KEY $SUPPORT_EMAIL' \
     > "$CONFIG_PATH"
 
 echo "starting pm2/enketo.."
-pm2-runtime app.js -n enketo
+exec pm2-runtime app.js -n enketo
--- a/files/nginx/setup-odk.sh
+++ b/files/nginx/setup-odk.sh
@@ -40,5 +40,5 @@ else
     perl -i -ne 'print if $. < 7 || $. > 14' /etc/nginx/conf.d/redirector.conf
     echo "starting nginx for custom ssl and self-signed certs..."
   fi
-  nginx -g "daemon off;"
+  exec nginx -g "daemon off;"
 fi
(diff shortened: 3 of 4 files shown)


enketo.dockerfile
```diff
@@ -19,4 +19,4 @@ RUN apt-get update && \
 
 EXPOSE 8005
 
-CMD ./start-enketo.sh
+CMD ["./start-enketo.sh"]
```
alxndrsn, line 22: I didn't realise this was different  :grimacing:, so I looked it up:

> Unlike the shell form, the exec form does not invoke a command shell.
> -_https://docs.docker.com/engine/reference/builder/#cmd_

...so I think (depending on `ENTRYPOINT`?) the previous code was the equivalent of:

```
CMD ["sh", "-c", "./start-enketo.sh"]
```

:exploding_head:

lognaturel:
Good catch about the db container. Let me look at that separately because it doesn't affect the Cloud setup and I'd like to make sure I understand that process specifically.

Are you feeling pretty confident that this is low-risk? Like I wrote in the risks section, I am not entirely sure that there aren't possible side effects I haven't considered.

sadiqkhoja:
I don't see any risk of zombie processes here as we are not spawning processing on the runtime.

lognaturel:
Have talked it through with @yanokwa and am back to feeling like it's low-risk. Moreover, I don't think we have a great way of identifying possible issues other than putting it out there on different systems.

lognaturel:
Oh great, thanks @sadiqkhoja.

PR #425, opened 2023-05-14
Replace Fathom with Cloudflare

yanokwa:
Fathom's custom domains are the reason I picked them but that feature doesn't work anymore. Also, it's gotten very expensive for what we use it for.


matthew-white (approved):
Should be safe to merge at any time!

Issue #409, opened 2023-03-09
Update Enketo, pyxform-http for v2023.3


matthew-white:

matthew-white:
Closed by #427.

PR #427, opened 2023-05-16
Update pyxform and Enketo

lognaturel:
Closes https://github.com/getodk/central/issues/409

pyxform has a small bug fix.

Enketo has some risk but has been running on https://getodk.org/xlsform and on the QA server for some days.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -96,7 +96,7 @@ services:
       options:
         max-file: "30"
   pyxform:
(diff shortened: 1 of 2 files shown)


matthew-white (approved):

matthew-white:
Are there changes in either that we should highlight in the forum announcement?

lognaturel:
I've put those at the top of the meeting doc!

PR #428, opened 2023-05-16
Update dependencies

### LINKED ISSUE #398: Update ops dependencies for v2023.3
This issue is for .3, not .2.

lognaturel:
Closes #398

I have standardized on pinning the major and minor versions for standard Docker images. This feels like the right balance between predictability and getting updates.

It's on the dev server.

I also considered pinning the patch version or a SHA to ensure consistency. But i think this is a good balance.

The only minor version changed is for node. I haven't looked at the release notes.

No.

### DIFF
--- a/postgres14.dockerfile
+++ b/postgres14.dockerfile
@@ -1,4 +1,4 @@
-FROM postgres:14
+FROM postgres:14.8
 
 COPY files/postgres14/start-postgres.sh /usr/local/bin/
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -121,7 +121,7 @@ services:
       - SUPPORT_EMAIL=${SYSADMIN_EMAIL}
       - HTTPS_PORT=${HTTPS_PORT:-443}
   enketo_redis_main:
-    image: redis:7.0.8
+    image: redis:7.0
     volumes:
       - ./files/enketo/redis-enketo-main.conf:/usr/local/etc/redis/redis.conf:ro
       - enketo_redis_main:/data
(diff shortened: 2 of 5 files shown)


matthew-white:
> The only minor version changed is for node. I haven't looked at the release notes.

If we wanted additional verification of this change, we could update Node in central-backend and central-frontend. These look to be the PRs from the last update: getodk/central-backend#786, getodk/central-frontend#746. The main benefit of that would be getting to run Backend and Frontend tests in CircleCI under the new version of Node. But also regression testing would probably catch most issues.

matthew-white (approved):

postgres14.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM postgres:14
+FROM postgres:14.8
```
matthew-white, line 1: I didn't realize this at first, but some users will have 14.7 or maybe even 14.6 installed, so this will be an upgrade for them. (I initially thought that this line was just pinning the current version.) I don't see why this change would be an issue, but I thought I'd make a note of it. Nothing special needs to happen when upgrading within a major version of Postgres, right?

matthew-white: test.central.getodk.org is at Postgres 14.6, so we'll get a test of 14.6→14.8.

lognaturel: > Nothing special needs to happen when upgrading within a major version of Postgres, right

It shouldn't be the case but mistakes happen so that's why the pinning seems like a good idea moving forward.

PR #429, opened 2023-05-16
Add 5G buffer to space required for postgres upgrade

### LINKED ISSUE #424: Add some disk space margin to check-available-space
https://forum.getodk.org/t/upgrade-from-1-2-to-2023-2-did-not-migrate-database-corectly/41168/6 suggest that if your free space is too close to the available space, you might run into trouble. Maybe we should add a buffer of 5 or 10 GB?

lognaturel:
Closes #424

On the dev server, I changed the line changed by this PR to 

```bash
used="$(( $(du -s /var/lib/docker | cut -f1) + 5000000" )) # add 5G of buffer
```

There's about 11G of stuff in that folder and I confirmed that the script said I needed 17G for the update.

I don't have a good strategy for requesting the space check when the migration has already been done (and there's no `postgres` container). I think this is good enough but the reviewer should see whether they agree!

I considered telling the user we were adding a buffer or only adding it in the error message. But this is simple and feels like the most fool-proof.

The issue mentioned the possibility of 10Gb of buffer. That seemed extreme to me.

I'm not totally sure whether the [arithmetic expansion](https://tldp.org/LDP/abs/html/arithexp.html) is necessary (it worked for me without) but I figure it can't hurt.

Risk is isolated to space determination. It might be broken.

No.

### DIFF
--- a/files/postgres14/upgrade/check-available-space
+++ b/files/postgres14/upgrade/check-available-space
@@ -31,7 +31,7 @@ if [[ "$volume" = "" ]]; then
   exit 1
 fi
 mountpoint="$(docker volume inspect --format='{{.Mountpoint}}' "$volume")"
-used="$(du -s "$mountpoint" | cut -f1)"
+used="$(( $(du -s "$mountpoint" | cut -f1) + 5000000 ))" # add 5G of buffer
 
 log
 log "    Free space: $(numfmt --to=si "$((free*1000))")B"


sadiqkhoja (approved):
Change is good to go. 

However, I noticed the numbers are multiplied by 1000 instead of 1024 in the file

lognaturel:
@sadiqkhoja I briefly wondered about that. I think it ends up being unimportant because of all the rounding so I matched what was already there.

PR #434, opened 2023-05-24
Rename POSTGRES_DB env var

alxndrsn:
Both of our base images ultimately inherit from the official docker-postgres image.  This has an env var POSTGRES_DB, and has never referenced POSTGRES_DATABASE.

See: https://hub.docker.com/_/postgres/
See: https://github.com/docker-library/postgres

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
     environment:
       POSTGRES_USER: odk
       POSTGRES_PASSWORD: odk
-      POSTGRES_DATABASE: odk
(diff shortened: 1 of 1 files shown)


alxndrsn:
An alternative would just be to delete this env var - it defaults to `POSTGRES_USERNAME`, which also explains why the current code works.

matthew-white:
Thanks, @alxndrsn! Just FYI, since we're about to ship .3, we're planning to wait until .4 for this change.

lognaturel (approved):
Indeed!

PR #435, opened 2023-05-26
Ignore docker-compose.override.yml

alanivey:
Allow for overrides of docker-compose.yml without impacting git status.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.swp
 *.swo
 /.env
+/docker-compose.override.yml
 /version.txt
 
 /files/allow-postgres14-upgrade


yanokwa:
@alanivey I'm curious, what are you typically overriding?

alanivey:
Setting `logging` for each service to send logs to Amazon CloudWatch, and making all named and anonymous volumes bind mounts to local directories for a few reasons (ease of backing up and restoring, being able to down+up the containers without losing LE certs)

yanokwa:
@alanivey Thanks for clarifying. I didn't know about `docker-compose.override.yml` and it's super helpful! If you come across Docker best-practices that would be broadly useful to bake into Central, please do file issues. 

@lognaturel This seems like a great change to me. It's a known [convention](https://docs.docker.com/compose/extends/#multiple-compose-files).

lognaturel:
Thanks, @alanivey, this is great. I've also filed a follow-up docs issue: https://github.com/getodk/docs/issues/1616

PR #443, opened 2023-06-30
fixes #438

### LINKED ISSUE #438: Image with $ sign in the name is broken in Enketo
Related to #433

If image contains $ sign in the name then Enketo doesn't preview it.

sadiqkhoja:
fixes #438: if URI is given in proxy_pass then nginx decodes and re-encodes path
which breaks submission attachment with special characters in webform more details: https://trac.nginx.org/nginx/ticket/2506

Manually verified it locally and on Dev server

This is simplest and safest fix. I am going to create separate issue to investigate encoding protocol between Central and Enketo.

None

None

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -25,7 +25,7 @@ server {
   gzip_types text/plain text/css application/json application/x-javascript text/xml text/csv;
 
   location /- {
-    proxy_pass http://enketo:8005/-;
+    proxy_pass http://enketo:8005;
     proxy_redirect off;
     proxy_set_header Host $host;
 


matthew-white:
Targeting the `next` branch because I think we'll release the patch off of `next`.

matthew-white (commented):
This seems like a reasonable approach for now, and I can't think of any downsides. This approach causes nginx to behave how we already thought it was. It's also more similar to how we use `proxy_pass` for Backend.

We've discussed how Enketo has unexpected requirements related to URL encoding. However, we're not sure yet how big changing that in Enketo would be. I'd be glad for us to continue that discussion, but I think this is a nice change for now.

matthew-white (approved):

Issue #432, opened 2023-05-17
Upgrade to Node 18


matthew-white:
Update Node in all three repos: central, central-backend, central-frontend.

PR #452, opened 2023-07-27
Update to node v18

alxndrsn:
As per https://github.com/getodk/central-backend/pull/919#issuecomment-1621949441. Progress toward #432.

### DIFF
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -1,4 +1,4 @@
-ARG node_version=16.20
+ARG node_version=18.17
 FROM node:${node_version} as intermediate
 
 COPY . .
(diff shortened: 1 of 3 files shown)


service.dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG node_version=16.20
+ARG node_version=18.17.0
```
matthew-white, line 1: @lognaturel was the last one to update Node (in #428). The patch version intentionally wasn't specified so that users can continue to get some updates. How about we specify 18.17 here?

matthew-white (approved):

PR #453, opened 2023-07-28
Add template for bug reports by QA team

matthew-white:
This PR adds an issue template that @srujner and @dbemke can use when filing issues. I modified the [Collect issue template](https://github.com/getodk/collect/blob/2d544e1a78174f0427b6f81bc35b6c7baac2c99d/.github/ISSUE_TEMPLATE.md). Suggestions welcome!

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,25 @@
+<!--
+The Central team uses GitHub issues for project management. Most bug reports and feature requests should start in the ODK forum: https://forum.getodk.org
+
+The QA team uses the template below to file bug reports. If you are not filing a bug report, feel free to clear the template.
+-->
+
+#### Problem description
+
+#### URL of the page
+
+#### Steps to reproduce the problem
+
+#### Screenshot
+
+Please don't attach images of QR codes, as those provide access to the server.
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
...
+
+Please don't attach images of QR codes, as those provide access to the server.
+
+#### Expected behavior
+
+#### Central version shown in version.txt
+
+#### Browser version
+
+#### Around when did you see the problem (in UTC)?
```
matthew-white, line 23: This is helpful to know when searching logs.

ktuite (approved):
Looks good to me! Was thinking screenshots _could_ include address bar, should we explicitly include or exclude that? Asking for the URL (and browser) directly seems most useful, and then it doesn't really matter what extra context is in the screenshot.

matthew-white:
> Asking for the URL (and browser) directly seems most useful, and then it doesn't really matter what extra context is in the screenshot.

That sounds right to me. If the URL is in the text of the issue, then we can copy and paste it more easily. As long as the URL is in the text of the issue, I think it's OK either way whether the screenshot shows the URL or doesn't.

srujner:
@matthew-white Everything looks good!

PR #455, opened 2023-08-02
Fix email sending templating bug

### LINKED ISSUE #454: docker-compose templating breaks email sending
In https://github.com/getodk/central/blob/master/docker-compose.yml#L60, we use this syntax:
> EMAIL_FROM=${EMAIL_FROM:-no-reply@${DOMAIN}}`

Given an address of `foo@bar.org`, password resets fail ([stack trace](https://getodk.sentry.io/issues/4350255251/?alert_rule_id=12365294&alert_timestamp=1691000897596&alert_type=email&environment=production&project=4503903535628288&referrer=alert_email)) with this error:
>Mail command failed: 501 Syntax error, parameters in command "MAIL FROM:<foo@bar.org}>" unrecognized or missing

Note the trailing `}` in the email. Somewhere the templating is failing. This is confirmed by running `docker compose config`

yanokwa:
Closes #454 

I ran `docker compose config | grep EMAIL_FROM` to see what the config was. 

* with EMAIL_FROM in .env set to odk@foo.org, I get odk@foo.org in the output. 
* with EMAIL_FROM in .env commented out, I get no-reply@odk-central.foo.org in the output.

I also confirmed that emails are delivered with this change.

There might be a bug in docker/compose v2.20.2/24.0.5, but thankfully https://docs.docker.com/compose/environment-variables/env-file/#parameter-expansion says unbraced vars are supported, so that's why this change works.

It should only affect people who might have changed docker-compose.yml. If something's broken when they upgrade, it will be pretty easy to see because they can't create or reset web users. Hopefully that will lead to a forum post or this issuel.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,7 +57,7 @@ services:
       - DB_PASSWORD=${DB_PASSWORD:-odk}
       - DB_NAME=${DB_NAME:-odk}
       - DB_SSL=${DB_SSL:-null}
-      - EMAIL_FROM=${EMAIL_FROM:-no-reply@${DOMAIN}}
+      - EMAIL_FROM=${EMAIL_FROM:-no-reply@$DOMAIN}
       - EMAIL_HOST=${EMAIL_HOST:-mail}
       - EMAIL_PORT=${EMAIL_PORT:-25}
       - EMAIL_SECURE=${EMAIL_SECURE:-false}


lognaturel (approved):

lognaturel:
I'm annoyed! But also glad there's an easily-verifiable fix.

matthew-white:
Are we thinking of releasing this change with .4? Or are we planning to merge to the `master` branch before .4 (as a patch release)?

yanokwa:
If we are already doing a hot fix, great. If not, it can wait.

matthew-white:
Sounds good. There's a chance we'll have a hot fix coming up, so I won't merge to `next` quite yet, as `next` already has a change for .4. I can merge once we know whether we'll have a hot fix.

matthew-white:
I now think it's less likely that we'll need a hot fix, so I'll go ahead and merge. I'll merge with a merge commit so that it's easy to include the commit in a hot fix if that does end up happening.

PR #456, opened 2023-08-04
Adjust container name if you have compose v2

### LINKED ISSUE #439: No such object: central_postgres_1
I get following error when I run `sudo ./files/postgres14/upgrade/check-available-space` while upgrading Central v1.2 to v2023.3:

```
No such object: central_postgres_1
```

I had to edit `check-available-space`:

```diff
-containerName="central_postgres_1"
+containerName="central-postgres-1"
```

yanokwa:
Closes #439

Sometimes people run the disk space checker after a failed upgrade or maybe they are using compose v2 with an pre-migration install. See https://forum.getodk.org/t/stuck-on-central-upgrade/41421/3

This is an issue because v1 used `_` in names and v2 uses `-`.

Ran the script on a machine with compose v2. I did NOT run it on a machine with v1.

The change isn't very robust because you could have used compose v1 to setup Central and still have v2 in your path, but I can't think of a way to check what version of compose was used to setup the current install.

You can force docker to [use the v1 scheme](https://stackoverflow.com/questions/69464001/docker-compose-container-name-use-dash-instead-of-underscore), but that doesn't really help with this script.

Anyway, seems like this change is not perfect, but it helps.

Slight improvement to running migration.

No.

### DIFF
--- a/files/postgres14/upgrade/check-available-space
+++ b/files/postgres14/upgrade/check-available-space
@@ -15,7 +15,14 @@ fi
 dockerRootDir="$(docker info --format '{{print .DockerRootDir}}')"
 free="$(df --output=avail "$dockerRootDir/volumes" | tail -n+2)"
 
-containerName="central_postgres_1"
+# Support both Docker Compose v1 & v2 container names.
+# See: https://docs.docker.com/compose/migrate/#service-container-names
+if ! docker compose version >/dev/null 2>/dev/null; then
+  containerName="central_postgres_1"
+else
+  containerName="central-postgres-1"
+fi
+
 # The --format argument is a Go template.  Adding spaces between the
 # curlies seems to affect the output.  For more info on this
 # wonderful language, see:


alxndrsn:
> I did NOT run it on a machine with v1.

I don't think it will work in its current form:

```sh
docker run --rm -it --name whatever-$RANDOM docker:1
...
/ # apk update && apk add bash
...
/ # cat ./test.sh 
#!/bin/bash -eu
composeVersion=$(docker compose version)
echo "composeVersion: $composeVersion"
...
/ # ./test.sh 
docker: 'compose' is not a docker command.
See 'docker --help'
/ # echo $?
1
```

A working form might be:

```sh
if docker compose; then
  # v2
  ...
else
  # v1
  ...
fi
```

files/postgres14/upgrade/check-available-space
```diff
@@ -15,7 +15,12 @@ fi
 dockerRootDir="$(docker info --format '{{print .DockerRootDir}}')"
 free="$(df --output=avail "$dockerRootDir/volumes" | tail -n+2)"
 
-containerName="central_postgres_1"
+composeVersion=$(docker compose version)
```
alxndrsn, line 18: Maybe helpful to add something like:
```
# Support both Docker Compose v1 & v2 container names.
# See: https://docs.docker.com/compose/migrate/#service-container-names
```
?

files/postgres14/upgrade/check-available-space
```diff
@@ -15,7 +15,14 @@ fi
 dockerRootDir="$(docker info --format '{{print .DockerRootDir}}')"
 free="$(df --output=avail "$dockerRootDir/volumes" | tail -n+2)"
 
-containerName="central_postgres_1"
+# Support both Docker Compose v1 & v2 container names.
+# See: https://docs.docker.com/compose/migrate/#service-container-names
+if ! docker compose version >/dev/null 2>&1; then
```
alxndrsn, line 20: Maybe intent would be clearer with `2>/dev/null` instead of `2>&1`.

alxndrsn (approved):
I can't find docs saying which version of docker includes which version of compose.

One small change suggested inline, but this looks good to me.

Here's a little test:

```sh
$ for v in 1 17 18 19 20 23 24; do
  echo -n "Docker v$v: "
  docker run --rm -it --name whatever-$RANDOM docker:$v sh -c '
    if ! docker compose version >/dev/null 2>/dev/null; then
      echo underscores
    else
      echo dashes
    fi
  '
done
Docker v1: underscores
Docker v17: underscores
Docker v18: underscores
Docker v19: underscores
Docker v20: dashes
Docker v23: dashes
Docker v24: dashes
```

files/postgres14/upgrade/check-available-space
yanokwa: If someone knows `2>/dev/null`, they'll most likely know `2>&1`. They are both somewhat magical incantations.

alxndrsn: It's the bash equivalent of `x = null; y = x`.  It's decipherable, but easy to confuse with `2>&1 >/dev/null`.

yanokwa:
@alxndrsn Really love the test! You always have clever shell scripts.

matthew-white:
Am I right in thinking that this PR is ready to be merged? @yanokwa, please feel free to merge if so!

Issue #449, opened 2023-07-17
Support Single Sign-on with OpenID Connect


matthew-white:
Creating an issue to track work on SSO:

- [Release criteria](https://docs.google.com/document/d/1lEIp5te0-SkA9xEPwqQdf3Kp6ciTqY8UxGBosoKPK5Q/edit#heading=h.a02nafhkjvhn)
- [Backend PR](https://github.com/getodk/central-backend/pull/910)
- [Frontend PR](https://github.com/getodk/central-frontend/pull/819)
- [`central` PR](https://github.com/getodk/central/pull/458)

ktuite:
### Adventures trying SSO in dev (july 25)

- checked out backend branch
- checked out frontend branch

Note: would be nice to be rebased with master soon to get newer database migrations

Ran on the backend:
`make dev-oidc`
`make fake-oidc-server`

Ran on the frontend:
`npm run dev` and go to the usual `http://localhost:8989/`

Note: At first, things didn't quite run properly and I wasn't sure why. I got hung up on this issue for a while: On my mac, [something](https://stackoverflow.com/a/70456715) is already running on port 5000 so i changed it to another port in `Profile` and `main.nginx.conf`. Then everything ran great.

Appreciated that the fake oidc server gave me useful tips when in entered bad credentials.

matthew-white:
The three PRs above are merged, and the QA team is now taking a look at this issue, so I'm going to go ahead and close. There are still some related to-dos that I've filed as separate issues: #477, getodk/central-backend#971, getodk/central-backend#976.

dbemke:
Logging out on Chrome feels a bit tricky. There a few scenarios that theoretically go along with "do not centralize log out” release criteria but make it difficult to "really log out” and switch a user afterwards. 

Steps to reproduce:
1. Log in to staging on Chrome (this logs in also the browser to the account).
2. Log out of staging.
3. Click continue – the user is logged in without having to choose a user or enter a password


![Image](https://github.com/getodk/central/assets/56479916/c4f8279f-8ef7-4d59-9fce-5c401fb36648)


One thing that is not present after logging out is the ability to choose an account to which I want to log in to (e.g. the second time I want to log in to a different account). So to change the user I need to find the account that I’m currently logged in the browser (it’s possible to make it that it’s not the one visible in the header) and log out of staging and then the option to choose a user appears while logging in.
[...]

PR #458, opened 2023-08-09
Pass OIDC config from .env to Backend and Frontend

matthew-white:
I'm pushing this PR mostly in order to start a conversation about where/how Central users should configure SSO and how that config will be passed through to Backend and Frontend.

Related to the issue for SSO, #449.

I haven't tested this yet, but I wanted to get initial thoughts from @alxndrsn.

This is similar to how we pass other config to Backend. We haven't had to pass config to Frontend before, so that part is new.


files/service/config.json.template
```diff
@@ -33,6 +33,12 @@
       "domain": "${BASE_URL}",
       "sysadminAccount": "${SYSADMIN_EMAIL}"
     },
+    "oidc": {
+      "enabled": ${OIDC_ENABLED},
+      "discoveryUrl": "${OIDC_DISCOVERY_URL}",
+      "clientId": "${OIDC_CLIENT_ID}",
+      "clientSecret": "${OIDC_CLIENT_SECRET}"
```
matthew-white, line 40: Is it possible that these environment variables will contain characters that need to be escaped in JSON (like a double quote)? Hopefully not?

.env.template
```diff
@@ -29,6 +29,12 @@ HTTPS_PORT=443
 # EMAIL_USER=
 # EMAIL_PASSWORD=
 
+# Optional: configure Single Sign-on with OpenID Connect
+# OIDC_ENABLED=
```
matthew-white, line 33: @alxndrsn, I saw your comment in Backend about making this an explicit setting. I personally feel like it'd be nice for users to have one less environment variable to set. What do you think about doing something like what we do with Sentry, whereby Sentry is disabled unless all three Sentry-related config properties are specified?

docker-compose.yml
```diff
@@ -85,6 +89,7 @@ services:
       - SENTRY_ORG_SUBDOMAIN=${SENTRY_ORG_SUBDOMAIN:-o130137}
       - SENTRY_KEY=${SENTRY_KEY:-3cf75f54983e473da6bd07daddf0d2ee}
       - SENTRY_PROJECT=${SENTRY_PROJECT:-1298632}
+      - OIDC_ENABLED=${OIDC_ENABLED:-false}
```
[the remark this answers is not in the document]
matthew-white: I checked, and it looks like the build mostly just has access to environment variables prefixed with `VUE_APP_`: https://cli.vuejs.org/guide/mode-and-env.html#environment-variables. I think I'm going to keep it named `OIDC_ENABLED` in docker-compose.yml, then add the prefix just when calling `npm run build`. It looks like the `VUE_APP_` prefix may need to change to `VITE_` once we move to Vite (getodk/central-frontend#671).

Also, I'm not sure, but for `nginx`, I think `OIDC_ENABLED` needs to be specified in docker-compose.yml under `build.args` rather than `environment` in order for it to make its way to build-frontend.sh. The variable journeys through a couple of layers before eventually reaching `npm build`.

nginx.dockerfile
```diff
@@ -1,8 +1,9 @@
+ARG OIDC_ENABLED
```
matthew-white, line 1: Most examples of `ARG` that I've seen have `ARG` after `FROM`. But service.dockerfile has `ARG` before `FROM`. Maybe because there are two `FROM`s? Things seem to be working here with `ARG` before `FROM`, but I could look more into this.

files/service/config.json.template
alxndrsn: From https://openid.net/specs/openid-connect-discovery-1_0.html#Terminology, it looks like client ID and secret are defined in RFC 6749 (https://www.rfc-editor.org/rfc/rfc6749.txt) Appendix A:

```
A.1.  "client_id" Syntax

   The "client_id" element is defined in Section 2.3.1:

     client-id     = *VSCHAR

A.2.  "client_secret" Syntax

   The "client_secret" element is defined in Section 2.3.1:

     client-secret = *VSCHAR
```

and `VSCHAR` is defined in Appendix A as:

```
VSCHAR     = %x20-7E
```

So theoretically, either value can contain `"` and/or `\`, which would both need to be escaped.

In practice, I haven't seen either character used, but maybe worth noting in docs?

alxndrsn (approved):
Approved if we're happy with current secret management.

nginx.dockerfile
alxndrsn: This seems like a strange use of `ARG`, or perhaps a strange position for it in the dockerfile.  Will setting it here mean every single layer in the dockerfile needs rebuilding when `OIDC_ENABLED` is toggled?

Is it even necessary to set this in the dockerfile at all?  Presumably `RUN files/prebuild/build-frontend.sh` can't pick up standard env vars set in `docker-compose.yml`(?)

files/service/config.json.template
matthew-white: I've left a note in the Google doc that we should mention this in the docs.

matthew-white:
> if we're happy with current secret management

I think what this PR does is consistent with what we've done in the past. Feel free to file an issue though if you think we should rethink that approach!

.env.template
matthew-white: I've made this change here, removing the `OIDC_ENABLED` config. In Frontend, `config.oidcEnabled` will be `true` only if all three remaining `OIDC_*` variables are set.

nginx.dockerfile
matthew-white: > perhaps a strange position for it in the dockerfile

I've moved `ARG` to after the first `FROM`. Based on how I understand the [docs](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact), I think that's what's needed here.

> Will setting it here mean every single layer in the dockerfile needs rebuilding when `OIDC_ENABLED` is toggled?

This question makes me wonder, should I move `ARG` even further down, to right before `RUN files/prebuild/build-frontend.sh`? I don't think what comes before build-frontend.sh takes much time, so I doubt there'd be a big difference. But to be explicit, if an `OIDC_*` config is changed, the Vue build will need to be rerun.

> Presumably `RUN files/prebuild/build-frontend.sh` can't pick up standard env vars set in `docker-compose.yml`(?)

That's my impression from reading some of the docs, but I'm definitely no Docker expert. I tried it out in a couple of commits here, and it does look like variables set in `environment` aren't available in the build. From the build for 735cbe72ad0b974f6905826c026b5733af1bbc17:

```
$OIDC_DISCOVERY_URL in the Dockerfile: [blank]
...
$OIDC_DISCOVERY_URL in build-frontend.sh: [blank]
```

From the build for 8171f6c4614f48298d96ea6378af7ad1f0529415 (using `build.args` instead):

```
$OIDC_DISCOVERY_URL in the Dockerfile: [foobar]
...
$OIDC_DISCOVERY_URL in build-frontend.sh: [foobar]
```

matthew-white:
OK, I'm feeling more confident about this approach! @alxndrsn, would you mind taking another look?

nginx.dockerfile
```diff
@@ -1,8 +1,12 @@
 FROM node:18.17 as intermediate
+ARG OIDC_DISCOVERY_URL
```
matthew-white, line 2: I wanted to note this [warning](https://docs.docker.com/engine/reference/builder/#arg) I'm seeing about `ARG`:

```
It is not recommended to use build-time variables for passing secrets like GitHub keys, user credentials etc. Build-time variable values are visible to any user of the image with the `docker history` command.
```

I don't think this is a particular problem for us, because anyone who could run `docker history` on the machine would also be able to read the .env file. Passing this information via `build.args` doesn't seem riskier than passing it via `environment`. However, I wanted to mention the warning in case I'm misunderstanding something about it.

matthew-white: Right now, build-frontend.sh derives `VUE_APP_OIDC_ENABLED` from the other three `OIDC_*` variables. It'd be nice if we could derive that value somewhere prior to the Dockerfile, then pass only that value to the Dockerfile as an `ARG`. However, I don't know of an easy way to do that with our current setup. And again, my understanding is that this approach isn't a particular risk for us.

matthew-white: @alxndrsn and I discussed this earlier today. Instead all three `OIDC_*` variables being passed to nginx.dockerfile, now only `OIDC_DISCOVERY_URL` will be passed. Unlike `OIDC_CLIENT_SECRET`, `OIDC_DISCOVERY_URL` isn't sensitive, so there shouldn't be an issue that it is passed via `ARG`. build-frontend.sh will derive `VUE_APP_OIDC_ENABLED` based only on the presence of `OIDC_DISCOVERY_URL`.

.env.template
matthew-white: Update: I've now set it up so that `config.oidcEnabled` will be set based only on the presence of `OIDC_DISCOVERY_URL`. Setting it up that way means that we don't need to pass `OIDC_CLIENT_SECRET` to nginx.dockerfile.

nginx.dockerfile
alxndrsn: Seems like it doesn't matter where the `ARG` is declared:

> ## Impact on build caching
> 
> `ARG` variables are not persisted into the built image as `ENV` variables are. However, `ARG` variables do impact the build cache in similar ways. If a Dockerfile defines an `ARG` variable whose value is different from a previous build, then a “cache miss” occurs upon its first usage, not its definition. In particular, all RUN instructions following an `ARG` instruction use the `ARG` variable implicitly (as an environment variable), thus can cause a cache miss. All predefined `ARG` variables are exempt from caching unless there is a matching `ARG` statement in the Dockerfile.
> -_https://docs.docker.com/engine/reference/builder/#impact-on-build-caching_

nginx.dockerfile
alxndrsn: Even if derived from `OIDC_DISCOVERY_URL`, I think it would still be helpful to call this `OIDC_ENABLED` or similar in the frontend image, as that is what it's used for.  I think seeing `OIDC_DISCOVERY_URL` in the frontend would raise questions about why it's being passed there.

nginx.dockerfile
matthew-white: Done!

nginx.dockerfile
matthew-white: Alex and I chatted about this earlier today. We decided to reintroduce a separate `OIDC_ENABLED` variable so that we don't need to pass the discovery/issuer URL to the Frontend build. The user will specify the `OIDC_ENABLED` variable: it's not derived from the other variables. I've made this change in a commit I just pushed.

.env.template
matthew-white: We ended up deciding to revert this change. Users will manually specify `OIDC_ENABLED`, which will be available to Frontend via `config.oidcEnabled`.

sadiqkhoja (approved):
looks good to me

matthew-white:
I'm going to go ahead and merge! In addition to @sadiqkhoja's approval, the final version of the code is very similar to the version that @alxndrsn approved. @alxndrsn, let me know if you have additional thoughts about anything here!

PR #475, opened 2023-09-01
service: install wait-for-it with apt-get

alxndrsn:
Replaces the version-controller `wait-for-it.sh` script with https://tracker.debian.org/pkg/wait-for-it.


lognaturel:
I'm having trouble finding the comment in https://github.com/getodk/central-backend/pull/910 that led to this change, can you please say a bit about the rationale?

If we do go this route, I think we should make sure the version is explicit.

The deb package is based on db049716e42767d39961e95dd9696103dca813f1 from https://github.com/vishnubob/wait-for-it/commits/master (July 2017). It looks like the version in the repo was one or two commits ahead but nothing significant.

alxndrsn:
> I'm having trouble finding the comment in [getodk/central-backend#910](https://github.com/getodk/central-backend/pull/910) that led to this change, can you please say a bit about the rationale?

Here it is: https://github.com/getodk/central-backend/pull/910#discussion_r1312015088

The rationale for doing this is somewhere on the same spectrum as why anything is installed from package management instead of committing e.g. [`node_modules` in `odk-central-backend`](https://github.com/getodk/central-backend) or [the `cron` binary in this repo](https://github.com/getodk/central/blob/master/service.dockerfile#L19) to git.

This script is smaller than those examples, but managing it like other dependencies seems neater than copy-pasting:

* avoid the temptation to make repo-specific modifications
* more easily know about and incorporate official bugfixes
* get rid of the special case for `shellcheck`
* reduce the amount of code in this repo
* remove the need to track changes made here and in other ODK repos it gets added to

If it was being added now, I think @sadiqkhoja's suggestion to install with `apt-get` is hard to argue with.

> It looks like the version in the repo was one or two commits ahead

I hadn't noticed this!  It seems like [the version in APT](https://tracker.debian.org/pkg/wait-for-it) is from 2018 :grimacing:

Confirmation:

```
$ docker run --rm node:18.17 bash -c 'apt-get -qq update && apt-cache show wait-for-it | grep Version'
Version: 0.0~git20180723-1
```

I think this is a strong argument to maintain the current approach, and review at some point if the debian package gets updated.

yanokwa:
I think we should use the apt version for tools like these because I trust Debian's process to keep binaries secure and up-to-date. I don't want us to have to watch wait-for-it's repo.

I also don't think we need to pin versions on things we install via apt. Here again, I think Debian does a reasonable job and would likely notice and patch before we do. We already have a bunch of floating dependencies (e.g., openssl, perl, netcat-openbsd, gettext-base) and it hasn't been a problem.

In general, I think we should reduce the number of special things we do unless there is a very compelling reason.

lognaturel:
> We already have a bunch of floating dependencies (e.g., openssl, perl, netcat-openbsd, gettext-base) and it hasn't been a problem.

Good point. Ok, I'm convinced. I assume it's also the kind of thing where if something changes in an unexpected way it would fail spectacularly, not subtly.

> I think this is a strong argument to maintain the current approach, and review at some point if the debian package gets updated.

I now disagree with my own argument, sorry. 😬

lognaturel (approved):

matthew-white:
Just wanted to check about the status of this PR. @alxndrsn, is this PR ready to be merged?

alxndrsn:
> is this PR ready to be merged?

I can't currently run the `next` branch locally, so I haven't tested the change.  But if you're happy it works, please merge :upside_down_face:

matthew-white:
Do you think it should be verified locally, or is it enough that it passes on CircleCI? Hopefully if it's going to fail, it'll fail spectacularly, similar to what @lognaturel said, and CircleCI would catch it.

alxndrsn:
@matthew-white I'd prefer to validate locally (and be able to check future work locally), but as above: if you're happy it works, please merge 🙃

alxndrsn:
I think I've got this working...

matthew-white:
> I think I've got central working locally...

That's exciting! What do you think about adding instructions for running locally to the readme whenever you have time? I'd be interested in trying out those instructions as well.

PR #478, opened 2023-09-02
docker-compose.yml: fix empty-string default values

alxndrsn:
It looks like the intention of e.g. `EMAIL_USER=${EMAIL_USER:-''}` is to set `EMAIL_USER` to an empty string.

It's actual effect is to set it to a string containing empty quotes (`''`):

```
$ docker compose -f <(cat <<EOF
version: "3"
services:
  tester:
    image: debian
    environment:
      - EMPTY_STRING=${EMPTY_STRING:-}
      - EMPTY_QUOTES=${EMPTY_QUOTES:-''}
    command: [ "env" ]
EOF
) up

[+] Running 1/0
 ✔ Container fd-tester-1  Created                                                                                                            0.0s 
Attaching to fd-tester-1
fd-tester-1  | PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
fd-tester-1  | HOSTNAME=9cdc11ef2481
fd-tester-1  | EMPTY_STRING=
fd-tester-1  | EMPTY_QUOTES=''
fd-tester-1  | HOME=/root
fd-tester-1 exited with code 0
```

This non-empty string maintains when substituted into JSON as into https://github.com/getodk/central/blob/410890a506c26da287a1a3ad9251afe3a92986b0/files/service/config.json.template#L19 at https://github.com/getodk/central/blob/410890a506c26da287a1a3ad9251afe3a92986b0/files/service/scripts/start-odk.sh#L7-L9:

```sh
$ export EMPTY_STRING=""; \
  export EMPTY_QUOTES="''"; \
  echo '{ "empty_string":"${EMPTY_STRING}", "empty_quotes":"${EMPTY_QUOTES}" }' |
    envsubst '$EMPTY_STRING $EMPTY_QUOTES'

{ "empty_string":"", "empty_quotes":"''" }
```

This looks like a future source of confusing behaviour/bugs.  As an example, @matthew-white I think this would have broken the derived approach to calculating `OIDC_ENABLED` we previously considered for `odk-central-frontend`.


yanokwa:
This change looks good to me. 

I'm not sure why we went with quotes to begin with, but I'm guessing it's because https://docs.docker.com/compose/environment-variables/env-file/#parameter-expansion doesn't show a good example of what an empty string looks like.

lognaturel (approved):

Issue #441, opened 2023-06-30
Update Enketo, pyxform-http for v2023.4


matthew-white:

lognaturel:
Realizing we need a pyxform update to match the Dataset -> Entity List change.

matthew-white:
Are there changes to Enketo or pyxform that we should highlight in the Central forum announcement? I'm assuming not for pyxform if the main change was for Dataset → Entity List. Is there anything that we should mention about Enketo?

Issue #440, opened 2023-06-30
Update infrastructure dependencies for v2023.4


matthew-white:
Update any dependencies other than Backend, Frontend, Enketo, and pyxform-http that need to be updated. We will be updating Node 18 as part of this release, but that's a separate issue: #432.

matthew-white:
#485 seems like a larger improvement to DKIM, so I've added that to the GitHub project separately. I'll go ahead and close this issue (closed by #484).

PR #484, opened 2023-09-16
Update dependencies

lognaturel:
Makes progress towards https://github.com/getodk/central/issues/440 Closes https://github.com/getodk/central/issues/441

`pyxform`: converted a form with `list_name` on `entities` sheet to verify the latest version is used

`postgres`: clicked around on dev server

`redis`: opened a few forms on dev server

<strike>`smtp`: this one was the most annoying. There's nothing particularly significant in the changelog but I still wanted to verify it. In this process, I discovered that Digital Ocean [now blocks SMTP](https://docs.digitalocean.com/support/why-is-smtp-blocked/). I pulled @yanokwa in. We verified exim logs inside the container to make sure messages were being queued in the same way as with prior versions. He also verified that https://www.mail-tester.com/ gets email delivered. In doing all this he discovered some DKIM-related bugs that he'll follow up on.</strike> @yanokwa will submit a separate PR for this update.

I checked all docker files and docker-compose for versions to update. I believe this is exhaustive.

I don't see much risk after the verification above.

No.

### DIFF
--- a/postgres14.dockerfile
+++ b/postgres14.dockerfile
@@ -1,4 +1,4 @@
-FROM postgres:14.8
+FROM postgres:14.9
 
 COPY files/postgres14/start-postgres.sh /usr/local/bin/
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -102,7 +102,7 @@ services:
       options:
         max-file: "30"
   pyxform:
-    image: 'ghcr.io/getodk/pyxform-http:v1.12.1'
+    image: 'ghcr.io/getodk/pyxform-http:v1.12.2'
     restart: always
   secrets:
     volumes:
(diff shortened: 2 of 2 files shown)


postgres14.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM postgres:14.8
+FROM postgres:14.9
```
matthew-white, line 1: It might be nice to update this in the Backend [CircleCI config](https://github.com/getodk/central-backend/blob/master/.circleci/config.yml) so that we're certain that Backend tests run successfully against the version of Postgres that we ship. Probably not a big deal though: it'd be pretty surprising if anything failed.

matthew-white (approved):

PR #485, opened 2023-09-17
Improve DKIM support

yanokwa:
DKIM is essentially a requirement to get emails delivered, so my high-level goal is to make it easier to turn on, without having to risk merge conflicts. Some things that you should know about the SMTP container.

* It needs to have an existing file or the volume mapping will create an empty folder. That's why `rsa.empty` is there as a place holder in the compose file. It prevent that from happening.
* It expects the `rsa.private` file at `/etc/exim4/dkim.key.temp` so that's why that's hard coded in the compose file.
* It enables DKIM when `DKIM_KEY_PATH` has a value, so by default, we set it to empty, via `DKIM_CONTAINER_KEY` to disable it. Only when `DKIM_CONTAINER_KEY` has a value will DKIM be enabled. And the container expects this value to be the same as the container location of the mapped file, so that's why the .env has that value hard coded.
* config.disabled can be removed because those rules are now baked into the container.

I don't love how confusing this is, but it's the best I could think of. Open to feedback.

I confirmed that, after rebuild, emails send after removing `rsa.private` and `rsa.public` files and with the `DKIM_*` in the env file commented out. The emails are not DKIM enabled when they are received. 

I've also confirmed, after rebuild,  that emails send after adding `rsa.private`  and `rsa.public` files and with the `DKIM_*` enabled. The emails are DKIM enabled.

I considered just changing the permissions on `rsa.private` and leaving our implementation as is, but that's not as secure.

It could be nicer if the upstream container generated the keys. That'd simplify things, but it feels out of scope.

Honestly, we should consider removing this SMTP container entirely and ask that people use an upstream mail service. That's what Discourse, Ghost, etc, do.

If you already have DKIM support, all you need to do is add values to .env and it should now work because the permissions will actually be correct. That said, I have not tried an upgrade. 

https://github.com/getodk/docs/pull/1655

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,8 @@
 !/files/postgres14/upgrade/check-available-space
 
 /files/local/customssl/*.pem
+
+/files/mail/rsa.private
+/files/mail/rsa.public
+
+/files/dkim/*
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,12 +27,12 @@ services:
       POSTGRES_PASSWORD: odk
       POSTGRES_DATABASE: odk
   mail:
-    image: "ixdotai/smtp:v0.2.0"
+    image: "ixdotai/smtp:v0.5.1"
     volumes:
-      - ./files/dkim/config:/etc/exim4/_docker_additional_macros:ro
-      - ./files/dkim/rsa.private:/etc/exim4/domain.key:ro
+      - ./files/mail/rsa.private:/etc/exim4/dkim.key.temp:ro
     environment:
       - MAILNAME=${DOMAIN}
+      - DKIM_KEY_PATH=/etc/exim4/dkim.key.temp
     restart: always
   service:
     build:
--- a/files/dkim/config.disabled
+++ b/files/dkim/config.disabled
@@ -1,5 +1,0 @@
-DKIM_DOMAIN = ${lc:${domain:$h_from:}}
-DKIM_KEY_FILE = /etc/exim4/domain.key
-DKIM_PRIVATE_KEY = ${if exists{DKIM_KEY_FILE}{DKIM_KEY_FILE}{0}}
-DKIM_SELECTOR = dkim
-DKIM_CANON = simple


.gitignore
```diff
@@ -9,4 +9,9 @@
 /files/postgres14/upgrade/*
 !/files/postgres14/upgrade/check-available-space
 
-/files/local/customssl/*.pem
\ No newline at end of file
+/files/local/customssl/*.pem
+
+/files/local/customssl/*.pem
+
```
yanokwa, line 15: Track everything but rsa.empty.

.env.template
```diff
@@ -11,6 +11,10 @@ SSL_TYPE=letsencrypt
 HTTP_PORT=80
 HTTPS_PORT=443
 
+# Optional: configure DKIM. Do not change these values.
+# DKIM_HOST_KEY=./files/dkim/rsa.private
+# DKIM_CONTAINER_KEY=/etc/exim4/dkim.key.temp
```
matthew-white, line 16: What do you think about moving these values closer to the `EMAIL_*` config? (Below them?) I feel like it's nice to keep thematically related config closer together.

.env.template
```diff
@@ -11,6 +11,10 @@ SSL_TYPE=letsencrypt
 HTTP_PORT=80
 HTTPS_PORT=443
 
+# Optional: configure DKIM. Do not change these values.
```
matthew-white, line 14: I probably missed the explanation, but just looking at this file, I feel unclear about whether these values are meant to configurable. If they're not configurable, I feel like we should try hard not to include them in this file, because it'd be one more thing for users to try to understand. Maybe there's something with variable defaults we could do in docker-compose.yml. (Defaults can be nested I think, if that helps.) On the other hand, if these values are configurable, I think we should soften the wording of "Do not change these values."

matthew-white (commented):
I left a couple of random comments about .env, but I'm thinking that @lognaturel should take  the first look at this PR.

.env.template
matthew-white: Another idea: maybe a `DKIM_ENABLED` variable, similar to `OIDC_ENABLED`?

yanokwa:
@lognaturel and I have found ways to move the variables out of docker-compose.

We've done this by creating an entirely new folder (`mail`, to match the container name) for the DKIM file. We'll create a zero byte `rsa.private` file by default so the volume binding in the container doesn't create an empty folder. Then during upgrade, we ask users to move their old files to this new location (see https://github.com/getodk/docs/pull/1655).

We've updated .gitignore to make sure the `rsa.p*` files don't show up as dirty files in the repo, don't cause merge conflicts, and aren't accidentally checked into the repo. The latter two concerns are why we've moved to an entirely new folder structure instead of overwriting the old `rsa.private`.

I've sent in a PR to https://github.com/ix-ai/smtp/pull/30 to only enable DKIM if the file has more than zero bytes, but I've also confirmed that mail sending works even if DKIM is enabled with an empty file. It sends, but includes an error message in the log.

Todo:
- [] Verify that upgrading works with existing installs. I'm particularly concerned about Docker holding on to `rsa.private` and   `rsa.p*` not being ignored via git.

yanokwa:
I have verified this works on a clean install.
* Fresh install with no DKIM.
* Confirm email sending works with no DKIM (but with an error in EXIM log).
* Run upgrade steps per https://github.com/getodk/docs/pull/1655
* Check out this PR.
* Build and restart mail container per  https://github.com/getodk/docs/pull/1655
* Enable DKIM per https://github.com/getodk/docs/pull/1655
* Confirm email sending works with DKIM (and no error)

I have verified this works on a install with a working DKIM, which admittedly is rare because someone would have to know to remove the config and rsa.private directories from the existing install and rebuild the mail container. But assuming that, I 
* Run upgrade steps per https://github.com/getodk/docs/pull/1655
* Check out this PR.
* Build and restart mail container per https://github.com/getodk/docs/pull/1655
* Enable DKIM per https://github.com/getodk/docs/pull/1655
* Confirm email sending works with DKIM (and no error)

docker-compose.yml
```diff
@@ -27,12 +27,12 @@ services:
       POSTGRES_PASSWORD: odk
       POSTGRES_DATABASE: odk
   mail:
-    image: "ixdotai/smtp:v0.2.0"
+    image: "ixdotai/smtp:v0.5.1"
     volumes:
-      - ./files/dkim/config:/etc/exim4/_docker_additional_macros:ro
```
lognaturel, line 32: Is this because it's considered there's no need for further configuration beyond the private key?

yanokwa: Correct. It's now in the container: https://github.com/ix-ai/smtp/blob/master/entrypoint.sh#L61

lognaturel: > config.disabled can be removed because those rules are now baked into the container.

lognaturel (approved):
I have talked extensively with @yanokwa about his process. I understand all the changes here and agree they make sense. I haven't tried it myself but I saw @yanokwa trying it.

PR #525, opened 2023-10-20
Add dev docker-compose file to manage supporting services

sadiqkhoja:
## Changes

- Added a new docker-compose file for development. It is supposed to be executed in conjunction with the main docker-compose.yml. This new file adds `profiles` attribute to control which services are needed for the development workflow. ~Additionally, network_mode is set to `host` driver so that bidirectional communication is possible. Bottleneck for `bridge` mode is Enketo which for some reason does not honor /etc/hosts for localhost, I was able to make it work with code modification on the fly, which looked dirty to me. _One caveat for `host` mode: this is not available for "Docker Desktop for Mac", works great with OrbStack_~
- ~~Added dev config templates for `nginx`, `enketo` and `service(central)` services. These dev templates have hardcoded `localhost` endpoints~~ 

Two objectives I've kept in my mind to achieve this 1) Not to touch main `docker-compose.yml` that might cause unintended consequences 2) Not to modify application source code

Verified locally. It would be great if all can try this and see if this is helpful.

Although I have modified enketo startup scripts, changes are contained in dev specific conditional block

None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,11 @@
+.PHONY: dev
+dev:
+	docker compose --profile central -f docker-compose.yml -f docker-compose.dev.yml up -d
+
+.PHONY: upgrade-successful
+upgrade-successful:
+	docker compose exec postgres14 touch /var/lib/odk/postgresql/14/.postgres14-upgrade-successful
+
+.PHONY: stop
+stop:
+	docker compose stop
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,6 @@
 .PHONY: dev
 dev:
 	docker compose --profile central -f docker-compose.yml -f docker-compose.dev.yml up -d
-
-.PHONY: upgrade-successful
-upgrade-successful:
 	docker compose exec postgres14 touch /var/lib/odk/postgresql/14/.postgres14-upgrade-successful
 
 .PHONY: stop
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 .PHONY: dev
 dev:
 	docker compose --profile central -f docker-compose.yml -f docker-compose.dev.yml up -d
-	docker compose exec postgres14 touch /var/lib/odk/postgresql/14/.postgres14-upgrade-successful
 
 .PHONY: stop
 stop:
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
+docker-compose-dev := docker compose --profile central -f docker-compose.yml -f docker-compose.dev.yml
+
 .PHONY: dev
 dev:
-	docker compose --profile central -f docker-compose.yml -f docker-compose.dev.yml up -d
+	$(docker-compose-dev) up -d
 
 .PHONY: stop
 stop:
-	docker compose stop
+	$(docker-compose-dev) stop
--- a/README.md
+++ b/README.md
@@ -28,7 +28,11 @@ In addition to the Backend and the Frontend, Central deploys services:
 * Central relies on [pyxform-http](https://github.com/getodk/pyxform-http) for converting Forms from XLSForm. It generally shouldn't be needed in development but can be run locally.
 * Central relies on [Enketo](https://github.com/enketo/enketo-express) for Web Form functionality. Enketo can be run locally and configured to work with Frontend and Backend in development by following [these instructions](https://github.com/getodk/central-frontend/blob/master/docs/enketo.md).
 
-If you want to work on Central codebase and don't want to setup dependent services like Postgresql, Enketo, etc manually then you can run `make dev`, which will start those services as Docker containers.
+If you want to work on Central codebase and don't want to setup dependent services like Postgresql, Enketo, etc manually then you can run `make dev`, which will start those services as Docker containers. With this setup you would need a local domain name, `central-dev` is a good choice. Following are the places where you need to add this:
+
+* Set `DOMAIN=central-dev` in .env file. 
+* Add an entry in your `/etc/hosts` file for `127.0.0.1 central-dev`.
+* Create `local.json` in central-backend directory and set value of default.env.domain to `http://central-dev:8989`
 
 Operations
 ----------
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ In addition to the Backend and the Frontend, Central deploys services:
 * Central relies on [pyxform-http](https://github.com/getodk/pyxform-http) for converting Forms from XLSForm. It generally shouldn't be needed in development but can be run locally.
 * Central relies on [Enketo](https://github.com/enketo/enketo-express) for Web Form functionality. Enketo can be run locally and configured to work with Frontend and Backend in development by following [these instructions](https://github.com/getodk/central-frontend/blob/master/docs/enketo.md).
 
-If you want to work on Central codebase and don't want to setup dependent services like Postgresql, Enketo, etc manually then you can run `make dev`, which will start those services as Docker containers. If you are setting up development environment for the first time, please run `make upgrade-successful` to skip the Postgresql upgrade.
+If you want to work on Central codebase and don't want to setup dependent services like Postgresql, Enketo, etc manually then you can run `make dev`, which will start those services as Docker containers.
 
 Operations
 ----------
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ In addition to the Backend and the Frontend, Central deploys services:
 * Central relies on [pyxform-http](https://github.com/getodk/pyxform-http) for converting Forms from XLSForm. It generally shouldn't be needed in development but can be run locally.
 * Central relies on [Enketo](https://github.com/enketo/enketo-express) for Web Form functionality. Enketo can be run locally and configured to work with Frontend and Backend in development by following [these instructions](https://github.com/getodk/central-frontend/blob/master/docs/enketo.md).
 
+If you want to work on Central codebase and don't want to setup dependent services like Postgresql, Enketo, etc manually then you can run `make dev`, which will start those services as Docker containers. If you are setting up development environment for the first time, please run `make upgrade-successful` to skip the Postgresql upgrade.
+
 Operations
 ----------
 
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,0 +1,51 @@
+# Assumption: This file will be merged by `docker compose`
+#
+# What it does:
+# Sets profiles for each service depending on whether that 
+# service is required for development of that particular application
+# For Central be/fe development, we need postgres14, pyxform, enketo, redis_main
+# For Enketo development, we need postgres14, service, nginx, redis_main
+# 'none' profile is set for services that are not need for either
+#  
+# depends_on of some services are reset using !reset custom yml tag
+# nb: can't replace depends_on so we have removed all the values, ok for dev
+services:
+  postgres14:
+    profiles:
+      - central
+    network_mode: host
+  postgres:
+    profiles:
+      - none
+  mail:
+    profiles:
+      - none
+  service:
+    profiles:
+      - none
+  nginx:
+    profiles:
+      - none
+  pyxform:
+    profiles:
+      - central
+    # changing port here - can't use `ports` mapping with host networking
+    command: ["gunicorn", "--bind", "0.0.0.0:5001", "--workers", "5", "--timeout", "600", "--max-requests", "1", "--max-requests-jitter", "3", "main:app()"]
+    network_mode: host
+  secrets:
+    profiles:
+      - none
+  enketo:
+    profiles:
+      - central    
+    depends_on: !reset []
+    environment:
+      - ENV=DEV
+    network_mode: host
+  enketo_redis_main:
+    profiles:
+      - central
+    network_mode: host
+  enketo_redis_cache:
+    profiles:
+      - none
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -2,10 +2,7 @@
 #
 # What it does:
 # Sets profiles for each service depending on whether that 
-# service is required for development of that particular application
-# For Central be/fe development, we need postgres14, pyxform, enketo, redis_main
-# For Enketo development, we need postgres14, service, nginx, redis_main
-# 'none' profile is set for services that are not need for either
+# service is required for Central development or not.
 #  
 # depends_on of some services are reset using !reset custom yml tag
 # nb: can't replace depends_on so we have removed all the values, ok for dev
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -13,7 +13,9 @@ services:
     network_mode: host
   postgres:
     profiles:
-      - none
+      - central
+    image: debian:bullseye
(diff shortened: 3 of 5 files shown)


spwoodcock:
Is work still ongoing on this PR?

I have just set up my own dev environment to work on central-backend and central-frontend & documented the process.
As a result, I have a slightly modified dev setup, also with a `docker-compose.dev.yml`.

I plan to submit my documentation for review, but should I also commit the dev setup changes?

sadiqkhoja:
Hi @spwoodcock, I will be working again on this PR next week, particularly removing duplicate *.json.dev.template from the individual services. 

Can you please share your setup, is it a different approach? Maybe we can combine both.

Makefile
```diff
@@ -0,0 +1,17 @@
+.PHONY: for-central-dev
+for-central-dev:
+	docker compose --profile secrets -f docker-compose.yml -f docker-compose.dev.yml up
```
sadiqkhoja, line 3: Running `secrets` container first to generate hardcoded secrets for dev.

can't override `depends_on` in docker-compose.dev.yml because `!override` is not released in OrbStack

postgres-upgrade.dockerfile
```diff
@@ -1,5 +1,5 @@
 # see: https://github.com/tianon/docker-postgres-upgrade/blob/master/9.6-to-14/Dockerfile
-FROM tianon/postgres-upgrade:9.6-to-14
+FROM --platform=linux/amd64 tianon/postgres-upgrade:9.6-to-14
```
sadiqkhoja, line 2: this image is available for amd64 architecture, we need to specify this we running it on apple chip (M1/M2/arm64)

Makefile
```diff
@@ -0,0 +1,17 @@
...
+	docker compose --profile central -f docker-compose.yml -f docker-compose.dev.yml up -d
+
+.PHONY: for-enketo-dev
+for-enketo-dev:
+	docker compose --profile secrets -f docker-compose.yml -f docker-compose.dev.yml up
+	docker compose --profile enketo -f docker-compose.yml -f docker-compose.dev.yml up -d
+
+.PHONE: upgrade-successful
```
alxndrsn, line 11: :phone:

files/postgres14/start-postgres.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -eu
 set -o pipefail
 
-flag_upgradeCompletedOk="$PGDATA/../.postgres14-upgrade-successful"
+flag_upgradeCompletedOk=$(readlink -f -m "$PGDATA/../.postgres14-upgrade-successful")
```
sadiqkhoja, line 4: `data` directory is not there when running for the first time without postgres upgrade

Makefile
sadiqkhoja: no longer required, I have set  dev secrets via `start-enketo.sh`

postgres-upgrade.dockerfile
sadiqkhoja: this is required only if we run `docker compose build` which tries to build images for all service. For development, we don't need to build this image, hence reverted this change.

files/enketo/start-enketo.sh
```diff
@@ -3,6 +3,15 @@
 CONFIG_PATH=${ENKETO_SRC_DIR}/config/config.json
 echo "generating enketo configuration..."
 
+if [ "$ENV" = "DEV" ]; then
+    sed -i -e 's/enketo_redis_main/localhost/g' \
+       -e 's/enketo_redis_cache/localhost/g' \
+       -e 's/6380/6379/g' \
+       -e 's/${SECRET}/s0m3v3rys3cr3tk3y/g' \
+       -e 's/${LESS_SECRET}/this $3cr3t key is crackable/g' \
+       -e 's/${API_KEY}/enketorules/g' "$CONFIG_PATH.template"
```
alxndrsn, line 12: Is there a reason these env vars can't be handled in `docker-compose.dev.yml`?

alxndrsn:
Anything to make this easier v welcome :clap: 

I've run `make dev`; what do I do now?

lognaturel:
@sadiqkhoja is away for the next couple weeks so I'll try to answer questions as best I can!

> I've run make dev; what do I do now?

I believe the current intended usage is that you would then run frontend and backend locally and they would have access to Enketo, pyxform, etc.

docker-compose.dev.yml
```diff
@@ -0,0 +1,51 @@
+# Assumption: This file will be merged by `docker compose`
+#
+# What it does:
+# Sets profiles for each service depending on whether that 
+# service is required for development of that particular application
+# For Central be/fe development, we need postgres14, pyxform, enketo, redis_main
+# For Enketo development, we need postgres14, service, nginx, redis_main
```
lognaturel, line 7: Remove stale comment.

lognaturel:
I believe the idea is that the supporting services are more complex to get up and running. The involve knowledge Central dev doesn't require and some have lots of dependencies to manage. The goal is to have a way to get them up and running that is as close to the production configuration as possible and that doesn't risk getting out of date.

You can see a different angle from @spwoodcock which targets also containerizing frontend and backend for dev: https://github.com/hotosm/odkcentral/pull/1/files#diff-97db29a7915320e63d41d38a0440360a87055ee8ed03757aa263116dbbb4aabe

That approach gets to what @moise10r is after with https://github.com/getodk/central-frontend/pull/1048

If we go in that direction, we should make sure there is an answer for hot reloading and debugging. We should also make sure the maintenance burden/risk of it getting out of sync is worth it.

files/postgres14/start-postgres.sh
alxndrsn: What happens currently when the "`data`" directory is missing?

sadiqkhoja: script keeps listening for the existence of `postgres14-upgrade-successful` file and never moves forward to start the database

files/enketo/start-enketo.sh
sadiqkhoja: These variables are read as local variable from the files in the next line. Setting them in docker-compose.dev.yml will not work because local variables have precedence over other variables

README.md
```diff
@@ -28,6 +28,8 @@ In addition to the Backend and the Frontend, Central deploys services:
 * Central relies on [pyxform-http](https://github.com/getodk/pyxform-http) for converting Forms from XLSForm. It generally shouldn't be needed in development but can be run locally.
 * Central relies on [Enketo](https://github.com/enketo/enketo-express) for Web Form functionality. Enketo can be run locally and configured to work with Frontend and Backend in development by following [these instructions](https://github.com/getodk/central-frontend/blob/master/docs/enketo.md).
 
+If you want to work on Central codebase and don't want to setup dependent services like Postgresql, Enketo, etc manually then you can run `make dev`, which will start those services as Docker containers. If you are setting up development environment for the first time, please run `make upgrade-successful` to skip the Postgresql upgrade.
```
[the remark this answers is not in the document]
sadiqkhoja: remove this extra step.

docker-compose.dev.yml
sadiqkhoja: done.

files/enketo/start-enketo.sh
alxndrsn: Having two different methods of substituting variables into the template seems confusing.  Are there other approaches that could work?  A couple of ideas:

1. in the test env, mount a local directory at `/etc/secrets` and store these simple values there, or
2. in this script:

```bash
if [ "$ENV" = "DEV" ]; then
  SECRET="s0m3v3rys3cr3tk3y"
  ..
else
  SECRET="$(cat /etc/secrets/enketo-secret)"
  ...
fi

envsubst ...
```

sadiqkhoja: Implemented the first approach.

docker-compose.dev.yml
```diff
@@ -0,0 +1,58 @@
...
+    profiles:
+      - none
+  nginx:
+    profiles:
+      - none
+  pyxform:
+    profiles:
+      - central
+    # changing port here - can't use `ports` mapping with host networking
+    command: ["gunicorn", "--bind", "0.0.0.0:5001", "--workers", "5", "--timeout", "600", "--max-requests", "1", "--max-requests-jitter", "3", "main:app()"]
```
alxndrsn, line 30: This seems like a significant deviation from the parent `docker-compose.yml`.  Is this running `pyxform` from source?

files/enketo/start-enketo.sh
```diff
@@ -3,6 +3,12 @@
 CONFIG_PATH=${ENKETO_SRC_DIR}/config/config.json
 echo "generating enketo configuration..."
 
+if [ "$ENV" = "DEV" ]; then
+    sed -i -e 's/enketo_redis_main/localhost/g' \
+       -e 's/enketo_redis_cache/localhost/g' \
+       -e 's/6380/6379/g' "$CONFIG_PATH.template"
+fi
```
alxndrsn, line 10: Is this change to reduce the number of redis containers required?  If so, is the increased complexity in these scripts and the deviation from the standard `docker-compose.yml` worth the savings?

files/postgres14/start-postgres.sh
alxndrsn: Would an alternative approach be to have a `postgres` container which creates the expected file and then exits?  Would this be closer to real system behaviour?

What's the significance of the change on this line?  What's the intention of `readlink`, and why is it now required?

Does `readlink -f -m` behave different from `readlink -m`?  It's not 100% clear from the manpage:

```
       -f, --canonicalize
              canonicalize  by following every symlink in every compo‐
              nent of the given name recursively;  all  but  the  last
              component must exist
...
       -m, --canonicalize-missing
              canonicalize  by following every symlink in every compo‐
              nent of the given name recursively, without requirements
              on components existence
```

but the source implies `-m` overrides `-f`:

https://github.com/coreutils/coreutils/blob/v9.1/src/readlink.c#L117-L122

```c
        case 'f':
          can_mode = CAN_ALL_BUT_LAST;
          break;
        case 'm':
          can_mode = CAN_MISSING;
          break;
```

Makefile
```diff
@@ -0,0 +1,8 @@
+.PHONY: dev
+dev:
+	docker compose --profile central -f docker-compose.yml -f docker-compose.dev.yml up -d
+	docker compose exec postgres14 touch /var/lib/odk/postgresql/14/.postgres14-upgrade-successful
+
+.PHONY: stop
+stop:
+	docker compose stop
```
alxndrsn, line 8: Does this command need to stop the dev profile?  E.g.

```sh
docker compose --profile central -f docker-compose.yml -f docker-compose.dev.yml stop
```

docker-compose.dev.yml
sadiqkhoja: It is just overriding the port in the command of the pyxform-http docker file: https://github.com/getodk/pyxform-http/blob/master/Dockerfile

Makefile
sadiqkhoja: I don't think that necessary. `docker compose stop` will stop all containers associated with `docker-compose.yml`

files/enketo/start-enketo.sh
sadiqkhoja: Agree. Originally I wanted this PR to facilitate Enketo development as well and Enketo suggests to run only one redis in development for the easiness. I have now removed the port substitution from this command and starting two redis instances. 

I still need to substitute `enketo_redis_main` and `enketo_redis_cache` with localhost because of host network mode.

Makefile
alxndrsn: :+1: this is fine as along as the two yaml files define the same services.

docker-compose.dev.yml
alxndrsn: Could the standard `command` be used, and this container's port `80` be exposed locally on port `5001`?  That would seem to simplify this section, and also protect against future divergence.

Makefile
alxndrsn: This could be abstracted via something like:

```makefile
docker-compose-dev := docker compose --profile central -f docker-compose.yml -f docker-compose.dev.yml

.PHONY: dev
dev:
	$(docker-compose-dev) up -d

.PHONY: stop
stop:
	$(docker-compose-dev) stop
```

files/enketo/start-enketo.sh
alxndrsn: > I still need to substitute `enketo_redis_main` and `enketo_redis_cache` with localhost because of host network mode.

Similar to `pyxform` config above, it would be great if there was a way to achieve this via port mapping rather than having to modify `files/enketo/start-enketo.sh` and make fragile changes with `sed`.

Makefile
sadiqkhoja: done

docker-compose.dev.yml
sadiqkhoja: I can't expose `ports` when using `host` networking mode. And `host` networking is required so that Enketo can communicate with Central via `http://localhost:8989`, which is running on the host.

I know we can use `host.docker.internal` to access services running on the host from the `bridge` container but configuration in Enketo and Central doesn't really support that. For example, Central uses `env.domain` to generate URLs in OpenRosa manifest, whose value is `http://localhost:8989` in local environment.

I had tried modifying `/etc/hosts` file in the Enketo container to map `localhost` to the IP assigned by docker to the `host.docker.internal` but Enketo using discontinued/deprecated `request.js` doesn't respects values in host file.

alxndrsn: > using discontinued/deprecated request.js doesn't respects values in host file.

Have you got a source for this?  I couldn't find anything obvious in the `request` issue tracker.

sadiqkhoja: The way requestjs resolves domain name has to do with it:

https://github.com/request/request/issues/2491

Note: above issue closed due to staleness

alxndrsn: I'm having trouble reproducing this issue.  Is it possible this has been resolved by enketo's move to `@cypress/request` (introduced by [37b8f8b2f0b057e7adaa2da472f88ab8a5e460e0](https://github.com/enketo/enketo/commit/37b8f8b2f0b057e7adaa2da472f88ab8a5e460e0#diff-60c4862d0a71d77cda5d4a8894ed75d100a0697310c2f7f63651307db71be798L59) in April 2024)?

sadiqkhoja: I was trying to map `localhost` to the IP Address of `host-gateway` in Enketo container, that didn't work. Localhost was still resolving to 127.0.0.1. 

So I ended up adding a custom domain in my local hosts file and changing `default.env.domain` to that host and now everything works. I have updated the instructions in readme file as well.

files/enketo/start-enketo.sh
sadiqkhoja: removed.

docker-compose.dev.yml
```diff
@@ -0,0 +1,62 @@
...
+    ports:
+      - 8005:8005
+  enketo_redis_main:
+    profiles:
+      - central
+  enketo_redis_cache:
+    profiles:
+      - central
+volumes:
+  dev_secrets:
```
alxndrsn, line 62: any chance of a trailing newline?

alxndrsn:
Looking nicely self-contained now :+1:

alxndrsn (approved):

PR #537, opened 2023-10-27
Release v2023.5

matthew-white:
This PR prepares the release of v2023.5. It should only contain changes from other PRs that have already been approved and merged (and possibly merge commits from the `master` branch).


matthew-white:
I think the only thing left to add here are submodule updates immediately before release, so I think this PR is now ready for review. @ktuite, I've tagged you. I think the only thing that needs to be verified is the timeline above. Did all commits come from approved PRs?

ktuite (approved):

Issue #519, opened 2023-10-02
Install requires a lot of space


lognaturel:
Related to https://github.com/getodk/docs/issues/1668, for now we'll document this, but it would be great to improve.

Some drivers I can think of:
- the postgres migration still being in play. We should make a plan to remove it. Maybe in mid-2024?
- Enketo and Central using different node base images. Hopefully that will be addressed soon and won't need to happen again
- frontend needing to be rebuilt for several config changes. Would be great if we could ship a built artifact instead.

matthew-white:
> frontend needing to be rebuilt for several config changes.

A config change specific to Frontend should be pretty rare, right? Currently that's just for SSO I believe. I think `docker image prune` will also recapture that space.

matthew-white:
Ah I see, even if you change a config to Backend, Frontend will get rebuilt. But as long as `docker image prune` is run, won't the user be OK?

lognaturel:
> as long as docker image prune is run, won't the user be OK?

I think so "at rest"? But they'll need more space than maybe expected at some point in time. I'm not entirely sure how big the impact is.

spwoodcock:
For FMTM at HOT I bundle ODKCentral into our stack.

To do that I have a customised backendDockerfile based on the one in this repo (service.dockerfile), to optimise the build.
Building the image from this repo --> **1.2GB**
Building the optimised image --> **372MB**

## Base Image

A good starting point is using the `slim` images provided by node.
The `node:18.17` image is 352MB, while `18.7-slim` is 75.37MB currently.
It just requires three additional packages to be installed (not bundled in `slim`):
- gpg
- procps
- netcat-traditional

## Efficient Layering

Dockerfile directives should be bundled together where possible for more efficient layer caching, for two reasons:
- Each new directive generates a new layer. There is overhead associated with very layer, and inefficiencies when stacking multiple layers. Also if a previous layer generates temp files that are not cleaned up, then they are carried forward to the next layer.
- Each layer is generated as an intermediate image during build. These bloat the filesystem (stored in the Docker image cache) & will increase the space required to actually do the build.

## Frontend

I also bundle a `central-frontend`. It's only 74.9MB.

## Contribution

Would it be useful for me to make a PR, is someone already working on this?

Also, as a separate but related issue, it would be great to have the images built and published to something like Github container registry. Is this something you would consider?

The only reason I had to generate my custom Dockerfiles is because there is no official distribution of ODK Container Images.
[...]

PR #549, opened 2023-11-27
Use slim node images to reduce download and running size

spwoodcock:
## Includes

- Built using `node:18-slim` instead of `node:18.17`.
  - I assumed we probably want patch versions of node that shouldn't be breaking.
  - Is the assumption correct, or should I change back to pinned `18.17`?
- Merged RUN directives where possible
  - This was less significant than I thought, maybe 50 - 100 MB saving.
  - The dockerfile was missing a cleanup of the apt list here too: https://github.com/getodk/central/blob/774b491961578bd6e03ceaa6cf8032dd27b828a7/service.dockerfile#L21-L22
- Absolute paths for WORKDIR (as per dockerfile-utils linting).
- Add HEALTHCHECK. This is useful for docker-compose setups. It's ignored in Kubernetes.
- Remove deprecated EXPOSE command and replace with LABELs.

## Discussion

I image the reason for not using a slim image originally is because they don't have git installed for the
`git describe --tags --dirty` command. Solved with multi-stage.

This PR was originally discussed as part of #519.
Reducing the space required to build images may be useful on systems with very limited space.
However, the gains that can be made are rather small considering >1 GB of cached images shouldn't really be a problem for most.

There have been discussions around providing pre-built images and an open PR to facilitate this, so this discussion will be moot if that is merged.

There is always a tradeoff between cache image bloat and dockerfile readability.
However, in this example I personally think the original is much more readable:

```dockerfile
FROM node:${node_version}-slim as intermediate
RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        git \
    && rm -rf /var/lib/apt/lists/*
COPY . .
RUN mkdir /tmp/sentry-versions
RUN git describe --tags --dirty > /tmp/sentry-versions/central
WORKDIR /server
RUN git describe --tags --dirty > /tmp/sentry-versions/server
WORKDIR /client
RUN git describe --tags --dirty > /tmp/sentry-versions/client
```

compared to less readable & not best practice of `cd` in dockerfiles (for little gain):

```dockerfile
FROM node:${node_version}-slim as intermediate
COPY . .
RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        git \
    && rm -rf /var/lib/apt/lists/* \
    && mkdir /tmp/sentry-versions \
    && git describe --tags --dirty > /tmp/sentry-versions/central \
    && cd server && git describe --tags --dirty > /tmp/sentry-versions/server \
    && cd ../client && git describe --tags --dirty > /tmp/sentry-versions/client
```

> Note that this is not the final build stage, so the discussion is about image cache bloat,
> not intermediate image layer size.

### DIFF
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -1,30 +1,68 @@
-ARG node_version=18.17
-FROM node:${node_version} as intermediate
+ARG node_version=18
 
+
+
+FROM node:${node_version}-slim as pgdg
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
+        gpg \
+    && rm -rf /var/lib/apt/lists/* \
+    && update-ca-certificates
+RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODENAME=\K\w+' /etc/os-release)-pgdg main" \
+      | tee /etc/apt/sources.list.d/pgdg.list \
+    && curl https://www.postgresql.org/media/keys/ACCC4CF8.asc \
+      | gpg --dearmor > /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg
+
+
+
+FROM node:${node_version}-slim as intermediate
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        git \
+    && rm -rf /var/lib/apt/lists/*
 COPY . .
 RUN mkdir /tmp/sentry-versions
 RUN git describe --tags --dirty > /tmp/sentry-versions/central
-WORKDIR server
+WORKDIR /server
 RUN git describe --tags --dirty > /tmp/sentry-versions/server
-WORKDIR ../client
+WORKDIR /client
 RUN git describe --tags --dirty > /tmp/sentry-versions/client
 
-FROM node:${node_version}
+
+
+FROM node:${node_version}-slim
+
+ARG node_version
+LABEL org.getodk.central.app-name="central" \
+      org.getodk.central.node-tag="${node_version}" \
+      org.getodk.central.maintainer="admin@hotosm.org" \
+      org.getodk.central.port="8383"
 
 WORKDIR /usr/odk
 
-RUN apt-get update && apt-get install wait-for-it && rm -rf /var/lib/apt/lists/*
-
-RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODENAME=\K\w+' /etc/os-release)-pgdg main" | tee /etc/apt/sources.list.d/pgdg.list && \
-  curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | gpg --dearmor > /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg && \
-  apt-get update && \
-  apt-get install -y cron gettext postgresql-client-14
+COPY --from=pgdg /etc/apt/sources.list.d/pgdg.list \
+    /etc/apt/sources.list.d/pgdg.list
+COPY --from=pgdg /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg \
+    /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        gpg \
+        cron \
+        wait-for-it \
+        gettext \
+        procps \
+        postgresql-client-14 \
+        netcat-traditional \
+    && rm -rf /var/lib/apt/lists/*
 
 COPY files/service/crontab /etc/cron.d/odk
 
 COPY server/package*.json ./
 
-RUN npm clean-install --omit=dev --legacy-peer-deps --no-audit --fund=false --update-notifier=false
+RUN npm clean-install --omit=dev --legacy-peer-deps --no-audit \
+        --fund=false --update-notifier=false
 
 COPY server/ ./
 COPY files/service/scripts/ ./
@@ -34,5 +72,5 @@ COPY files/service/odk-cmd /usr/bin/
 
 COPY --from=intermediate /tmp/sentry-versions/ ./sentry-versions
 
-EXPOSE 8383
-
+HEALTHCHECK --start-period=10s --interval=5s --retries=10 \
+    CMD nc -z localhost 8383 || exit 1
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -35,9 +35,7 @@ RUN git describe --tags --dirty > /tmp/sentry-versions/client
 FROM node:${node_version}-slim
 
 ARG node_version
-LABEL org.opencontainers.image.title="ODK Central backend" \
-      org.opencontainers.image.vendor="ODK" \
-      org.opencontainers.image.source="https://github.com/getodk/central"
+LABEL org.opencontainers.image.source="https://github.com/getodk/central"
 
 WORKDIR /usr/odk
 
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -37,7 +37,7 @@ FROM node:${node_version}-slim
 ARG node_version
 LABEL org.getodk.central.app-name="central" \
       org.getodk.central.node-tag="${node_version}" \
-      org.getodk.central.maintainer="admin@hotosm.org" \
+      org.getodk.central.maintainer="support@getodk.org" \
       org.getodk.central.port="8383"
 
 WORKDIR /usr/odk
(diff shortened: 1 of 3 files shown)


lognaturel:
Thanks for opening! I'm not entirely sure when we'll be able to take a focused look at this and #546, it may be after the current release goes out.

> Built using node:18-slim instead of node:18.17

We do want to pin the minor version because there are more likely to be accidental breaking changes between minor versions and usually they're not critical to apply. We update minor versions when we release. We do want to leave the patch version floating so that a rebuild can get the latest patches (so we don't want e.g. node:18.17.1).

service.dockerfile
```diff
@@ -1,38 +1,73 @@
...
 
 COPY files/service/config.json.template /usr/share/odk/
+COPY files/service/crontab /etc/cron.d/odk
 COPY files/service/odk-cmd /usr/bin/
 
 COPY --from=intermediate /tmp/sentry-versions/ ./sentry-versions
 
-EXPOSE 8383
-
+HEALTHCHECK --start-period=10s --interval=5s --retries=10 \
```
lognaturel, line 72: Let's separate out the healthcheck concern. An interval of 5s sounds aggressive to me but maybe it doesn't matter because it's not resource-intensive. It does seem like a bit of an oversight not to have any explicit simple healthcheck but if we do add one that just checks if there's a response at the port maybe it could go outside (e.g. in the Docker Compose config). We've also considered adding a healthcheck endpoint in backend. Let's sort it out in an issue! For deployments we tend to use entirely separate services for health monitoring.

lognaturel (changes requested):
Couple of quick things to update. Other than that, this looks ready to me. Current plan is to merge it right after the release ~Dec 15. We may also end up publishing images in which case it's less important but let's still get it in first.

service.dockerfile
```diff
@@ -1,38 +1,73 @@
...
+FROM node:${node_version}-slim
 
-RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODENAME=\K\w+' /etc/os-release)-pgdg main" | tee /etc/apt/sources.list.d/pgdg.list && \
-  curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | gpg --dearmor > /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg && \
-  apt-get update && \
-  apt-get install -y cron gettext postgresql-client-14
+ARG node_version
+LABEL org.getodk.central.app-name="central" \
+      org.getodk.central.node-tag="${node_version}" \
+      org.getodk.central.maintainer="admin@hotosm.org" \
```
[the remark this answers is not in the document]
spwoodcock: Oops! Was clearly in a rush

service.dockerfile
spwoodcock: Good point - it's easier for the user to configure via docker compose instead of here 👍

lognaturel:
Before:
> docker inspect -f "{{ .Size }}" central-service | numfmt --to=si
1.3G

After:
> docker inspect -f "{{ .Size }}" central-service | numfmt --to=si
374M

Secrets goes from 1.1G to 201M and nginx only gets a change to the intermediate stage but that's still nice.

lognaturel:
> I image the reason for not using a slim image originally is because they don't have git installed

https://github.com/getodk/central/pull/249#issuecomment-949986638 has some context. The past attempt was to move to alpine. There used to be a lua script in the nginx container that didn't work well with that but it's been removed. We have historically been somewhat more concerned about download size than local storage size. The change to `20.10-slim` will **add** ~[60MB](https://hub.docker.com/layers/library/node/20.10-slim/images/sha256-18aacc7993a16f1d766c21e3bff922e830bcdc7b549bbb789ceb7374a6138480?context=explore) to the overall download because Enketo will pull the `20.10` image. I think that's ok and we can align Enketo in the next release.

This discussion made me realize we should use an alpine image for postgres14 so I added that change.

service.dockerfile
```diff
@@ -68,3 +67,5 @@ COPY files/service/crontab /etc/cron.d/odk
 COPY files/service/odk-cmd /usr/bin/
 
 COPY --from=intermediate /tmp/sentry-versions/ ./sentry-versions
+
+EXPOSE 8383
```
lognaturel, line 71: The original PR description states that `EXPOSE` is deprecated. I see no evidence of this: https://docs.docker.com/engine/reference/builder/#expose

service.dockerfile
```diff
@@ -35,10 +35,9 @@ RUN git describe --tags --dirty > /tmp/sentry-versions/client
 FROM node:${node_version}-slim
 
 ARG node_version
-LABEL org.getodk.central.app-name="central" \
-      org.getodk.central.node-tag="${node_version}" \
-      org.getodk.central.maintainer="support@getodk.org" \
-      org.getodk.central.port="8383"
+LABEL org.opencontainers.image.title="ODK Central backend" \
```
lognaturel, line 38: The only accepted standard for labels that I know of is [this one](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys). GHCR does recognize a few of them: https://docs.github.com/en/enterprise-server@3.8/packages/working-with-a-github-packages-registry/working-with-the-container-registry#labelling-container-images

lognaturel:
@yanokwa can you please do a sanity check? I'm satisfied that:
- it's a clean update
- there's user value
- it doesn't introduce anything non-standard
- it applies the improvement consistently and as broadly as it can

service.dockerfile
yanokwa: My bias is to only include `org.opencontainers.image.source`. It's what GHCR supports and so is likely to be useful. It's ambiguous what to put there, which I like. Anything else seems to be of minor value so I'd rather not commit to it.

service.dockerfile
yanokwa: Expose is for documenting the port the builder of the image intends to publish. It does actually publishing it, so maybe that's the confusion? Either way, agreed that we can leave this.

yanokwa (approved):

service.dockerfile
spwoodcock: Sorry, my mistake!

I always leave out EXPOSE and prefer a LABEL, as expose doesn't actually do anything to the image.

Using a label allows a user to `docker inspect` to easily know which port they need to bind.

yanokwa: It's not a bad idea. Is it used anywhere else? Or documented somewhere as a best practice?

spwoodcock: Not as far as I can see!

For a long time LABEL had no standard set, so it's good to see opencontainers create one.

I added the port label from my own frustrations with using other images.
If the author doesn't have a label, then I had to hunt down either the dockerfile (if they use EXPOSE) or the application config to find the port.

**Updates in my knowledge**

- LABEL and ENV no longer require multiline definition. It's more readable to use individual lines. 

- EXPOSE is now used by both docker and podman `-P` flag, to map all exposed ports to random ports on the host (does not seem useful to me, but maybe to some?).

lognaturel: Thanks again for getting these improvements in!

When doing `docker inspect`, doesn't `Config/ExposedPorts` show the ports that were specified with `EXPOSE`?

spwoodcock: You are correct!
That's an oversight from me.

I thought that `EXPOSE` was a useless command for years.
By the looks of it, it is actually a nice way to document via the metadata.

lognaturel: I think so! And downstream tools could use it programmatically, hopefully to do more useful things than random port binding. 😄 🤷🏻‍♀️ It's also very possible that `Config/ExposedPorts` is relatively new. There's a lot to learn and keep track of!

lognaturel:
I just noticed that the Enketo image we produce automatically for GHCR gets some labels. I assume it's the Github action we use that does this:

```
  "labels": {
    "org.opencontainers.image.url": "https://github.com/enketo/enketo",
    "org.opencontainers.image.licenses": "Apache-2.0",
    "org.opencontainers.image.title": "enketo",
    "org.opencontainers.image.revision": "2f27f07b9d72eb27fb8576570501682b2f3ca5d2",
    "org.opencontainers.image.created": "2024-01-26T20:11:10.683Z",
    "org.opencontainers.image.version": "7.1.0",
    "org.opencontainers.image.description": "Enketo web forms monorepo",
    "org.opencontainers.image.source": "https://github.com/enketo/enketo"
  }
```

looks like `version` uses a tag if it exists or falls back to the branch name. I have not yet investigated whether this or what's specified in a Dockerfile takes precedence.

spwoodcock:
Interesting - the part of the workflow that does it is here: https://github.com/enketo/enketo/blob/2f27f07b9d72eb27fb8576570501682b2f3ca5d2/.github/workflows/ghcr.yml#L50

I use docker/build-push-action a lot, but only ever set the tags from docker/metadata-action, not the labels 🤦

lognaturel:
Looks like a good approach, we can look at adding that to https://github.com/getodk/central/pull/546

Issue #347, opened 2022-11-10
Simplify deployments/upgrades


alxndrsn:
Ideas etc. for simplifying the deployment/upgrade process:

* cut git out of the install?
* pre-building all images?
  - make builds reproducible
  - make installs/upgrades faster(?)
  - less disk space required on host machines(?)
  - use github container registry to serve images?
  - may have to allow more configuration to be done via env var, e.g.
    - changing database
    - disabling sentry
    - different SSL termination options(?)

matthew-white:
> pre-building all images?

See also #165 and #249.

> may have to allow more configuration to be done via env var, e.g.

We don't document this, but it's currently possible to configure Frontend by modifying src/config.js (you can see that on the QA server, for example).

matthew-white:
>   * may have to allow more configuration to be done via env var, e.g.
>     
>     * changing database

Related: #389

spwoodcock:
Currently I build a custom image for the central-backend.
I have to duplicate the Dockerfile in my repo, as using a setup like this won't work with submodules:

```yaml
services:
  central:
    build: https://raw.githubusercontent.com/getodk/central/master/service.dockerfile
    ...
```

The COPY directives will all fail due to the submodule files not being present.

I would propose a solution like this:

```dockerfile
ARG node_version=18


FROM docker.io/bitnami/git:2 as repo
ARG ODK_CENTRAL_TAG
RUN git clone --depth 1 --branch ${ODK_CENTRAL_TAG} \
    "https://github.com/getodk/central.git" \
    && cd central && git submodule update --init


FROM docker.io/node:${node_version}-slim
WORKDIR /usr/odk
COPY --from=repo central/files/service/crontab /etc/cron.d/odk
COPY --from=repo central/files/service/scripts/ ./
...
```

What are your thoughts?

alxndrsn:
> Currently I build a custom image for the central-backend.
> I have to duplicate the Dockerfile in my repo, as using a setup like this won't work with submodules:

Hi @spwoodcock, it would be great to understand better what you're changing in your custom docker image.  Can you share a bit more context?

spwoodcock:
Sure thing, the Dockerfile is:

```dockerfile
ARG node_version=18



FROM docker.io/bitnami/git:2 as repo
ARG ODK_CENTRAL_TAG
RUN git clone --depth 1 --branch ${ODK_CENTRAL_TAG} \
    "https://github.com/getodk/central.git" \
    && cd central && git submodule update --init



[...]

PR #588, opened 2024-01-16
Determining backend memory limit from container cgroup

### LINKED ISSUE #577: Central sets WORKER_COUNT based on host, not container resources
In https://github.com/getodk/central/blob/HEAD/files/service/scripts/start-odk.sh#L33 we set `WORKER_COUNT` based on total memory. 

According to https://forum.getodk.org/t/limiting-resources-memory-that-the-odk-central-service-can-consume/44357/2 that total memory is the host's memory not the container's memory, so you can have a situation where you are running Central in a memory limited container but using more workers than is recommended.

My guess is that most people don't allocate less than 1 GB, so this isn't urgent, but I wanted to document it.

spwoodcock:
Closes #577

# * Original Comment *

#### Includes

- Correctly export the `WORKER_COUNT` variable in `start-odk.sh`, otherwise the value is not respected (see screenshots).
  - This is because `pm2.config.js` uses `process.env.WORKER_COUNT`, but the variable is only available in the bash shell, not the pm2 command (it needs to be exported for that).
- Fixed the determination of the available memory:
  - Currently the `vmstat` command gets the total available memory on the **host system**.
  - To get the available memory from either the host machine OR a restricted container in docker compose / Kubernetes, we need to inspect the cgroup.
- Added an optional for environment variable `WORKER_COUNT` for the backend service.
  - If this value is provided, it overrides the automatically generated value for `WORKER_COUNT`.
  - Added the empty variable to `.env.template` so users know it's an option.
  - The rationale for adding this is twofold:
    - During development it's good to be able to specify 1 worker, even on a machine with large memory.
    - If running in Kubernetes, it is generally not good practice to spawn multiple workers per pod. Ideally 1 process/worker per pod allows for easier load management, autoscaling, etc.
- Minor pedantic bash syntax fix: moved a `then` to the same line as `if` (in line with shellcheck and bashate linter recommendations).

#### Screenshots

**Before: 4 workers determined, but only one runs**

Log:
four_pm2_workers_specified

System info:
one_pm2_process

**After: 4 workers determined, 4 workers run**

Log:
four_pm2_workers_correct

System info:
four_pm2_processes

- Tests inside docker containers running on a Debian Bookworm host.
- This has not been tested via Kubernetes, but I see no reason why the underlying approach would not work.
- Also not tested on Docker Desktop or similar, but as they rely on emulation of a Linux VM, this should also work fine.

- This is the most straightforward solution to determine the memory restrictions from a cgroup.
- The official nginx image has a [best practice](https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/30-tune-worker-processes.sh) approach that is quite convoluted, as it also involves determining other metrics.

- This will not affect users running on systems with memory ~1GB memory.
- However, as it fixes the worker allocation, it will affect users with >1GB memory.
- Users may find that ODK Central takes more memory to run that it did previously, due to correctly spawning 4 processes.

- It probably does require a docs update. I can make a PR.
- It would be useful to document the `WORKER_COUNT ` environment variable is available.

# * Updated Comment *

- Decided not to go ahead with `WORKER_COUNT` env var configuration, as a better approach is running one process per container.
- Container replicas can be configured via docker compose instead.
- This PR includes the total memory check via container cgroup, which should work in containers run both via Docker and Kubernetes.

### DIFF
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -30,7 +30,7 @@ fi
 echo "starting cron.."
 cron -f &
 
-MEMTOT=$(vmstat -s | grep 'total memory' | awk '{ print $1 }')
+MEMTOT=$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)
 if [ "$MEMTOT" -gt "1100000" ]
 then
   export WORKER_COUNT=4


spwoodcock:
Oops, branched off an old `next` commit. Will update 👍

lognaturel:
Thanks so much for fixing the memory limit check!

I'm surprised to see the issue related to pm2 workers. I see 4 node processes on the 2Gb servers I checked. Everything else was stock when you identified that problem?

spwoodcock:
Hmm if that's the case, then further debugging is required!

I'm surprised how pm2.config.js would be able to determine 4 workers from process.env, as setting without an export would mean the variable is not available in process.env.

I tested this as part of my dev setup for working on Central, using `next` branch for this repo and `master` branches for backend and frontend.

I use the Dockerfiles and compose file from this repo, with the certs set to `upstream` in the dotenv.

spwoodcock:
Also, updated my original comment to explain the rationale for adding the WORKER_COUNT as a configurable environment variable. It's especially important if running in Kubernetes.

spwoodcock:
I just tested this on a temp machine:

image

The 4 workers started up. Now I'm puzzled why they don't start on my WSL Debian instance.

Either way, this functionality seems to work fine currently, so there is no concern than this will cause users to go from 1 --> 4 workers on their deployment.

Adding `export` to the WORKER_COUNT fixed the issue on my WSL Debian machine, and shouldn't have any other implications, so should I leave it in this PR, or remove it?

lognaturel:
That's very strange!

I'd prefer to split out the new env var, please:
- I don't think `WORKER_COUNT` is the right name at that level. Maybe `BACKEND_REPLICAS`?
- I'm not sure we want to continue running multiple processes in a container: https://github.com/getodk/central/issues/347#issuecomment-1908944507 Once we expose an env variable, we want to continue supporting it so I think we should first resolve that question.

Once that's split off the mem calculation change is ready to merge.

spwoodcock:
Done 👍

lognaturel (approved):

lognaturel:
Thank you! Good discussion all around. I'm glad you're liking the ideas at https://github.com/getodk/central/issues/347#issuecomment-1908944507 👍

PR #594, opened 2024-01-26
Prep v2023.5.1

lognaturel:
I have a draft forum post ready to go at https://forum.getodk.org/t/odk-central-v2023-5-entity-updates-via-submissions/44311/2

This branch is currently running on https://dev.getodk.cloud I verified converting a form with pyxform and making submissions with Enketo.

We've now fixed all known issues coming out of the pyxform 2.0.0 changes so it feels like a good time to upgrade pyxform.

Enketo as currently in Central v2023.5.0 changes annotation files any time an edit is made to the corresponding submission, whether or not the edit involved changes to the annotation file. That seems worth fixing asap.

I believe the pyxform change is very low risk. The Enketo change will be somewhat disruptive because it rolls back some improvements but I think it's the right call (see https://github.com/enketo/enketo/pull/1292)

No

- [ ] branched off and targeted the `next` branch OR only changed documentation/infrastructure (`master` is stable and used in production) -- Branching off master because this is intended to be released right away.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -102,7 +102,7 @@ services:
       options:
         max-file: "30"
   pyxform:
-    image: 'ghcr.io/getodk/pyxform-http:v2.0.0'
+    image: 'ghcr.io/getodk/pyxform-http:v2.0.2'
     restart: always
   secrets:
     volumes:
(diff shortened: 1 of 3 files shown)


matthew-white (approved):
Looks good! Just the date of the news could use an update.

PR #595, opened 2024-01-29
Release v2024.1

matthew-white:
This PR prepares the release of v2024.1. It should only contain changes from other PRs that have already been approved and merged (and possibly merge commits from the `master` branch).


matthew-white:
I'm noticing that the submodules were rolled back to v2023.3.1, I think in #617. That shouldn't be an issue, because we'll update the submodules to the latest before release. Just thought I'd make a note of it so that no one's confused.

ktuite (approved):

PR #617, opened 2024-03-21
Upgrade to non-tagged release of nginx

yanokwa:
Point release is out so we should use that


lognaturel (approved):

lognaturel:
Maybe we should wait for dependency updates until we have scheduled regression testing. That's generally what I do since things can change quickly.

PR #619, opened 2024-03-27
Check if running container was launched with compose v1

yanokwa:
As shown at https://forum.getodk.org/t/how-to-upgrade-central-from-v1-5-3-to-latest/45797/10

The check disk command doesn't work because even if you have compose v2 installed, if the container was launched with v1, it will have underscores in the name. This patch checks to see if there is a running container with underscores and returns a 1 or 0.

I haven't tested this on a real system, but I have verified the individual docker and grep commands work as expected.


lognaturel (approved):

lognaturel:
I agree with the approach and have also verified the command. I don't think we need further verification than that.

PR #636, opened 2024-04-15
Leaving version in docker-compose results in warning

yanokwa:
Recent versions of Docker will show a warning if there is a version key. Removing the key fixes the problem.

Screenshot 2024-04-15 at 14 56 35

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3"
 services:
   postgres14:
     build:


lognaturel (approved):

matthew-white:
I think it'll be worth mentioning this PR in the release notes, since the warning is visible to users. I'll add the PR to the project so that it makes its way into the release notes.

Issue #572, opened 2023-12-19
Update usage information metrics for v2024.1


matthew-white:
**Usage information metrics to add/remove/update**

- [x] Add metrics for the following:
  - Number of `entity.bulk.create` events
  - Biggest size of bulk upload (number of rows/entities created)
- [x] Remove the following metrics:
  - (none)

---

**Checklist**

- [x] Update the form version string in the Backend config.
- [x] Publish the new form version on data.getodk.cloud.
- [x] Update the usage report in Frontend.

PR #638, opened 2024-04-21
Update Node version for service

matthew-white:
#637 updated Node for `nginx` and `secrets`, but I think we need to do so for `service` as well.

### DIFF
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -1,4 +1,4 @@
-ARG node_version=20.10
+ARG node_version=20.12.2
 
 
 


lognaturel:
Sigh. Thank you! I'm not sure why I seem to miss something every time. There are so many steps that go into the pyxform and Enketo releases that maybe I just don't have enough focus once I get here. Will try to get a better system in place.

matthew-white:
We use a checklist for the issue for usage information metrics that's part of each release, e.g., #572. Maybe a similar checklist would help for the issue for updating infrastructure dependencies?

lognaturel:
Sounds good, let me write one for the next release. Let's combine all of the infrastructure updates including pyxform and Enketo into one issue.

Issue #656, opened 2024-05-01
Prebuild Frontend


matthew-white:
It's taking increasingly long to build Frontend. It'd be nice if we could ship built Frontend assets rather than requiring users to build Frontend. We just discussed a specific approach to this:

- [x] 1. Rather than include the [Frontend config](https://github.com/getodk/central-frontend/blob/3fb0c22b1cbdc3a6004963afcc3847a82c09307d/src/config.js) in the Frontend bundle, send a request for the config when Frontend is loaded.
- [x] 2. Serve the Frontend config from nginx. We could template it in a similar way as we do the Backend config.
- [x] 3. Build Frontend automatically.

I'm going to work on (1) and (2), and @lognaturel is going to take a look at (3).

Issue #473, opened 2023-08-31
There is no need to explicitly list variables in envsubst


sadiqkhoja:
https://github.com/getodk/central/blob/f09c50bc3cd7537545916c25b792079c072a6ae1/files/service/scripts/start-odk.sh#L7

matthew-white:
Closed by #830.

PR #657, opened 2024-05-02
Generate Frontend config and serve it

matthew-white:
This PR makes progress on getodk/central#656. It generates the Frontend config and serves it so that Frontend can then fetch it. The corresponding Frontend PR is getodk/central-frontend#988.

So far just that CI passes. Once it's on staging, I can double-check that /client-config.json is served correctly.

I'm going to leave comments about a few lines.

getodk/central-frontend#988 mentions the cost of an additional request. I don't think it will be much of an issue.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -67,6 +67,10 @@ server {
       include /usr/share/odk/nginx/common-headers.conf;
       add_header Cache-Control no-cache;
     }
+    location /client-config.json {
+      include /usr/share/odk/nginx/common-headers.conf;
+      add_header Cache-Control no-cache;
+    }
     location /index.html {
       include /usr/share/odk/nginx/common-headers.conf;
       add_header Cache-Control no-cache;
--- a/nginx.dockerfile
+++ b/nginx.dockerfile
@@ -8,9 +8,9 @@ RUN apt-get update \
 
 COPY ./ ./
 RUN files/prebuild/write-version.sh
-ARG OIDC_ENABLED
-RUN envsubst < files/nginx/client-config.json.template > /tmp/client-config.json
 RUN files/prebuild/build-frontend.sh
+ARG OIDC_ENABLED
+RUN files/prebuild/write-client-config.sh
 
 
 
--- a/nginx.dockerfile
+++ b/nginx.dockerfile
@@ -3,12 +3,14 @@ FROM node:20.12.2-slim as intermediate
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         git \
+        gettext-base \
     && rm -rf /var/lib/apt/lists/*
 
 COPY ./ ./
 RUN files/prebuild/write-version.sh
 ARG OIDC_ENABLED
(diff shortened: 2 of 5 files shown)


files/nginx/odk.conf.template
```diff
@@ -67,6 +67,10 @@ server {
       include /usr/share/odk/nginx/common-headers.conf;
       add_header Cache-Control no-cache;
     }
+    location /client-config.json {
```
matthew-white, line 70: This `location` block and the ones above and below it are identical. However, I don't see a way to specify a list to `location`. We could use a regular expression, but I'm not sure that that's preferable.

nginx.dockerfile
```diff
@@ -3,12 +3,14 @@ FROM node:20.12.2-slim as intermediate
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         git \
+        gettext-base \
```
matthew-white, line 6: Below I run `envsubst`. Originally I did so without installing `gettext-base`. However, the build resulted in an error:

```
/bin/sh: 1: envsubst: not found
```

I'm not sure the best way to get `envsubst`. [This answer](https://stackoverflow.com/a/63746812/2420970) on StackOverflow suggested installing `gettext` as a way to get it. I don't know much about `gettext`, but I saw that enketo.dockerfile installs `gettext-base`. Given that enketo.dockerfile installs it, I figured it'd be OK to do so here as well.

An alternative would have been to wait to run `envsubst` further below. setup-odk.sh runs `envsubst`, so it must get installed somehow. However, I thought it was clearer overall to run `envsubst` here in `intermediate`. `intermediate` is where version.txt and the rest of the Frontend assets are generated, so it feels like the right place to generate client-config.json.

nginx.dockerfile
```diff
@@ -3,12 +3,14 @@ FROM node:20.12.2-slim as intermediate
...
     && rm -rf /var/lib/apt/lists/*
 
 COPY ./ ./
 RUN files/prebuild/write-version.sh
 ARG OIDC_ENABLED
-RUN OIDC_ENABLED="$OIDC_ENABLED" files/prebuild/build-frontend.sh
+RUN envsubst < files/nginx/client-config.json.template > /tmp/client-config.json
```
matthew-white, line 12: Elsewhere, we specify a list of variables to `envsubst. However, with #473 in mind, I didn't do so here.

ktuite (approved):

files/nginx/odk.conf.template
matthew-white, line 70: Add `Content-Type` header?

matthew-white: I think this isn't needed and that nginx will automatically the header. I'll check once the PR is on staging though.

matthew-white: I just checked staging, and I can see that `Content-Type` is `application/json`.

matthew-white:
> Once it's on staging, I can double-check that /client-config.json is served correctly.

I can see that client-config.json is served and that its content matches my expectation: https://staging.getodk.cloud/client-config.json

PR #664, opened 2024-05-13
Update image used for CircleCI

matthew-white:
When I view a build in CircleCI, I see the following warning (e.g., [here](https://app.circleci.com/pipelines/github/getodk/central/927/workflows/cdc558cb-6857-419e-9fc2-b555b2069a07/jobs/972)):

> This job is using a [deprecated](https://discuss.circleci.com/t/linux-image-deprecations-and-eol-for-2024/) image 'ubuntu-2004:202201-02', please update to a newer image

This PR resolves the warning by updating the image.

CircleCI continues to pass.

This PR targets the current version of Ubuntu 20.04. If we were concerned that we might introduce something to Central that would work in a new version of Ubuntu 20.04, but not an older one, we could intentionally target an older version. It looks like one of the [options](https://circleci.com/docs/linux-vm-support-policy/) is `previous`, which is 3–6 months old, though I think that's Ubuntu 22.04. That link says, "We recommend using the `default` version and not pinning to a date version." The list of images available for Ubuntu 20.04 is [here](https://circleci.com/developer/machine/image/ubuntu-2004).

Relatedly, I saw that Ubuntu 20.04 will reach EOL in 2025. With that in mind, I tried specifying `image: default`. It looks like that's currently `ubuntu-2204:2024.04.4`. However, when I do so, CircleCI [fails](https://app.circleci.com/pipelines/github/getodk/central/929/workflows/38d6cca0-f985-43d1-a37a-eec98c8b9cc4/jobs/974). From a glance, it looks like pm2 isn't working as expected. That sounds like it could be an issue worth looking into. I'm not able to do so at this particular moment, but I'd be happy to file an issue about it if that'd be helpful.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     machine:
-      image: ubuntu-2004:202201-02
+      image: ubuntu-2004:current
 
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,4 +37,9 @@ jobs:
               appropriate/curl -4 --insecure --retry 20 --retry-delay 2 --retry-connrefused https://localhost/v1/projects \
               | tee /dev/tty \
               | grep -q '\[\]'
-            docker compose exec -T service npx pm2 list | grep -c "online" | grep -q 4 || exit 1
+      - run:
+          name: Verify pm2
+          command: |
+            docker compose exec -T service npx pm2 list \
+              | tee /dev/tty \
+              | grep -c "online" | grep -q 4


matthew-white:
> I tried specifying `image: default`. It looks like that's currently `ubuntu-2204:2024.04.4`. However, when I do so, CircleCI [fails](https://app.circleci.com/pipelines/github/getodk/central/929/workflows/38d6cca0-f985-43d1-a37a-eec98c8b9cc4/jobs/974). From a glance, it looks like pm2 isn't working as expected.

A little more information on this. The output from `npx pm2 list` wasn't visible in that failure, so I redid it [here](https://app.circleci.com/pipelines/github/getodk/central/933/workflows/200ac70e-d3cb-4243-a222-cd43285f351c/jobs/980). I see:

```
+ docker compose exec -T service npx pm2 list
┌────┬────────────┬─────────────┬─────────┬─────────┬──────────┬────────┬──────┬───────────┬──────────┬──────────┬──────────┬──────────┐
│ id │ name       │ namespace   │ version │ mode    │ pid      │ uptime │ ↺    │ status    │ cpu      │ mem      │ user     │ watching │
├────┼────────────┼─────────────┼─────────┼─────────┼──────────┼────────┼──────┼───────────┼──────────┼──────────┼──────────┼──────────┤
│ 0  │ service    │ default     │ 0.1.0   │ cluster │ 77       │ 35s    │ 0    │ online    │ 0%       │ 85.2mb   │ root     │ disabled │
└────┴────────────┴─────────────┴─────────┴─────────┴──────────┴────────┴──────┴───────────┴──────────┴──────────┴──────────┴──────────┘
```

Compare that to the passing build [here](https://app.circleci.com/pipelines/github/getodk/central/931/workflows/3b3d0e66-f5dd-4099-90fc-9fd940d3b829/jobs/978):

```
+ docker compose exec -T service npx pm2 list
┌────┬────────────┬─────────────┬─────────┬─────────┬──────────┬────────┬──────┬───────────┬──────────┬──────────┬──────────┬──────────┐
│ id │ name       │ namespace   │ version │ mode    │ pid      │ uptime │ ↺    │ status    │ cpu      │ mem      │ user     │ watching │
├────┼────────────┼─────────────┼─────────┼─────────┼──────────┼────────┼──────┼───────────┼──────────┼──────────┼──────────┼──────────┤
│ 0  │ service    │ default     │ 0.1.0   │ cluster │ 76       │ 15s    │ 0    │ online    │ 0%       │ 117.5mb  │ root     │ disabled │
│ 1  │ service    │ default     │ 0.1.0   │ cluster │ 83       │ 15s    │ 0    │ online    │ 0%       │ 118.2mb  │ root     │ disabled │
│ 2  │ service    │ default     │ 0.1.0   │ cluster │ 90       │ 15s    │ 0    │ online    │ 0%       │ 117.2mb  │ root     │ disabled │
│ 3  │ service    │ default     │ 0.1.0   │ cluster │ 97       │ 15s    │ 0    │ online    │ 0%       │ 117.9mb  │ root     │ disabled │
└────┴────────────┴─────────────┴─────────┴─────────┴──────────┴────────┴──────┴───────────┴──────────┴──────────┴──────────┴──────────┘
```

matthew-white:
> The output from `npx pm2 list` wasn't visible in that failure

I just pushed a commit that will display the output from `npx pm2 list`. It takes the same approach as right above in the CircleCI config.

.circleci/config.yml
```diff
@@ -37,4 +37,9 @@ jobs:
               appropriate/curl -4 --insecure --retry 20 --retry-delay 2 --retry-connrefused https://localhost/v1/projects \
               | tee /dev/tty \
               | grep -q '\[\]'
-            docker compose exec -T service npx pm2 list | grep -c "online" | grep -q 4 || exit 1
```
matthew-white, line 40: I removed `exit 1`, because if `grep -q 4` doesn't find 4, it should already return with an exit status of 1. That alone will cause the command to exit, since CircleCI runs bash with `-e` ([docs](https://circleci.com/docs/configuration-reference/#default-shell-options)). The other uses of `grep` above don't use `exit`.

lognaturel (approved):
Thank you! I left it as-is when I did the last update pass here thinking that there were unlikely to be changes that would affect host functionality between releases of 20.04. I do think we should generally track what we use in Cloud which is currently 20.04 latest. That means this is indeed an improvement and we can look into the pm2 issue when we plan on doing an upgrade to hosts.

PR #676, opened 2024-06-24
Add workflow to publish nginx and service images to GHCR

### LINKED ISSUE #165: Build and host Docker containers

lognaturel:
Closes #165
Progress towards https://github.com/getodk/central/issues/656

Based on #249, https://github.com/getodk/central/pull/546

TODO:
- [x] Support custom SSL
- [x] Template client config at run time
- [x] Verify nginx image with Let's Encrypt certs
- [x] Verify client config with OIDC on (requires using https://github.com/getodk/central-frontend/pull/988)
- [x] Support [arm64](https://github.com/getodk/central/pull/676#issuecomment-2187913532)
- [x] Publish on master tag and manual trigger

The built images can be seen at https://github.com/lognaturel/central/pkgs/container/central-nginx and https://github.com/lognaturel/central/pkgs/container/central-service

I have verified that the service image + nginx image with custom certs work with the standard Docker Compose file.

I mostly followed the groundwork laid by @mattelacchiato in #249 and then @tobiasmcnulty at #546 with some further simplifications.

I used the image naming suggested by @tobiasmcnulty: `central-service` and `central-nginx`. I considered calling them frontend and backend but I think matching the Dockerfile names is the least confusing option.

It doesn't affect users yet because the images aren't used. I'm proposing splitting that off at https://github.com/getodk/central/issues/677

Not yet but https://github.com/getodk/central/issues/677 will.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -77,8 +77,6 @@ services:
   nginx:
     build:
       context: .
-      args:
-        - OIDC_ENABLED=${OIDC_ENABLED:-false}
       dockerfile: nginx.dockerfile
     depends_on:
       - service
@@ -90,6 +88,13 @@ services:
       - SENTRY_ORG_SUBDOMAIN=${SENTRY_ORG_SUBDOMAIN:-o130137}
       - SENTRY_KEY=${SENTRY_KEY:-3cf75f54983e473da6bd07daddf0d2ee}
       - SENTRY_PROJECT=${SENTRY_PROJECT:-1298632}
+      - OIDC_ENABLED=${OIDC_ENABLED:-false}
+    volumes:
+      - ./files/local/customssl/:/etc/customssl/live/local/
+      - ./files/nginx/redirector.conf:/etc/nginx/conf.d/redirector.conf
+      - ./files/nginx/common-headers.conf:/usr/share/odk/nginx/common-headers.conf
+      - ./files/nginx/odk.conf.template:/usr/share/odk/nginx/odk.conf.template
+      - ./files/nginx/client-config.json.template:/usr/share/odk/nginx/client-config.json.template
     ports:
       - "${HTTP_PORT:-80}:80"
       - "${HTTPS_PORT:-443}:443"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -90,11 +90,11 @@ services:
       - SENTRY_PROJECT=${SENTRY_PROJECT:-1298632}
       - OIDC_ENABLED=${OIDC_ENABLED:-false}
     volumes:
-      - ./files/local/customssl/:/etc/customssl/live/local/
-      - ./files/nginx/redirector.conf:/etc/nginx/conf.d/redirector.conf
-      - ./files/nginx/common-headers.conf:/usr/share/odk/nginx/common-headers.conf
-      - ./files/nginx/odk.conf.template:/usr/share/odk/nginx/odk.conf.template
-      - ./files/nginx/client-config.json.template:/usr/share/odk/nginx/client-config.json.template
+      - ./files/local/customssl/:/etc/customssl/live/local/:ro
+      - ./files/nginx/redirector.conf:/etc/nginx/conf.d/redirector.conf:ro
+      - ./files/nginx/common-headers.conf:/usr/share/odk/nginx/common-headers.conf:ro
+      - ./files/nginx/odk.conf.template:/usr/share/odk/nginx/odk.conf.template:ro
+      - ./files/nginx/client-config.json.template:/usr/share/odk/nginx/client-config.json.template:ro
     ports:
       - "${HTTP_PORT:-80}:80"
       - "${HTTPS_PORT:-443}:443"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -91,8 +91,6 @@ services:
       - OIDC_ENABLED=${OIDC_ENABLED:-false}
     volumes:
       - ./files/local/customssl/:/etc/customssl/live/local/:ro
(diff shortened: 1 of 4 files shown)


spwoodcock:
What do you think about building the images multi-arch?

Similar to pyxform-http:

```yaml
            # Setup QEMU emulator to build multi-arch images
            # https://github.com/docker/setup-qemu-action
            - name: Set up QEMU
              uses: docker/setup-qemu-action@v3

            # Configure Buildx for Docker build
            # https://github.com/docker/setup-buildx-action
            - name: Set up Buildx
              uses: docker/setup-buildx-action@v3

            # Build and push Docker image with Buildx
            # https://github.com/docker/build-push-action
            - name: Build and push Docker image
              uses: docker/build-push-action@v5
              with:
                  context: .
                  push: true
                  tags: ${{ steps.meta.outputs.tags }}
                  labels: ${{ steps.meta.outputs.labels }}
                  platforms: 'linux/amd64,linux/arm64'
```

The main things are the two additional actions above the build + the specified `platforms` during the build.

lognaturel:
Oh yes, I meant to add that to the todo! We can do that here. 👍

spwoodcock:
Although this does run on every push, so building multi-arch will add to the build time and might not make sense. 

Do you think it might be good to build on release instead, or perhaps in addition?

The docker/metadata action will handle correct image tagging for releases.

lognaturel:
> this does run on every push

That I do have in my todo list to address! I'm just leaving it like that until all the other issues are addressed. I can add multiarch at that same time since you've told me exactly how. 🎉

docker-compose.yml
```diff
@@ -90,6 +88,13 @@ services:
       - SENTRY_ORG_SUBDOMAIN=${SENTRY_ORG_SUBDOMAIN:-o130137}
       - SENTRY_KEY=${SENTRY_KEY:-3cf75f54983e473da6bd07daddf0d2ee}
       - SENTRY_PROJECT=${SENTRY_PROJECT:-1298632}
+      - OIDC_ENABLED=${OIDC_ENABLED:-false}
+    volumes:
+      - ./files/local/customssl/:/etc/customssl/live/local/
+      - ./files/nginx/redirector.conf:/etc/nginx/conf.d/redirector.conf
+      - ./files/nginx/common-headers.conf:/usr/share/odk/nginx/common-headers.conf
```
lognaturel, line 95: @yanokwa we do want users to be able to customize `redirector.conf` and `common-headers.conf`, right?

yanokwa: No customization on either.

files/nginx/setup-odk.sh
```diff
@@ -17,7 +27,6 @@ fi
 
 # start from fresh templates in case ssl type has changed
 echo "writing fresh nginx templates..."
-cp /usr/share/odk/nginx/redirector.conf /etc/nginx/conf.d/redirector.conf
```
yanokwa, line 20: You'll likely need to add this back. It was there because if you go from upstream to letsencrypt, the file at `/etc/nginx/conf.d/redirector.conf` gets overwritten and you need a fresh template.

files/nginx/setup-odk.sh
```diff
@@ -1,5 +1,15 @@
 #!/bin/bash
 
+
+echo "writing client config..."
+if [[ $OIDC_ENABLED != 'true' ]] && [[ $OIDC_ENABLED != 'false' ]]; then
```
yanokwa, line 5: This is more robust, but we don't do any other such checking elsewhere. I'd bias toward if true, then it's enabled. If it's anything else, it's not enabled.

yanokwa (changes requested):

docker-compose.yml
lognaturel: Got it. I've moved this back to a copy in the Dockerfile, then. That means those files will be copied at build time so will be static in the image.

files/nginx/setup-odk.sh
lognaturel: Ahh, yes. I've added a comment.

yanokwa (commented):
Before merging, verify that toggling between self-sign, upstream, and letsencrypt works

lognaturel:
- Started with customssl, confirmed working
- Switched to selfsign, confirm Firefox gives self-signed security error, confirm `/etc/selfsign/live/local/` has fullchain and privkey, confirmed let's encrypt redirect is gone from redirector.conf
- Switched back to customssl, confirmed working
- Switched to letsencrypt, confirmed working; `/etc/nginx/conf.d/redirector.conf` has challenge reply, Firefox shows Let's Encrypt cert
- Switched to upstream, `redirector.conf` is gone, I see `X-Forwarded-Proto https` in nginx conf
- Switched back to customssl, confirmed working

All of this is near instant: modify .env, `docker compose up -d`, that's it.

yanokwa (approved):

lognaturel:
Thanks for the review @spwoodcock and @yanokwa! Let’s merge now and then file issues as we see improvements to make.

PR #701, opened 2024-09-03
Support S3 blob storage

alxndrsn:
Companion to https://github.com/getodk/central-backend/pull/1071


lognaturel (approved):
We briefly discussed whether to use a prefix more generic than S3 but came back to it for its simplicity.

I added the keys to the env template.

This has been running well on the dev server with both manual and scheduled blob updates.

PR #715, opened 2024-09-19
nginx: add tests for config

alxndrsn:


test/test/.gitignore
alxndrsn, line 1: The files in this directory could probably move up a level so there is not `test/test`

test/nginx-test.dockerfile
```diff
@@ -0,0 +1,28 @@
+#
+# KEEP THIS FILE AS CLOSE AS POSSIBLE TO ../nginx.dockerfile
+#
+# It is ESPECIALLY IMPORTANT to keep IDENTICAL BASE IMAGES.
+#
```
alxndrsn, line 5: One alternative to this copy/paste could be `INCLUDE+` (https://github.com/edrevo/dockerfile-plus)

lognaturel: Let's address this as part of https://github.com/getodk/central/issues/677 Then I think we can use our actual published nginx image as the base image.

lognaturel (approved):
I agree that having some dedicated testing of the nginx config is a great idea. I'd like to talk through whether we really need a separate dockerfile for nginx but we can revise later.

PR #716, opened 2024-09-19
s3: handle 307 redirects

alxndrsn:
Depends on https://github.com/getodk/central/pull/715

### DIFF
--- a/.github/workflows/test-nginx.yml
+++ b/.github/workflows/test-nginx.yml
@@ -22,3 +22,5 @@ jobs:
       run: docker logs test-service-1
     - if: always()
       run: docker logs test-enketo-1
+    - if: always()
+      run: docker logs test-mock_s3-1
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -58,6 +58,22 @@ server {
     proxy_request_buffering on;
     proxy_buffering off;
     proxy_read_timeout 2m;
+
+    # transparently follow 307 redirects
+    # See: https://serverfault.com/a/792035
+    proxy_intercept_errors on;
+    error_page 307 = @follow_redirect_transparently;
+  }
+
+  location @follow_redirect_transparently {
+    # TODO 127.0.0.11 is docker DNS, and may not be required in production
+    # TODO 8.8.8.8 is google DNS, and may not be palatable to some users
+    # TODO resolver config will need testing IRL
+    # See: https://nginx.org/en/docs/http/ngx_http_core_module.html#resolver
+    # See: https://stackoverflow.com/a/40331256
+    resolver 127.0.0.11 8.8.8.8;
+    set $hdr_location '$upstream_http_location';
+    proxy_pass '$hdr_location';
   }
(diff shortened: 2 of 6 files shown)


files/nginx/odk.conf.template
```diff
@@ -58,6 +68,22 @@ server {
     proxy_request_buffering on;
     proxy_buffering off;
     proxy_read_timeout 2m;
+
+    # transparently follow 307 redirects
+    # See: https://serverfault.com/a/792035
+    proxy_intercept_errors on;
+    error_page 307 = @transparent_proxy_no_auth;
```
alxndrsn, line 75: this is triplicated; remove 2 copies!

files/nginx/odk.conf.template
```diff
@@ -58,6 +68,22 @@ server {
...
+  }
+
+  location @transparent_proxy_no_auth {
```
alxndrsn, line 78: Is `_no_auth` suffix relevant here?

service.dockerfile
```diff
@@ -17,7 +17,7 @@ RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODEN
 
 
 
-FROM node:${node_version}-slim as intermediate
+FROM node:${node_version}-slim AS intermediate
```
alxndrsn, line 20: already merged to `next`!

.github/workflows/test-nginx.yml
```diff
@@ -15,3 +15,12 @@ jobs:
...
     - run: cd test && ./run-tests.sh
+
+    - if: always()
+      run: docker logs test-nginx-1
+    - if: always()
+      run: docker logs test-service-1 
+    - if: always()
+      run: docker logs test-enketo-1
+    - if: always()
+      run: docker logs test-mock_s3-1
```
alxndrsn, line 26: Move to #715

test/mock-http-server/index.js
```diff
@@ -29,7 +38,7 @@ function withStdLogging(fn) {
   };
 }
 
-function logRequestType(method) {
+function ok(method) {
```
alxndrsn, line 41: Move to #715

lognaturel:
Capturing some quick initial thoughts that I will chew on for a bit:
- the behavior is nice. It doesn't fully hide the storage used because some headers may still come through from the provider but it does keep things like the bucket name private (though that's not actually private info)
- there's non-trivial overhead (see below)
- it would be nice for the DNS server to be configurable
- maybe the whole behavior could/should be opt-in


I did a few rounds of looking at request time and got consistent results that look like about 200ms extra than without which is itself already more than when the blobs were in the db (infrastructure-dependent, etc, etc):

Without:
Screenshot 2024-09-20 at 12 51 21 PM

With:
Screenshot 2024-09-20 at 12 55 25 PM

From db:
Screenshot 2024-09-20 at 1 02 23 PM

alxndrsn:
Closing after discussion with @lognaturel.

PR #724, opened 2024-09-26
Remove write-client-config.sh

matthew-white:
I don't see anything running write-client-config.sh, so I think it's safe to remove that file. It looks like #676 added the code in write-client-config.sh to another file, so doesn't run the script anymore.


lognaturel (approved):
Good catch!

PR #361, opened 2022-12-01
Dockerignore: convert to whitelist

alxndrsn:

PR #738, opened 2024-10-05
Dockerignore: convert to whitelist

alxndrsn:
Update to https://github.com/getodk/central/pull/361

* conflicts resolved
* convert .dockerignore from blacklist-based to whitelist
* decrease docker build context size ~significantly~ (most savings would come from excluding .git dirs)
* make docker builds less dependent on local state (e.g. local node_modules in submodules)
* add script for checking docker build context
* add CI tests to monitor if surprisingly large changes are made to the build context

To check build context locally, run:

    ./test/check-docker-context.sh --report
    
To check against previous `.dockerignore`:

    git checkout next -- .dockerignore && ./test/check-docker-context.sh --report


alxndrsn:
Unfortunately generating `version.txt` requires all `.git` directories to be included in the build context.

alxndrsn:
Currently, `.git` in published `central-service` images is a file:

```sh
/usr/odk# cat .git
gitdir: ../.git/modules/server
```

lognaturel:
Could we maybe delete it after the build step in the docker file?

alxndrsn:
> Could we maybe delete it after the build step in the docker file?

That could be done, but the file itself looks harmless.

.dockerignore
```diff
@@ -1,2 +1,33 @@
...
+!/client/icomoon.json
+!/client/jsconfig.json
+!/client/package.json
+!/client/package-lock.json
+!/client/vue.config.js
+!/client/bin/
+!/client/docs/
+!/client/public/
+!/client/src/
+!/client/transifex/
```
matthew-white, line 23: Not all of these are needed to build Frontend. I could list the ones that are, but that list will also change once we move from Vue CLI to Vite (getodk/central-frontend#671). One option for now is that we could use a blacklist just for `client`.

alxndrsn: Which ones can be skipped?

alxndrsn: I've removed icomoon etc. from the list above :+1:

alxndrsn:
Current stats:

## `next`

```
[check-docker-context.sh] File count:  3712
[check-docker-context.sh] Total size: [check-docker-context.sh] 85 MB
```

## this branch 

```
[check-docker-context.sh] File count:  3619
[check-docker-context.sh] Total size: [check-docker-context.sh] 67 MB
```

## without .git directories

```
[check-docker-context.sh] File count:  673
[check-docker-context.sh] Total size: [check-docker-context.sh] 7 MB
```
```

**Comment by matthew-white** [comment]
@alxndrsn, it looks like you moved test/check-docker-context.sh into its own PR (#894). Does that sound right? I'm in favor of that idea. I don't totally follow the Docker things happening in that script, so it'd probably be better for someone else to review that PR.

What do you think about removing the script from this PR? I should be able to review the .dockerignore file at least.

**Comment by alxndrsn** [comment]
> @alxndrsn, it looks like you moved test/check-docker-context.sh into its own PR (#894). Does that sound right?

:+1: Currently integral to this PR is checking that the docker context size has not changed significantly.  I'll revisit that idea when #894 is merged/rejected, and see how this PR can be simplified.

File: .dockerignore [hunk 2]
```diff
@@ -1,2 +1,29 @@
-node_modules
-npm-debug.log
+**
+
+# .git directories required for generating version.txt
+
+!/.git/
+!/docs/
```
matthew-white: I think docs/ can be removed. Those are served separately via github.io.

.dockerignore
```diff
@@ -1,2 +1,29 @@
...
+!/files/
+!/test/files/
+
+!/client/.browserslistrc
+!/client/.git/
+!/client/index.html
+!/client/jsconfig.json
```
matthew-white: I think jsconfig.json can be removed. I'm pretty sure it's just for local development / the IDE.

.dockerignore
```diff
@@ -1,2 +1,29 @@
...
+!/client/package.json
+!/client/package-lock.json
+!/client/vite.config.js
+!/client/vue.config.js
```
matthew-white: Given that we're excluding client/test/, I think we should exclude vue.config.js as well. It's only used in testing.

.dockerignore
```diff
@@ -1,2 +1,29 @@
...
+!/client/public/
+!/client/src/
+
+!/server/.npmrc
+!/server/.git/
+!/server/package.json
+!/server/package-lock.json
+!/server/Makefile
```
matthew-white: Do we actually use the Makefile in Docker? Not sure, but I feel like it's just used for local development and `central-backend` CI.

.dockerignore
```diff
@@ -1,2 +1,29 @@
...
+!/server/pm2.config.js
+!/server/config/
+!/server/docs/
```
matthew-white, line 23: I think server/docs/ can be excluded from Docker. It's API docs and developer docs.

alxndrsn: Could API docs be helpful?  Especially as they should match the `odk-central-backend` version they're bundled with...

matthew-white: That sounds reasonable to me. /server/docs contains files other than API docs, but it's probably OK to bundle those as well. Or we could just whitelist /server/docs/api.yaml.

.dockerignore
```diff
@@ -1,2 +1,25 @@
-node_modules
-npm-debug.log
+**
+
+# .git directories required for generating version.txt
+
+!/.git/
+!/files/
+!/test/files/
```
matthew-white: Could you say a little bit about this line? I see that #1652 ignores the entire test/ directory, but do we want to include test/files/?

.dockerignore
```diff
@@ -1,2 +1,25 @@
-node_modules
-npm-debug.log
+**
+
+# .git directories required for generating version.txt
+
+!/.git/
```
matthew-white, line 5: I feel like we should include .gitignore files as well. Those could conceivably affect what is logged to version.txt (including whether the Git working directory is clean).

.dockerignore
matthew-white, line 1: Do we need to include `.dockerfile`s or `docker-compose.yml`? How about `.env`?

.dockerignore
```diff
@@ -1,2 +1,25 @@
...
+!/test/files/
+
+!/client/.browserslistrc
+!/client/.git/
+!/client/index.html
+!/client/package.json
+!/client/package-lock.json
+!/client/vite.config.js
+!/client/public/
+!/client/src/
```
matthew-white, line 15: This doesn't include /client/dist/, but is that because those files are built after the ignore step?

.dockerignore
matthew-white: Actually, now that I say that, if we're no longer copying all files that are committed to the repo, won't version.txt indicate that the working directory isn't clean?

matthew-white (commented):
This is looking very close to me!

In the previous round of review, I was mostly thinking about files that were listed that needn't be listed. This time, I did a scan through the three Central repos for files that weren't listed here but maybe need to be.

.dockerignore
alxndrsn: Good catch, thanks.  This line dates from 2024, so may be out of date.

.dockerignore
alxndrsn: Great question.  I'll investigate.  Given the importance of `version.txt` it would be good to have a test that it's created as expected.

.dockerignore
alxndrsn: It's intentional - the client `/dist` directory should be created inside the build container rather than passed in via the docker context.

With the current configuration, anything in the host machine's `client/dist` directory can end up in the generated image.

.dockerignore
alxndrsn: If a file is required inside any of the containers, it needs to be included in the docker context.

* `.dockerfile`s: these are used by docker to build containers; not used inside the containers
* `docker-compose.yml`: this is used by docker to orchestrate containers; not used inside the containers
* `.env`: looks like it's used for configuring `docker compose`; is it used inside any of the containers?

.dockerignore
alxndrsn: I don't think `version.txt` is affected by changes to the git working tree.

PR at https://github.com/getodk/central/pull/1660

.dockerignore
matthew-white: > is it used inside any of the containers?

I don't think so, so it sounds like we're all good here. :+1:

test/test-images.sh
matthew-white, line 1: Maybe a merge or rebase is in order? I wouldn't expect to see this file in the diff.

matthew-white (approved):

test/test-images.sh
alxndrsn: You should!  This PR significantly decreases the number of files in the docker context 🙂

.dockerignore
alxndrsn: Which raises the question: _should_ local changes be reflected in `version.txt`?

test/test-images.sh
matthew-white: Very exciting to see those numbers go down!

It looks like test-images.sh is no longer in the diff since the latest merge, so I think this comment can be considered resolved.

.dockerignore
matthew-white: > I don't think `version.txt` is affected by changes to the git working tree.
> 
> PR at #1660

I know I was the one to approve that PR, but I thought of it more as a general test of `version.txt` rather than a test of changes to the git working tree. Does test-images.sh make changes to the working tree before generating `version.txt`? Or is that what #1662 will do?

> Which raises the question: _should_ local changes be reflected in `version.txt`?

I'm not 100% sure that local changes are reflected in `version.txt`. I think they used to be, but I'm not sure what the current status is. It'd be kind of nice if they were, just because it gives us more information on the build. E.g., if a Sentry report comes in for a version marked `-dirty`, it could be that the error was due to a user changing the code. I'm not sure it's important enough to block this PR though.

matthew-white (changes requested):
Now that Frontend has become a monorepo, I think we'll probably need to make changes to this PR, in particular to the `client` submodule. How we do so may depend on how we end up building Web Forms as part of deploying Central / building the `nginx` image.

PR #750, opened 2024-10-09
Update pyxform to fix geojson support

lognaturel:
Closes #

Made docker-compose change on dev, verified I can convert a form with geojson attachment.

This fix comes in Validate where we've added the jackson dependency back in. Ideally we would not need it but for now this fix brings back broken functionality.

Should just fix geojson support. That's all that's changed across the stack.

No.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -108,7 +108,7 @@ services:
       options:
         max-file: "30"
   pyxform:
-    image: 'ghcr.io/getodk/pyxform-http:v2.1.0'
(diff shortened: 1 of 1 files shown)


matthew-white:
Is it OK to ship everything on the `next` branch with the patch release? I think so, but just wanted to check. If so, I'll retitle the PR for the `next` branch so that it's clear that it's for the patch.

lognaturel:
Yes and thank you!

sadiqkhoja (approved):
Validated locally, Forms with external geojson files can be uploaded.

Issue #770, opened 2024-11-09
Eliminate false CSP violatoins


alxndrsn:
Originally introduced in https://github.com/getodk/central-frontend/pull/631, there seem to be some false-positives.

Reported violations include:

* `Blocked style attribute from 'inline:'`
* https://github.com/getodk/central/issues/629
* https://github.com/getodk/central-backend/issues/1221
* `Blocked 'connect' from 'maps.googleapis.com'`

matthew-white:
> Originally introduced in https://github.com/getodk/central-frontend/pull/631

I think that PR was adding it to the nginx config in development. For production, it looks like it was added in #313.

alxndrsn:
> Some reported violations may be the result of google chrome's built-in translator, or google translate.

There are various positive reports which may be worth ignoring, either in the CSP, or in Sentry:

* various relating to gc.kis.v2.scr.kaspersky-labs.com, javascript.browser.wasscan.tenable - presumably these are some kind of security product injecting scripts into pages
* translate.google.com, translate.googleapis.com - maybe Chrome page translation
* custom fonts within central-frontend - caused by custom browser settings?

matthew-white:
Given the current timeline for v2025.3, I'm going to bump this issue to .4.

matthew-white:
We could consider flipping our current approach:

- Start with a looser protection that is fully enforced (not just reporting). Then strengthen it over time.
- Instead of: Stronger protection that is reporting only that we continually need to loosten.

PR #771, opened 2024-11-09
enketo: relax Content-Security-Policy for inline style attributes

alxndrsn:
Related https://github.com/getodk/central/issues/770


matthew-white:
@alxndrsn, just to make sure, this PR is ready for review, right?

matthew-white:
Would it make sense for @lognaturel to take a look? I think she's more familiar with our CSP than I am.

lognaturel (approved):

PR #814, opened 2024-12-03
nginx: separate cert paths from server_name

alxndrsn:
Working on #809 I noticed that the location of SSL certs is based either on the domain name, or on the method of supply of SSL certs.

Cert provision approach should probably not affect the nginx "server_name" setting.

Also, the old variable name `CNAME` (short for "certificate name?") is easily confused with the DNS concept of CNAME records ("canonical names") (https://en.wikipedia.org/wiki/CNAME_record).

Ran tests.

It may not be - there may be a subtle reason for the current use and/or naming of `CNAME`.

Behaviour that may have accidentally been affected:

Servers with `SSL_TYPE` = `"customssl"` may have their nginx `server_name` changed.  This is probably a positive change - currently it has no effect, but will do once #809 is merged.

It should not.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -1,10 +1,10 @@
 server {
   listen 443 ssl;
-  server_name ${CNAME};
+  server_name ${DOMAIN};
 
-  ssl_certificate /etc/${SSL_TYPE}/live/${CNAME}/fullchain.pem;
-  ssl_certificate_key /etc/${SSL_TYPE}/live/${CNAME}/privkey.pem;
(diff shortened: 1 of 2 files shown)


alxndrsn:
`CNAME` was first used for `server_name` in baa828bd2a654c6667c9649b79f4b4ee18358790 (cc @yanokwa)

yanokwa (approved):
I don't know why I called it CNAME. The change looks good to me.

alxndrsn:
> I don't know why I called it CNAME. The change looks good to me.

You didn't introduce the original `CNAME` variable - that came from @issa-tseng in 9b972e772e90a77057818975af436d8f06007f1a.

PR #840, opened 2024-12-10
nginx/setup-odk.sh: make file executable

alxndrsn:
This change brings consistency: all other executables in `files/**` are marked executable in git rather than `chmod`'d in `.dockerfile` files.

The `chmod` call was originally introduced in https://github.com/getodk/central/pull/676 without discussion.

There is a wider debate whether git can be trusted to manage file permissions, touched on at https://github.com/getodk/central/pull/830#discussion_r1875703910


alxndrsn:
> I don't see any downsides to being explicit about intent and mechanism and desired outcome by putting a couple of `install` or `chmod`+`chown` calls in there.

I think the obvious downside is that it's more code to read, and could be considered noise if you weren't aware/concerned about environments with `core.fileMode` or less common `umask` values set.  In addition to this, it's more code _which is unlikely to be required_ most of the time, so unless there are explicit tests added for it, missing `chmod` calls are likely to be introduced in future, or existing ones likely to be removed when their intention can't be easily understood.

So I suggest merging this PR now to standardise executable flag handling across the repo.  This could be followed up with another PR to add a test to CI which fails without `chmod +x` calls, and then add the required `chmod` calls.

lognaturel (approved):

Issue #774, opened 2023-02-08
Show message if version of Central is out-of-date


matthew-white:
Frontend fetches /version.txt. We could have Frontend show a message recommending an upgrade if the version in version.txt is out-of-date. The message would only be shown to sitewide administrators. (A sitewide administrator might not be able to complete the upgrade themselves, but they probably know who to contact to do so.)

Now that we're using date-based versioning, we could consider checking the year part of the version. The easiest case is if the year part is <= current year - 2, because then the version of Central is definitely quite out-of-date. If the year part is current year - 1, it's hard to know exactly how old the version of Central is without fetching the current version or a list of versions from an external source. If the year part is current year - 1, then the version of Central could be the latest version, or on the other extreme, it could be almost two years out-of-date.

PR #844, opened 2024-12-10
add outdated-version.html

sadiqkhoja:
Related to #774 

### DIFF
--- a/docs/outdated-version.html
+++ b/docs/outdated-version.html
@@ -1,0 +1,108 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <title>ODK Central</title>
+  <link rel="stylesheet" href="style.css">
+</head>
+
+<body>
+  <!-- this page is rendered in an iframe of size 600x250 -->
+  <div class="outdated-version" style="display: none;">
+    <h1 class="heading"></h1>
+    <p>
+      <span class="sentence1"></span>
+      <span class="sentence2"></span>
+    </p>
+    <script defer src='https://static.cloudflareinsights.com/beacon.min.js'
+      data-cf-beacon='{"token": "81f288331d6e4638be205e0e63388165"}'></script>
+  </div>
+
+  <script>
+    const i18n = {
+      "en": {
+        "heading": "You’re using a significantly outdated version of ODK Central",
+        "sentence1": "Upgrade now to protect your data and take advantage of the latest features.",
+        "sentence2": "If you don't want to maintain Central, try {OdkCloudLink}."
+      },
+      "fr": {
+        "heading": "Vous utilisez une version de ODK Central considérablement obsolète",
+        "sentence1": "Mettez à jour maintenant pour protéger vos données et profiter des dernières fonctionnalités.",
+        "sentence2": "Si vous ne voulez pas maintenir Central, essayez {OdkCloudLink}."
+      },
+      "es": {
+        "heading": "Estás usando una versión significativamente desactualizada de ODK Central",
+        "sentence1": "Actualiza ahora para proteger tus datos y aprovechar las últimas funciones.",
+        "sentence2": "Si no quieres mantener Central, prueba {OdkCloudLink}."
+      },
+      "cs": {
+        "heading": "Používáte výrazně zastaralou verzi ODK Central",
+        "sentence1": "Aktualizujte nyní, abyste ochránili svá data a využili nejnovější funkce.",
+        "sentence2": "Pokud nechcete spravovat Central, vyzkoušejte {OdkCloudLink}."
+      },
+      "de": {
+        "heading": "Sie verwenden eine deutlich veraltete Version von ODK Central",
+        "sentence1": "Aktualisieren Sie jetzt, um Ihre Daten zu schützen und die neuesten Funktionen zu nutzen.",
+        "sentence2": "Wenn Sie Central nicht warten möchten, probieren Sie {OdkCloudLink} aus."
+      },
+      "id": {
+        "heading": "Anda menggunakan versi ODK Central yang sangat usang",
+        "sentence1": "Perbarui sekarang untuk melindungi data Anda dan memanfaatkan fitur terbaru.",
+        "sentence2": "Jika Anda tidak ingin memelihara Central, coba {OdkCloudLink}."
+      },
+      "it": {
+        "heading": "Stai usando una versione di ODK Central significativamente obsoleta",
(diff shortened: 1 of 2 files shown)


docs/outdated-version.html
```diff
@@ -0,0 +1,38 @@
...
+  </head>
+  <body>
+    <!-- this page is rendered in an iframe of size 600x250 -->
+    <div class="outdated-version" style="display: none;">
+      <h1 class="heading">You're using a significantly outdated version of ODK Central</h1>
+      <p>
+        <span class="sentence1">Upgrade now to protect your data and take advantage of the latest features.</span>
+        <span class="sentence2">If you don't want to maintain Central, try<a href="https://getodk.org/#success-stories" target="_blank"> ODK Cloud.</a></span>
+      </p>
+      <script defer src='https://static.cloudflareinsights.com/beacon.min.js' data-cf-beacon='{"token": "81f288331d6e4638be205e0e63388165"}'></script>
```
matthew-white, line 16: Do we need this here? I feel like we just need it in news.html.

docs/i18n.json
matthew-white, line 1: One thing we could do is create a component in Frontend with the source text, but never use the component in Frontend. The source text would still get sent to Transifex for translation though. We could copy the translations here whenever they change.

If we took that approach, the structure of the JSON would be [language] > [message key], not [message key] > [language].

docs/outdated-version.html
```diff
@@ -0,0 +1,38 @@
...
+      document.addEventListener('DOMContentLoaded', () => {
+        const queryParams = new URLSearchParams(location.search);
+        const lang = queryParams.get('lang');
+        const odkCloudLink = '<a href="https://getodk.org/#success-stories" target="_blank">ODK Cloud</a>';
+        
+        fetch('./i18n.json').then(r => r.json()).then(i18n => {
+          const localText = Object.keys(i18n).reduce((a, k) => {a.set(k, i18n[k][lang] ?? i18n[k]['en']); return a;}, new Map());
+          document.querySelector('.heading').textContent = localText.get('heading');
+          document.querySelector('.sentence1').textContent = localText.get('sentence1');
+          document.querySelector('.sentence2').innerHTML = localText.get('sentence2').replace('{{OdkCloudLink}}', odkCloudLink);
```
matthew-white, line 29: We don't need `{{OdkCloudLink}}` anymore, right? I feel a little unsure about setting `innerHTML` like this. I'd think about a translator adding malicious HTML to their translation of `sentence2`.

docs/outdated-version.html
```diff
@@ -0,0 +1,38 @@
...
+      <script defer src='https://static.cloudflareinsights.com/beacon.min.js' data-cf-beacon='{"token": "81f288331d6e4638be205e0e63388165"}'></script>
+    </div>
+
+    <script>
+      document.addEventListener('DOMContentLoaded', () => {
+        const queryParams = new URLSearchParams(location.search);
+        const lang = queryParams.get('lang');
+        const odkCloudLink = '<a href="https://getodk.org/#success-stories" target="_blank">ODK Cloud</a>';
+        
+        fetch('./i18n.json').then(r => r.json()).then(i18n => {
```
matthew-white, line 25: Rather than fetching it, I wouldn't have a problem with bundling all the translations in this file. There aren't many messages to translate, so I don't think it'd be too large. Either way (fetching or not) seems good to me.

docs/outdated-version.html
```diff
@@ -0,0 +1,38 @@
...
+          const localText = Object.keys(i18n).reduce((a, k) => {a.set(k, i18n[k][lang] ?? i18n[k]['en']); return a;}, new Map());
+          document.querySelector('.heading').textContent = localText.get('heading');
+          document.querySelector('.sentence1').textContent = localText.get('sentence1');
+          document.querySelector('.sentence2').innerHTML = localText.get('sentence2').replace('{{OdkCloudLink}}', odkCloudLink);
+
+          document.querySelector('.outdated-version').style = 'display: flex;'
+        })
+      })
```
matthew-white, line 33:
```suggestion
      });
```

docs/outdated-version.html
```diff
@@ -0,0 +1,38 @@
...
+        const lang = queryParams.get('lang');
+        const odkCloudLink = '<a href="https://getodk.org/#success-stories" target="_blank">ODK Cloud</a>';
+        
+        fetch('./i18n.json').then(r => r.json()).then(i18n => {
+          const localText = Object.keys(i18n).reduce((a, k) => {a.set(k, i18n[k][lang] ?? i18n[k]['en']); return a;}, new Map());
+          document.querySelector('.heading').textContent = localText.get('heading');
+          document.querySelector('.sentence1').textContent = localText.get('sentence1');
+          document.querySelector('.sentence2').innerHTML = localText.get('sentence2').replace('{{OdkCloudLink}}', odkCloudLink);
+
+          document.querySelector('.outdated-version').style = 'display: flex;'
```
matthew-white, line 31: I was confused at first as to why this was necessary. I thought it was just duplicating style.css. But it's needed because the element is initially hidden via `display: none`. How about adding a comment along those lines?

matthew-white (commented):
This approach looks good to me!

docs/outdated-version.html
sadiqkhoja: I think this will help us in version analytics

matthew-white: Ah I see. Will the analytics show the `version` query parameter?

docs/outdated-version.html
sadiqkhoja: I think we need it. What do you think of making it safer:

```
const parts = localText.get('sentence2').split('{{OdkCloudLink}}');
document.querySelector('.sentence2').textContent = '';
document.querySelector('.sentence2').appendChild(document.createTextNode(parts[0]));
document.querySelector('.sentence2').appendChild(odkCloudLink);
document.querySelector('.sentence2').appendChild(document.createTextNode(parts[1]));
```

docs/outdated-version.html
sadiqkhoja: I think cloudflareinsights logs full URL, let's confirm it with Yaw

matthew-white:
I changed the base branch to `next`, but maybe `master` would be fine in this case actually. We sort of need it deployed right away so that it can be shown in Frontend.

docs/i18n.json
sadiqkhoja: I like that approach

matthew-white:
> We sort of need it deployed right away so that it can be shown in Frontend.

Just to be explicit, it will only be deployed right away if we merge the PR into the `master` branch, not `next`. The target branch is currently `next`, since I changed it.

PR #868, opened 2025-01-07
Add Node version update process

lognaturel:
Follow-up from v2024.3 retrospective. In that release, we upgraded node to 22 and I had some hesitation around when it should be done and by whom.

I considered explicitly saying that I do the upgrade similar to what [Collect does](https://github.com/getodk/collect/?tab=readme-ov-file#release-cycle) but I don't think it's necessary until we decide to document more of the release process here.

While here I changed the heading syntax to use `#` because it allows deeper nesting and I think it's easier to read.

Targets `master` since it's docs only.

Read it.

It provides the info succinctly.

No user-facing change, this just documents existing process.

This is the docs update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-ODK Central
-===========
+# ODK Central
 
 ![Platform](https://img.shields.io/badge/platform-Docker-blue.svg)
 [![License](https://img.shields.io/badge/license-Apache_2.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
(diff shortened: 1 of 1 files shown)


lognaturel:
I see a 👍 from @sadiqkhoja so I've changed the review request!

sadiqkhoja (approved):

lognaturel:
Dopamiiiiiine! Hope you got a bit too, @sadiqkhoja 😅🤘

PR #870, opened 2025-01-13
backend: allow configuring S3_REGION

### LINKED ISSUE #846: Add optional S3_REGION config
We originally decided not to make region configurable because minio detects, caches and uses the region when making a request against a generic storage URL. 

We've now discovered that for Amazon S3, regions introduced after 2019 like `af-south-1` can't use the generic URL: https://forum.getodk.org/t/using-external-aws-s3-bucket-in-region-other-than-us-east-1/51392

alxndrsn:
With https://github.com/getodk/central-backend/pull/1357 & https://github.com/getodk/docs/pull/1912, closes #846

e2e tests in central-backend.  Suggestions for higher-level tests appreciated!

It seems simple and in-line with existing s3 configuration.

Should not have a direct effect on users, although changes to central-backend may.

See linked PR for https://github.com/getodk/docs/


lognaturel:
We may want to add this feature at some point but for now the short-term need was addressed by a minio upgrade: https://github.com/getodk/central/issues/846#issuecomment-2610714967

PR #876, opened 2025-01-16
Update start-odk.sh memory calculation to support cgroups v2

### LINKED ISSUE #875: start-odk.sh memory calculation does not support cgroups v2
### Issue

- A made a [commit](https://github.com/hotosm/odkcentral/commit/7476a9110ed5d29279eed9e9f6a5b36f4249f646) a while ago to determine the total memory [in the container](https://github.com/getodk/central/issues/577) using cgroups v1.
- This fix works fine for containers running cgroups v1, but does not work for cgroups v2.
- The version used is primarily determined by the host OS.
- I made a bit of an oversight, as my machine uses cgroups v1, but v2 is increasingly common. More info for [Docker](https://docs.docker.com/engine/containers/runmetrics/#enumerate-cgroups) and [Kubernetes](https://kubernetes.io/docs/concepts/architecture/cgroups/).
- In scenarios where cgroups v2 is used, we get the following error (although the server still starts correctly):
    Image

### Solution

- The script should handle both scenarios.
- It's easy enough to determine the cgroup version.

spwoodcock:
Closes #875 

#### Included in this PR

- The original check for cgroups v1.
- A check for cgroups v2.
  - If there are memory constraints set, then they are used.
  - If there are no memory constraints, the value will be set to `max` and we fallback to use the total memory available on the system.
- Note that it's possible to determine hard and soft memory limits via cgroups, but this is probably overkill.

- Tested on my local Debian Bookworm setup (using WSL, so it's cgroups v1).
- Tested on a Debian Bookworm remote server, using cgroups v2.

- It seems to be a pretty common way to determine memory available inside a container.
- Containers are essentially wrappers for standard tools such as cgroups, so the total memory available in the container is determined by a cgroup.
- It doesn't require any additional tools / only uses standard Linux stuff.
- This is backwards compatible with cgroups v1 and should work for the foreseeable future (unlikely cgroups will change behaviour for a long time now).

- Users that deployed since the commit https://github.com/hotosm/odkcentral/commit/7476a9110ed5d29279eed9e9f6a5b36f4249f646, and using cgroups v2 in their deployment infrastructure, may have an incorrectly determined number of workers!
- Likely they may be running with a single worker.
- This would increase the number of workers, depending on the memory available.

- Probably not.

### DIFF
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -73,7 +73,8 @@ determine_worker_count() {
 }
 
 MEMTOT=$(get_memory_limit)
-export WORKER_COUNT=$(determine_worker_count "$MEMTOT")
+WORKER_COUNT=$(determine_worker_count "$MEMTOT")
+export WORKER_COUNT
 echo "using $WORKER_COUNT worker(s) based on available memory ($MEMTOT).."
 
 echo "starting server."
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -30,13 +30,50 @@ fi
 echo "starting cron.."
 cron -f &
 
-MEMTOT=$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)
-if [ "$MEMTOT" -gt "1100000" ]
-then
-  export WORKER_COUNT=4
-else
-  export WORKER_COUNT=1
-fi
+get_cgroup_version() {
+  # The max memory calculation is different between cgroup v1 & v2
+  local cgroup_type
+  cgroup_type=$(stat -fc %T /sys/fs/cgroup/)
+  if [ "$cgroup_type" == "cgroup2fs" ]; then
+    echo "v2"
+  else
+    echo "v1"
+  fi
+}
+
+get_memory_limit() {
+  local cgroup_version
+  cgroup_version=$(get_cgroup_version)
+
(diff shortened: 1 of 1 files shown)


lognaturel (changes requested):
Thanks! This looks good to me but tests are failing in CircleCI. Is it possible that you branched off of a very old version? Could you please rebase on the latest `next` so we can get a CircleCI sanity check before merging?

spwoodcock:
> Thanks! This looks good to me but tests are failing in CircleCI. Is it possible that you branched off of a very old version? Could you please rebase on the latest `next` so we can get a CircleCI sanity check before merging?

Ah possibly! I'm afk for the weekend now, but will address this on Monday 😄

spwoodcock:
You were right, I had synced `master` but not `next` on our fork - rebased to the latest `next` branch!

Hopefully the CircleCI build can be re-triggered without much hassle 🙏

spwoodcock:
Made a small fix for shellcheck linting:

```
In files/service/scripts/start-odk.sh line 76:
export WORKER_COUNT=$(determine_worker_count "$MEMTOT")
       ^----------^ SC2155: Declare and assign separately to avoid masking return values.

For more information:
  https://www.shellcheck.net/wiki/SC2155 -- Declare and assign separately to ...
```

lognaturel (approved):

lognaturel:
Thank you!

PR #881, opened 2025-01-23
Release v2025.1

matthew-white:
This PR prepares the release of v2025.1. It should only contain changes from other PRs that have already been approved and merged (and possibly merge commits from the `master` branch).


matthew-white:
@alxndrsn, GitHub notes that CircleCI is failing here. We've gotten rid of CircleCI, right? Is there a config change that needs to happen in the GitHub settings for the repo?

alxndrsn:
> Is there a config change that needs to happen in the GitHub settings for the repo?

That sounds right, but I can't find where to (un)configure the check.  Maybe it needs to be deactivated through the CircleCI UI?

matthew-white:
Hmm I think you might be right. If it were configured within GitHub, I'd expect to see it under branch protection rules, but I'm not seeing CircleCI there. I am seeing an option within CircleCI to "Stop Building", maybe that's it. But I think we should probably hold off on that until we're closer to the release of v2025.1. If we were to put out a patch release before then, we'd still rely on CircleCI for that, so I don't think we should turn off CircleCI completely until we're sure that we're done with patch releases.

ktuite (approved):

Issue #342, opened 2022-10-24
Standardise shebangs and script invocations


alxndrsn:
Ref https://github.com/getodk/central/issues/330, https://github.com/getodk/central/pull/331#issue-1400724251

matthew-white:
Closed by #891.

PR #890, opened 2025-02-11
postgres-upgrade: halt if any stage of upgrade fails

alxndrsn:
Noted while working on https://github.com/getodk/central/pull/882: `files/postgres/upgrade-postgres.sh` is not actually run with `/bin/bash -eu` as its shebang declares, but with `/bin/sh`.

## This PR

1. fixes the shell used for `upgrade-postgres.sh`
2. introduces a test which would have detected this issue, and should prevent similar issues in future
3. fixes use of deprecated dockerfile syntax detected by the new test
4. pulls `shellcheck` invocation into a script file to allow easier use locally, and possible future extension

## Docker build checks

Currently in beta, "docker build checks" are a best-practice linter for dockerfiles.

See: https://docs.docker.com/build/checks/#check-a-build-without-building

## Dockerfile fixes

* `postgres-upgrade.dockerfile`: https://docs.docker.com/reference/build-checks/json-args-recommended/
* `postgres14.dockerfile`: https://docs.docker.com/go/dockerfile/rule/legacy-key-value-format/

-----

Related: https://github.com/getodk/central/issues/342

-----

CI

Docker build checks are built-in to docker.  However, they're in beta, so there is a chance they could be changed or dropped in future.  This PR could be restricted to simply changing `postgres-upgrade.dockerfile`.

Should not affect users.

No.


postgres14.dockerfile
```diff
@@ -11,7 +11,7 @@ COPY files/postgres14/start-postgres.sh /usr/local/bin/
 #
 # To ensure future accessibility, PGDATA must be stored _outside_ the
 # anonymous volume.
-ENV PGDATA /var/lib/odk/postgresql/14/data
+ENV PGDATA=/var/lib/odk/postgresql/14/data
 
 ENTRYPOINT []
-CMD ["start-postgres.sh"]
\ No newline at end of file
+CMD ["start-postgres.sh"]
```
alxndrsn, line 17: revert this change

.circleci/config.yml
```diff
@@ -10,6 +10,8 @@ jobs:
       - run: sudo apt install shellcheck
       - run: cat <(git grep -El '^#!.*sh\b') <(git ls-files | grep -E '.sh$') | sort -u | xargs shellcheck --exclude=SC2016
 
+      - run: ./test/check-dockerfiles.sh
```
lognaturel, line 13: This will need to move to github actions once other PRs are merged

lognaturel (approved):

.github/workflows/test.yml
```diff
@@ -5,13 +5,16 @@ on:
...
-  shellcheck:
+  lint:
     timeout-minutes: 2
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
     - run: sudo apt-get install shellcheck
-    - run: cat <(git grep -El '^#!.*sh\b') <(git ls-files | grep -E '.sh$') | sort -u | xargs shellcheck --exclude=SC2016
+    - name: shellcheck
+      run: cat <(git grep -El '^#!.*sh\b') <(git ls-files | grep -E '.sh$') | sort -u | xargs shellcheck --exclude=SC2016
```
alxndrsn, line 15: pull this into a script

PR #891, opened 2025-02-11
scripts: standardise shebangs

### LINKED ISSUE #341: Add shell flags to make scripts more predictable
Introduce some shell flags/options (e.g. `-e`, `-u`, `-o pipefail`) to make script behaviour more predictable.

Ref https://github.com/getodk/central/issues/330

### LINKED ISSUE #342: Standardise shebangs and script invocations

alxndrsn:
This came up as part of work on https://github.com/getodk/central-backend/pull/1363 when trying to understand the purpose of the `check-migrations` make target in `odk-central-backend`.

----

* use bash everywhere, instead of a mixture of `sh`, `bash`, and (implicitly) `dash`
* for consistency, always set flags/options:
  * `-e`: "Exit immediately if a pipeline...returns a non-zero status."
  * `-u`: "Treat unset variables...as an error"
  * `set -o pipefail`: "the return value of a pipeline is the value of the last (rightmost) command to exit with a non-zero status"
  * `shopt -s inherit_errexit`: "command substitution inherits the value of the errexit [`-e`] option"

See: https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#The-Set-Builtin
See: https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#The-Shopt-Builtin

In the service container's `start-odk.sh` script, this also removes the need for the explicit `make check-migrations` check, as all steps in scripts are now checked implicitly.

Closes #341
Closes #342

CI

\There are various things in this PR which could be broken down to simplify understanding and reduce the risk of breaking 3rd-party production environments:

* don't standardise on `-u` flag - this flag has a higher likelihood than others of having surprising effects
* don't bother checking for indirect invocations - there currently aren't any, and the check isn't perfect

`-u` flag might cause problems in custom envs.  The error messages generated should be quite obvious though, e.g.

```sh
$ bash -c 'set -u; echo $x'
bash: line 1: x: unbound variable
```

I don't think so.


ktuite (approved):
I think this should be fine? Just summarizing my own understanding of the PR: 

Having every file have the same header seems good:

``` bash
#!/bin/bash -eu
set -o pipefail
shopt -s inherit_errexit
```

The only other change in here was what we talked about with the `check-migrations` command, and how that was only introduced so that the server wouldn't start running if there were any problems with the migrations. 

I'm approving this but I want someone else with more ops insight like @lognaturel to be aware of this change, too.

files/service/odk-cmd
```diff
@@ -1,3 +1,5 @@
-#!/bin/sh
```
alxndrsn, line 1: Consider reverting to `sh`?

lognaturel (approved):
The original standard shebang was `#!/bin/sh`. The idea was to have maximal portability and specifically leave the door open to using `alpine` images which deliberately don't include bash for size reasons.

At some point, I think @alxndrsn started introducing `#!/bin/bash` shebangs, including for the frontend build script, and we didn't catch the inconsistency in review.

We could do `#!/bin/sh -eu` but the other error handling improvements are bash-specific.

Given we'd already accidentally been using some bash shebangs anyway and that we have no immediate plans to move to alpine images, I think we can go ahead with this.

alxndrsn:
> The original standard shebang was `#!/bin/sh`. The idea was to have maximal portability and specifically leave the door open to using `alpine` images which deliberately don't include bash for size reasons.
> 
> At some point, I think @alxndrsn started introducing `#!/bin/bash` shebangs, including for the frontend build script, and we didn't catch the inconsistency in review.

For the record, various bash shebangs and invocations have been used in this repo over time:

6c38322157e58110edf8f266b2d3759897789b76 (2018):

https://github.com/getodk/central/blob/6c38322157e58110edf8f266b2d3759897789b76/nginx.dockerfile#L14

https://github.com/getodk/central/blob/6c38322157e58110edf8f266b2d3759897789b76/files/nginx/odk-setup.sh#L25

19c87edb1309f2956fc512c9d245c7d6aeaf0d22 (2018):

https://github.com/getodk/central/blob/19c87edb1309f2956fc512c9d245c7d6aeaf0d22/files/service/scripts/start-odk.sh#L5

b6a6208adcc0defec5473c4f2dfa6517123c0f44 (2019):

https://github.com/getodk/central/blob/b6a6208adcc0defec5473c4f2dfa6517123c0f44/files/nginx/run_certbot.sh#L1

52248f848c5599d997ba8464dda1584b60223740 (2020):

https://github.com/getodk/central/blob/52248f848c5599d997ba8464dda1584b60223740/files/enketo/start-enketo.sh#L3

https://github.com/getodk/central/blob/52248f848c5599d997ba8464dda1584b60223740/files/service/scripts/start-odk.sh#L3

5a6b6cba608767256dcb469591182a9994afda1b (2022):

https://github.com/getodk/central/blob/5a6b6cba608767256dcb469591182a9994afda1b/files/nginx/odk-setup.sh#L1

lognaturel:
Record redressed! Sorry, I didn't mean to sound blaming, thanks for the fact check!

PR #894, opened 2025-02-13
ci/test-images: print docker context

alxndrsn:
This commit adds visibility into the docker context used for building container images.

This should make it easier to spot:

* unwanted files creeping into docker images
* context bloat

The script can also be run locally to understand differences between local and CI docker contexts.

Run locally & in CI.

Should not affect users.

No.


matthew-white:
I mentioned this at https://github.com/getodk/central/pull/738#issuecomment-2679950147, but I think I'm probably not the best person to review the script in this PR. I'm thinking that @sadiqkhoja or @lognaturel would be in a better position to take a look.

test/check-docker-context.sh
```diff
@@ -0,0 +1,112 @@
...
+    --skip-size)  skip_size=true ;;
+
+    --min-count)  shift;min_count="$1" ;;
+    --max-count)  shift;max_count="$1" ;;
+    --skip-count) skip_count=true ;;
+
+    *) log "!!! Unrecognised arg: $1"; exit 1 ;;
+  esac
+  shift
+done
```
sadiqkhoja, line 22: If I don't pass any argument then I get:

```
check-docker-context.sh: line 94: min_count: unbound variable
```

if I pass `--min-size 10000 --skip-count` only then I get:

```
check-docker-context.sh: line 103: max_size: unbound variable
```

test/check-docker-context.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash -eu
+set -o pipefail
+log() { echo "[$(basename "$0")] $*"; }
+
+# See: https://stackoverflow.com/a/71751097
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --report) skip_size=true; skip_count=true ;;
```
sadiqkhoja, line 9: can you add a `--help` with some basic usage - bare minimum

sadiqkhoja (commented):
This is super useful and clever 👏  and should be publish somewhere globally

test/check-docker-context.sh
alxndrsn: Added.

test/check-docker-context.sh
alxndrsn: Updated to show the `--help` message if no args are passed:

```sh
$ ./test/check-docker-context.sh
    USAGE
      ./test/check-docker-context.sh --report
      ./test/check-docker-context.sh [--min-size NUM --max-size NUM | --skip-size] [--min-count NUM --max-count NUM | --skip-count]
```

sadiqkhoja (approved):

PR #898, opened 2025-02-18
test/check-scripts: add checks to standardise whitespace

alxndrsn:
Incidental to #891.

Whitespace changes can be introduced inadvertently due to different editor configurations.  These changes then add noise to PRs, and if committed, add noise to future diffs and the git log.

This PR adds checks for the following in `.sh` files:

* trailing spaces on lines
* trailing newlines in files
* tabs vs spaces

Added new tests!

The status quo could be maintained.

No effect.

No.

### DIFF
--- a/test/check-scripts.sh
+++ b/test/check-scripts.sh
@@ -6,6 +6,37 @@ log() { echo >&2 "[$(basename "$0")] $*"; }
 
 scriptFiles="$(cat <(git grep -El '^#!.*sh\b') <(git ls-files | grep -E '.sh$') | sort -u)"
 
+for script in $scriptFiles; do
+  log "Checking $script ..."
+
+  log "  Checking trailing whiespace on lines..."
+  if grep -E '\s+$' "$script"; then
+    log "    !!! Whitespace found at end of line !!!"
+    exit 1
+  fi
+  log "    Passed OK."
+
+  log "  Checking trailing newline in files..."
+  if [[ -n "$(tail -c 1 < "$script")" ]]; then
+    log "    !!! Missing final newline !!!"
+    exit 1
+  fi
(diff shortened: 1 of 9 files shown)


test/check-scripts.sh
```diff
@@ -6,6 +6,37 @@ log() { echo >&2 "[$(basename "$0")] $*"; }
 
 scriptFiles="$(cat <(git grep -El '^#!.*sh\b') <(git ls-files | grep -E '.sh$') | sort -u)"
 
+for script in $scriptFiles; do
+  log "Checking $script ..."
+
+  log "  Checking trailing whiespace on lines..."
```
matthew-white, line 12:
```suggestion
  log "  Checking trailing whitespace on lines..."
```

test/check-scripts.sh
```diff
@@ -6,6 +6,37 @@ log() { echo >&2 "[$(basename "$0")] $*"; }
...
+    exit 1
+  fi
+  if [[ -z "$(tail -c 2 < "$script")" ]]; then
+    log "    !!! Blank lines at end of file !!!"
+    exit 1
+  fi
+  log "    Passed OK."
+
+  log "  Checking for tab-based indentation..."
+  if grep -P '\t' "$script"; then
```
matthew-white, line 31: `grep` doesn't have `-P` on my Mac. Would we ever run this script locally? If so, is there something other than `-P` that we could use?

matthew-white (approved):

test/check-scripts.sh
alxndrsn: I've updated to avoid `-P`; can you see if the new version works for you?

matthew-white: Awesome, yes, that seems to be working for me!

matthew-white (approved):

PR #905, opened 2025-02-26
Add Portuguese translations

lognaturel:
I inadvertently broke protocol a little bit here by cherry picking directly onto https://github.com/getodk/central-frontend/tree/v2024.3.x. I could do PRs targeting that branch if we prefer. We'll also want to tag v2024.3.2 on frontend and maybe we'll want to update here to include the tag, I'm not sure whether that matters currently.

Here's the frontend diff: https://github.com/getodk/central-frontend/compare/v2024.3.1...v2024.3.x?expand=1

There are also test failures on frontend that seem unrelated (maybe because of a change in infra versions?).

Ran locally, verified that Portuguese translations show.

I cherry-picked the original commit to add PT. Then I downloaded the latest from Transifex. I wanted to deep merge the two only keeping keys from the original file. I asked ChatGPT to help:

```
jq -s '
  def deep_replace(base; override):
    base as $b | override as $o |
    reduce ($b | keys[]) as $k ($b;
      if $k | in($o) then
        .[$k] = (
          if ( $b[$k] | type ) == "object" and ( $o[$k] | type ) == "object" then
            deep_replace($b[$k]; $o[$k])
          else
            $o[$k]
          end
        )
      else
        .
      end
    );

  deep_replace(.[0]; .[1])
'
```

Adds PT. Probably the biggest risk is that the merge didn't actually update everything from Transifex. That's not a big deal.

No.


matthew-white:
> There are also test failures on frontend that seem unrelated

Yes, I think that it's unrelated. Tests that used to pass randomly started failing in Chrome. We fixed it at getodk/central-frontend#1129 (discussion at getodk/central-frontend#1122). I don't think it's a problem to release a version of Frontend where that test is failing.

matthew-white:
> I inadvertently broke protocol a little bit here by cherry picking directly onto https://github.com/getodk/central-frontend/tree/v2024.3.x. I could do PRs targeting that branch if we prefer.

A branch sounds fine to me for this one!

matthew-white (approved):
Looks great! 🎉

I see that this PR updates the `client` submodule to the tip of the v2024.3.x branch in Frontend. I didn't review the cherry-picked commit on that branch, since it corresponds to a PR that was reviewed and approved (getodk/central-frontend#1130). I did review the other commit on the branch. I saw that it only updated Portuguese translations.

I can see in that commit that the structure of the Transifex/I18n JSON stayed the same: there were no messages that were entirely new, nor were any messages removed. Messages were only updated. I agree that the only risk would be a translation that wasn't updated, though we can see that many updates did go through.

Issue #906, opened 2025-02-26
Use web history routes instead of hash


sadiqkhoja:

matthew-white:
This is a change we've thought about doing on and off again, but it hasn't been prioritized: some effort seemed required, and it was unclear what the benefits would be. However, this change is now useful for something specific: #883. From getodk/central-frontend#1141:

> This change is prompted by the need to have cleaner and shorter public access link as part of web-forms integration.

Web Forms will be integrated within Frontend, yet we want Web Form links not to include the current Frontend hash.

PR #909, opened 2025-02-26
Features C#906: Enable web history for vue route, catch-all to index.html

sadiqkhoja:
Closes getodk/central#906

Manual verification

This is necessary to make webhistory routing mode work correctly in the frontend (Vue). Change is suggested at https://router.vuejs.org/guide/essentials/history-mode.html#nginx

None

No 

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -71,6 +71,7 @@ server {
 
   location / {
     root /usr/share/nginx/html;
+    try_files $uri $uri/ /index.html;
 
     location /version.txt {
       include /usr/share/odk/nginx/common-headers.conf;


files/nginx/odk.conf.template
```diff
@@ -71,6 +71,7 @@ server {
 
   location / {
     root /usr/share/nginx/html;
+    try_files $uri $uri/ /index.html;
```
matthew-white, line 74: For a moment, I was wondering why we wouldn't just always serve index.html. But then I realized that of course we have other assets to serve as well (CSS, other JS, etc.).

Another thing I was unsure about is why `$uri/` is needed in addition to `$uri`. [This discussion](https://serverfault.com/questions/1094254/what-is-the-difference-beween-uri-and-uri-in-the-try-files-directive) on StackOverflow was helpful to me. And like you mentioned, it's what the Vue docs recommend.

matthew-white (approved):

PR #935, opened 2025-03-26
service: add environment variable SENTRY_TRACE_RATE

### LINKED ISSUE #877: Configure Sentry for performance tracing
To identify/prioritize performance improvements, we can enable performance tracing in Sentry

Performance - e.g. for HTTP requests, see slowest requests (worst case, worse average, worst 95th-percentile etc.)

-   https://docs.sentry.io/product/performance/
-   https://docs.sentry.io/product/performance/transaction-summary/

Tracing:

- https://docs.sentry.io/concepts/key-terms/tracing/
- https://docs.sentry.io/product/tracing/
- https://docs.sentry.io/product/explore/traces/

alxndrsn:
Accompanies https://github.com/getodk/central-backend/pull/1447

Closes #877

:crossed_fingers: 

Could skip the `config.json` step, but that seems to be unidiomatic, and a backwards step with respect to https://github.com/getodk/central-backend/issues/1453

Could break if code chokes on missing/misconfigured `SENTRY_TRACE_RATE`?

This page could be updated: https://docs.getodk.org/central-install-digital-ocean/, but there are already other sentry config options which are not mentioned there.


.env.template
```diff
@@ -39,6 +39,7 @@ HTTPS_PORT=443
 # SENTRY_ORG_SUBDOMAIN=
 # SENTRY_KEY=
 # SENTRY_PROJECT=
+# SENTRY_TRACE_RATE=
```
sadiqkhoja, line 42: I think a comment here explaining what is it and it's acceptable value would be nice

sadiqkhoja (approved):

.env.template
alxndrsn: I can't see any precedent for that in https://github.com/getodk/central/blob/e4221ebeb41cd6ccb0cedad0461e5b603c207339/.env.template

alxndrsn: What would you suggest?

sadiqkhoja: `# The value should be a number between 0 and 1, representing the percentage chance that transactions will be sent to Sentry for performance tracing. Default is 0.1` would be nice but I let you decide if this would be helpful or not.

However information about this new variable should definitely be added here:
https://github.com/getodk/docs/blob/master/docs/central-install-digital-ocean.rst#disabling-or-customizing-sentry

alxndrsn: Added; docs PR at https://github.com/getodk/docs/pull/1933

sadiqkhoja (approved):

PR #965, opened 2025-04-18
Fixes/950 single submission link

### LINKED ISSUE #950: Enketo is enabled - Single Submission Public link doesn't work
Example: See "Single" public link at https://staging.getodk.cloud/projects/18/forms/ses2018/public-links

~Potential cause: `st` query parameter has `!`~ This can still be true, but we need to fix the following issue first.

Root cause: enketoOnceId is incorrect - [notes](https://docs.google.com/document/d/1AzOaax9xszZt32fJ1kh2LlGIT--KQO4PRECC3t30VEc/edit?tab=t.0#heading=h.mftcbgagpm3f)

sadiqkhoja:
Part of #950 

**Background**
For Single Submission links, Enketo’s frontend sets a cookie upon successful submission. The cookie has the format `enketoOnceId=timestamp_of_submission` and is scoped to the path /-/single.

However, since we now serve the Enketo Form from /enketo-passthrough, the cookie is no longer sent to the server, which prevents it from enforcing single submissions.

To fix this, we have added a dummy location /-/single/check-submitted in the Nginx configuration. The Central frontend loads this location in a hidden iframe and checks for the presence of the cookie with `<enketoOnceId>` key. If the cookie is found, the user is redirected to the thank-you page.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -65,6 +65,11 @@ server {
     # Form fill link (non-public), or Draft
     return 301 "/f/$enketoId/new$is_args$args";
   }
+  # To read single submission cookies
+  location = /-/single/check-submitted {
+      alias /usr/share/nginx/html/blank.html;
+      default_type text/html;
+  }
 
   # For that iframe to work, we'll need another path prefix (enketo-passthrough) under which we can
   # reach Enketo — this one will not be intercepted.
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -67,8 +67,8 @@ server {
   }
   # To read single submission cookies
   location = /-/single/check-submitted {
-      alias /usr/share/nginx/html/blank.html;
-      default_type text/html;
+    alias /usr/share/nginx/html/blank.html;
+    default_type text/html;
   }
 
   # For that iframe to work, we'll need another path prefix (enketo-passthrough) under which we can
--- a/test/nginx/test-nginx.js
+++ b/test/nginx/test-nginx.js
@@ -163,6 +163,7 @@ describe('nginx config', () => {
     // then
     assert.equal(res.status, 200);
     assert.isEmpty((await res.text()).trim());
+    await assertEnketoReceivedNoRequests();
   });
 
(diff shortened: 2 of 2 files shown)


test/nginx/test-nginx.js
```diff
@@ -156,6 +156,15 @@ describe('nginx config', () => {
     });
   });
 
+  it('should serve blank page on /-/single/check-submitted', async () => {
+    // when
+    const res = await fetchHttps('/-/single/check-submitted');
+
+    // then
+    assert.equal(res.status, 200);
+    assert.isEmpty((await res.text()).trim());
```
matthew-white, line 165: Shouldn't `res.text()` include the HTML? Or does `res.text()` only return the inner content of the `<body>` element? Is there a way for the test to verify that blank.html is being served?

test/nginx/test-nginx.js
```diff
@@ -156,6 +156,15 @@ describe('nginx config', () => {
     });
   });
 
+  it('should serve blank page on /-/single/check-submitted', async () => {
+    // when
+    const res = await fetchHttps('/-/single/check-submitted');
+
+    // then
```
matthew-white, line 163: How about also asserting:

```js
await assertEnketoReceivedNoRequests();
```

files/nginx/odk.conf.template
```diff
@@ -65,6 +65,11 @@ server {
     # Form fill link (non-public), or Draft
     return 301 "/f/$enketoId/new$is_args$args";
   }
+  # To read single submission cookies
+  location = /-/single/check-submitted {
+      alias /usr/share/nginx/html/blank.html;
+      default_type text/html;
```
matthew-white, line 71: It looks like this is indented by 4 spaces rather than 2.

matthew-white (approved):
The change to the nginx config here matches the change in getodk/central-frontend#1186, which I also reviewed.

test/nginx/test-nginx.js
sadiqkhoja: When `SKIP_FRONTEND_BUILD` is true, then `build-frontend.sh` creates an empty blank.html

PR #982, opened 2025-04-24
ci: test secrets generation

alxndrsn:
@sadiqkhoja these tests demonstrate the issue you observed, and the code changes resolve the `PIPE` error that is thrown (on some systems).

It looks like SIGPIPE is generated when `head` completes more quickly than `tr`.

I think it's more concerning that enketo is happy to start when secrets are missing/empty.

Add tests.

An alternative would be to ignore SIGPIPE (13 / exit code 141).  This seems risky, as there may be other reasons for pipes to fail.

N/A?

No.


sadiqkhoja (approved):
I've tried the updated generate-secrets.sh and it no longer exits with 141 code. All three secret files are generated.

PR #988, opened 2025-04-25
enketo: fail startup if secrets missing

alxndrsn:
Relates to https://github.com/getodk/central/pull/982

- [x] added test for failure condition
- [x] existing tests
- [ ] `test-images.sh` should also be checking that the enketo service is starting ok

It's quite a simple approach.  An obvious alternative would be to move the check scripts to a `healthcheck` in the secrets service, and add `service_healthy` to enketo's `depends_on` config (see https://docs.docker.com/compose/how-tos/startup-order/).

That seems more complex, and unclear if that will work given that the `secrets` service shuts down after it has generated the secrets.

There's a danger that the approach in this PR introduces a race condition - perhaps the file/size checks should be repeated if they fail.

No.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -124,9 +124,8 @@ services:
     build:
       context: .
       dockerfile: enketo.dockerfile
-    restart: always
+    restart: no
     depends_on:
-      - secrets
       - enketo_redis_main
       - enketo_redis_cache
     environment:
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -124,9 +124,8 @@ services:
     build:
       context: .
       dockerfile: enketo.dockerfile
-    restart: always
+    restart: no
```
alxndrsn, line 127: only apply this in testing

test/test-enketo.sh
alxndrsn, line 1: this script needs to be executed in CI

PR #984, opened 2025-04-25
nginx: apply cache strategies to frontend and misc

alxndrsn:
Closes #980

#### What has been done to verify that this works as intended?

New tests and existing tests.

#### Why is this the best possible solution? Were any other approaches considered?

* simplifies and centralises common headers, as proposed at https://github.com/getodk/central/pull/936
* Replacing `if {}` blocks solves a VERY surprising* number of footguns.

\* maybe not surprising to experienced nginx configurers, as there is plenty of documentation of `if`'s issues, e.g.

- https://github.com/nginxinc/nginx-wiki/blob/master/source/start/topics/depth/ifisevil.rst
- https://news.ycombinator.com/item?id=37235738

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

Should mean less caching issues.

#### Does this change require updates to documentation? If so, please file an issue [here](https://github.com/getodk/docs/issues/new) and include the link below.

No.

#### Before submitting this PR, please make sure you have:

- [x] branched off and targeted the `next` branch OR only changed documentation/infrastructure (`master` is stable and used in production)
- [x] verified that any code or assets from external sources are properly credited in comments or that everything is internally sourced

PR #991, opened 2025-04-28
nginx: add test for proxy_redirect config

alxndrsn:
Noted while working on https://github.com/getodk/central/pull/984

- [ ] added new passing tests
- [ ] removed the `proxy_redirect` directive and checked the new tests now fail

This rule's meaning is subtle, and it's important that it continues to work.

### DIFF
--- a/test/nginx/mock-http-server/index.js
+++ b/test/nginx/mock-http-server/index.js
@@ -16,6 +16,12 @@ app.get('/reset',       withStdLogging((req, res) => {
 
 app.get('/v1/reflect-headers', withStdLogging((req, res) => res.json(req.headers)));
 
+app.get('/v1/generate-redirect/:status', withStdLogging((req, res) => {
+  const { status } = req.params;
+  const { location } = req.query;
+  res.redirect(status, location);
+}));
+
 [
   'delete',
   'get',
--- a/test/nginx/test-nginx.js
+++ b/test/nginx/test-nginx.js
@@ -257,6 +257,21 @@ describe('nginx config', () => {
     socket.on('error', reject);
   }));
 
+  [
+    '/some/path',
+    'https://example.org/some/path',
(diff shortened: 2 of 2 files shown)


matthew-white:
Just FYI, we're currently thinking that we should wait for .2 to try to merge this PR. We're hoping to start regression testing early next week. Let me know if that doesn't sound right!

matthew-white:
@alxndrsn, we're planning to start regression testing next Tuesday or Wednesday, so I wanted to check about the status of this PR. It's ready for review, right? Looking at the diff count of +138 -2 made me think that it's a big change, but actually only 2 lines of source changed. The rest is tests.

Who would work best to review this PR? Sadiq?

Could you clarify whether the following is still a TODO?

> - [ ] consider if this change is helpful

It sounds like you think this change would be helpful, right? Or is that still an open question?

test/nginx/mock-http-server/index.js
```diff
@@ -35,6 +35,16 @@ app.get('/v1/projects', (_, res) => {
   res.send('OK');
 });
 
+const redirectGenerator = withStdLogging((req, res) => {
```
alxndrsn: `withStdLogging()` no longer exists.  Logging middleware should handle this automatically.

alxndrsn: Removed.

matthew-white:
> IMO it's helpful

Sounds good. :+1: I went ahead and tagged @sadiqkhoja for review given his knowledge of nginx. If this PR could get reviewed and merged tomorrow, then we could include it in the v2025.2 release. If not, I think it'd also work well to include it in .3.

matthew-white:
Regression testing has started for v2025.2, so I think we should hold off on merging this PR. I'll move it to v2025.3 in the GitHub project. I'm still thinking that @sadiqkhoja is probably the right person to review.

PR #987, opened 2025-04-25
test/nginx: add explicit checks for CSP contents

alxndrsn:
Noted while working on https://github.com/getodk/central/pull/986

It can be hard to reason about exactly where a rewrite rule will end up being processed.  These test additions ensure that the correct CSP is applied to various paths.

#### What has been done to verify that this works as intended?

Ran the updated tests.

#### Why is this the best possible solution? Were any other approaches considered?

This should make the current definition of CSPs easier to read, and make future changes easier to track.

Previously discussed at https://github.com/getodk/central/pull/936#discussion_r2015734746

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

No effect.

#### Before submitting this PR, please make sure you have:

- [x] branched off and targeted the `next` branch OR only changed documentation/infrastructure (`master` is stable and used in production)
- [x] verified that any code or assets from external sources are properly credited in comments or that everything is internally sourced

PR #992, opened 2025-04-28
nginx: simplify enketo passthrough rewrites

alxndrsn:
Noted while working on https://github.com/getodk/central/pull/987

Current tests continue to pass.

Using `rewrite ... last` and separating the two blocks means that passthrough URLs are now rewritten, and the entire config file re-evaluated for the new URLs:

> stops processing the current set of `ngx_http_rewrite_module` directives and starts a search for a new location matching the changed URI; 
> \- _https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite_

This is in contrast to the current approach using `rewrite ... break`, which:

> stops processing the current set of `ngx_http_rewrite_module directives` as with the [`break`](https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#break) directive;
> \- _https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite_

and the `break` directive is defined as:

> If a directive is specified inside the [location](https://nginx.org/en/docs/http/ngx_http_core_module.html#location), further processing of the request continues in this location.

Which means that `break`, unintuitively, allows continuing of the current `location` block.


files/nginx/odk.conf.template
```diff
@@ -117,8 +117,11 @@ server {
 
   # For that iframe to work, we'll need another path prefix (enketo-passthrough) under which we can
   # reach Enketo — this one will not be intercepted.
-  location ~ ^/(?:-|enketo-passthrough)(?:/|$) {
-    rewrite ^/enketo-passthrough(/.*)?$ /-$1 break;
+  location ~ ^/enketo-passthrough(/.*)?$ {
+    rewrite ^/enketo-passthrough(/.*)?$ /-$1 last;
+  }
+
+  location ~ ^/-(/.*)?$ {
```
sadiqkhoja, line 124: I think this can be simplified:

```suggestion
  location /- {
```

sadiqkhoja (approved):

files/nginx/odk.conf.template
alxndrsn: That change fails the following tests:

```
  1) nginx config
       should not forward to enketo; /--/:
...
  2) nginx config
       should not forward to enketo; /-some:
```

PR #996, opened 2025-04-29
ci: test/nginx: precompute DH keys for faster testing

alxndrsn:
Time taken for Diffie Helman key generation can take minutes, and is a frequent cause of test timeouts.  Pre-generating them and including the generated file in the test nginx container significantly speeds up the nginx test phase.  Key generation is still testing in the later test-images step of CI.

This change also allows for various timeouts to be decreased.

Example test failures:

* https://github.com/getodk/central/actions/runs/14724159913/job/41323307777

To demonstrate the problem locally (output times in `minutes:seconds`):

```sh
$ for i in {0..9}; do /bin/env time --output=/dev/stdout --format "$i: %E" openssl dhparam -out /dev/null 2048 2>/dev/null; done
0: 0:21.53
1: 0:28.02
2: 0:21.01
3: 0:11.20
4: 0:05.08
5: 0:08.13
6: 0:28.23
7: 0:46.23
8: 1:02.59
9: 0:08.63
```

Tests!

An alternative would be to configure a smaller DH key size when running tests.  This would introduce more complexity into the production script at `files/service/scripts/start-odk.sh`, and risk generating small keys in production if misconfigured.  There may also be reasons why smaller keys could be rejected by nginx; I did not attempt this path.

No change.

No.


sadiqkhoja:
I don't see any harm in having dh.pem in repository but another option is to generate it once via `run-tests.sh` before starting docker containers

alxndrsn:
> another option is to generate it once via `run-tests.sh` before starting docker containers

I think this would have the same problem of taking somewhere between 1 second and a few minutes.

PR #1002, opened 2025-04-29
nginx/enketo: strip upstream cache headers

sadiqkhoja:
Without this PR: nginx returns multiple Vary and Cache-Control headers:

```
curl -I https://staging.getodk.cloud/-/x/css/theme-kobo.css
```

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -118,6 +118,8 @@ server {
     proxy_pass http://enketo:8005;
     proxy_redirect off;
     proxy_set_header Host $host;
+    proxy_hide_header Vary;
+    proxy_hide_header Cache-Control;
 
     # More lax CSP for enketo-express:
     # Google Maps API: https://developers.google.com/maps/documentation/javascript/content-security-policy
--- a/test/nginx/mock-http-server/index.js
+++ b/test/nginx/mock-http-server/index.js
@@ -12,6 +12,13 @@ app.use((req, res, next) => {
   next();
 });
 
+// Enketo express returns response with Vary and Cache-Control headers
+app.use('/-/', (req, res, next) => {
+  res.set('Vary', 'Accept-Encoding');
+  res.set('Cache-Control', 'public, max-age=0');
+  next();
+});
+
 app.get('/health',      (req, res) => res.send('OK'));
 app.get('/request-log', (req, res) => res.json(requests));
 app.get('/reset',       (req, res) => {


test/nginx/test-nginx.js
```diff
@@ -451,6 +451,16 @@ describe('nginx config', () => {
         });
       });
     });
+
+    it('should have only one vary header', async () => {
+      // when
+      const res = await fetchHttps('/-/x/css/theme-kobo.css');
```
alxndrsn, line 457: I'm in two minds about this test - it's nice that it's explicit, but I'd expect the suite `enketo caching` would already cover this.

test/nginx/test-nginx.js
```diff
@@ -451,6 +451,16 @@ describe('nginx config', () => {
...
+
+      // then
+      assert.equal(res.status, 200);
+      // and
+      assertCacheStrategyApplied(res, 'single-use');
```
alxndrsn, line 462: Is it weird that CSS for single-use forms is served from a different URL to the same CSS for a multi-use form?  In an ideal world, this file would be served with strategy `revalidate`...

test/nginx/mock-http-server/index.js
```diff
@@ -7,6 +7,15 @@ const requests = [];
 
 const app = express();
 
+// Enketo express returns response with Vary and Cache-Control headers
+app.use((req, res, next) => {
+  if (req.path.startsWith('/-/')) {
```
alxndrsn, line 12: I think there's a more idiomatic way of specifying this, should be something like

test/nginx/mock-http-server/index.js
```diff
@@ -7,6 +7,15 @@ const requests = [];
...
+    res.set('Vary', 'Accept-Encoding');
+    res.set('Cache-Control', 'public, max-age=0');
```
alxndrsn, line 14: Looks good :+1:

alxndrsn (changes requested):
Some minor queries; otherwise looking good :+1:

test/nginx/test-nginx.js
sadiqkhoja: #1005 addressed this

test/nginx/test-nginx.js
sadiqkhoja: removed this test as #1005 added this case in `enketo caching` suite

alxndrsn:
Suggested PR/commit title:

```
nginx: strip upstream enketo cache headers
```

alxndrsn (approved):

files/nginx/odk.conf.template
```diff
@@ -118,6 +118,8 @@ server {
     proxy_pass http://enketo:8005;
     proxy_redirect off;
     proxy_set_header Host $host;
+    proxy_hide_header Vary;
+    proxy_hide_header Cache-Control;
```
alxndrsn, line 122: N.B.:

* `proxy_set_header` sets a **_request_** header; 
* `proxy_hide_header` removes a **_response_** header :facepalm:

PR #1018, opened 2025-05-02
Revert "nginx: simplify enketo passthrough rewrites"

sadiqkhoja:
Reverts getodk/central#992

After `rewrite ^/enketo-passthrough(/.*)?$ /-$1 ....`, we don't want enketo redirection rules (`/-/...`  -> `/f/...` ) to execute again

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -110,11 +110,8 @@ server {
 
   # For that iframe to work, we'll need another path prefix (enketo-passthrough) under which we can
   # reach Enketo — this one will not be intercepted.
-  location ~ ^/enketo-passthrough(/.*)?$ {
-    rewrite ^/enketo-passthrough(/.*)?$ /-$1 last;
-  }
-
-  location ~ ^/-(/.*)?$ {
+  location ~ ^/(?:-|enketo-passthrough)(?:/|$) {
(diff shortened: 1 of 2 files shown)


matthew-white (commented):
I'm not sure I'm understanding all the details here, but I do see that this PR reverts #992. I'll go ahead and approve so that @sadiqkhoja can get this on staging while @sadiqkhoja and @alxndrsn continue discussing.

matthew-white (approved):
Actually approving now 😅

alxndrsn:
> After `rewrite ^/enketo-passthrough(/.*)?$ /-$1 ....`, we don't want enketo redirection rules (`/-/...` -> `/f/...` ) to execute again

@sadiqkhoja can you give a worked example of what the problem was?  It would be great to have a regression test to prevent this bug from being reintroduced.

sadiqkhoja:
The test added in this PR isn't it enough?

If you want to see the actual problem, I have patched the change of #992 on dev server, over there click on the "New Submission" button, you will see the iframe "inception" i.e. `/enketo-passthough/:enketoId` returns 301 `/f/:enketoId` which loads `/enketo-passthough/:enketoId` in an iframe, which again returns 301 and the loop continues forever.

Issue #1048, opened 2025-05-09
List of Projects on Main page takes much longer to load


srujner:
**Problem description**
List of Projects on Main page takes much longer to load. You can notice spinning wheel next to the Projects tab and that it takes few seconds before loading the whole page.
In addition to the main page, we have a general impression that the application has started to run more slowly. You can see when switching between tabs or when WebForms are loading, but it'll probably need some performance tool to check it with details

**Steps to reproduce the problem**

1. Navigate to any project page (00:00)
2. Click on the ODK Central logo (00:01)
3. Observe that the projects are loading for several seconds (00:03)

https://jam.dev/c/87812fa1-b129-486f-9e21-1a381a2fcbf1

**Expected behavior:**
Projects should load faster

**Versions:**
7114a06d0e28a4bfeef7d9d3336099cdece6ab6e (v2024.3.2-52-g7114a06)
+409474ce8f6695690f1388f9c2f62b90b0f3386f client (v2024.3.1-118-g409474ce)
+3e5ead30df49fd79d42b78a469ee5dae2ce3ed97 server (v2024.3.1-94-g3e5ead30)

**Other Information**
I think it'll be good to check performance when there is more than 1k projects and 10k projects to see how many second will take the page to completely load.

lognaturel:
I wonder whether this might be psychological. I believe previously either the whole page or more of the page was blocked on getting the project count. The progress indicator is new and ironically I think it makes one more aware of the operation taking time.

Luckily, it's rare to have more than a handful of projects so I don't think we need to dig much further into this now.

> it'll probably need some performance tool to check it with details

Yes, and unfortunately I think it would also require a baseline from the prior version. That's made more complicated by the fact that we're constantly changing data on that server which also affects performance.

lognaturel:
Ah. Nope, there's a significant difference here which is that the project list used to respond with a 304 and now it always gives a fresh response. On the staging server that's 1.39MB that needs to be downloaded every time.

I wonder whether we may have had a misunderstanding around intended caching behavior. From my perspective the intent was for client caching to still happen but for revalidation with the server to always occur. Maybe we now have no client caching?

I'm also seeing that js and css are now uncached. Maybe that's pending #1007 though

lognaturel:
I see `Cache-Control: no-store` in the response headers for every request so that's likely the issue. That's being sent in addition to `Cache-Control: private, no-cache`. I also see `Pragma: no-cache` I don't know where those would be coming from.

[...]

PR #1059, opened 2025-05-12
Pass through backend cache headers

lognaturel:
Closes https://github.com/getodk/central/issues/1048

Put it on the dev server. Verified that there's exactly one `Cache-Control` header and that it's set as expected.

We could also set the headers here to `private, no-cache`, etc. But then we'd be setting duplicate headers which could drift between nginx and backend. I think it's clearer to make caching headers clearly the backend's responsibility.

Note that this does not set the `Pragma` header for backend. I think we could do away with it altogether because it has been deprecated for a long time.

Backend responses should be cached in private caches. We should get 304s.

No.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -14,6 +14,9 @@ map "$request_method::$uri$is_args$args" $cache_strategy {
   ~^(GET|HEAD)::/robots.txt$         "revalidate";
   ~^(GET|HEAD)::/version.txt$        "revalidate";
 
+  # central-backend
+  ~^(GET|HEAD)::/v1/                 "passthrough";
+
   # central-frontend - unversioned
   ~^(GET|HEAD)::/$             "revalidate";
   ~^(GET|HEAD)::/index.html$   "revalidate";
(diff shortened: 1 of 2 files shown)


files/nginx/odk.conf.template
```diff
@@ -41,16 +44,19 @@ map "$request_method::$uri$is_args$args" $cache_strategy {
 map $cache_strategy $cache_header_cache_control {
   "immutable"  "max-age=31536000";
   "revalidate" "no-cache";
+  "passthrough"  "";
```
lognaturel, line 47: I verified that if there's no value set for a header, that header isn't set by nginx.

files/nginx/odk.conf.template
```diff
@@ -41,16 +44,19 @@ map "$request_method::$uri$is_args$args" $cache_strategy {
 map $cache_strategy $cache_header_cache_control {
   "immutable"  "max-age=31536000";
```
lognaturel, line 45: I think this should be `public, max-age=31536000, immutable` otherwise we don't know if intermediary caches will cache.

files/nginx/odk.conf.template
```diff
@@ -41,16 +44,19 @@ map "$request_method::$uri$is_args$args" $cache_strategy {
...
   "revalidate" "no-cache";
```
lognaturel, line 46: Same as above, I think we can/should add `public`.

files/nginx/odk.conf.template
sadiqkhoja: I see a PR to update nginx docs: "The add_header directive will not generate a header if the value provided is an empty string." except if it is multi-value header

https://github.com/nginx/nginx.org/pull/51

files/nginx/odk.conf.template
sadiqkhoja: makes sense

sadiqkhoja (approved):
Looks good to me. However, we should add one or two tests in `test-nginx.js`

lognaturel:
@sadiqkhoja please verify the tests and either open a PR to fix or send me some comments if they're not right/complete!

PR #1066, opened 2025-05-14
Allow passthrough of command arguments to postgres14 entrypoint

spwoodcock:
#### This PR

- There is no linked issue to this.
- It's a minor fix for convenience.

##### The Issue

- We are doing a mapping campaign with 450 simultaneous users.
- We have to support making submissions to ODK Central via the **REST API** from many users at once.
- When load testing we get some unexpected disconnects from Central --> Postgres.
- This is likely due to the default `max_connections=100` for Postgres.

##### The Solution

- Ideally we want to bump the max_connections, but there is no way to do this with the current compose / entrypoint setup.
- Adding `$@` to the exec of the original entrypoint in the postgres container, allows for passing through config params to postgres (the postgres container supports passing `$@` upstream).
- As a result, we can do something like this:

docker-compose.override.yml
```yaml
services:
  postgres14:
    command: start-postgres.sh -c 'max_connections=300'
```

- "$@" won't evaluate to anything for any existing setups, so this will change nothing.

- Standard practice in container entrypoints to do this.

- No change.

- Probably not - it's an advanced config option.


alxndrsn (approved):
LGTM - thanks for this :+1:

PR #1107, opened 2025-05-27
Fix: remove edit submission redirection from the nginx config

sadiqkhoja:
Adding `/-/edit/<enketo-id>` in the negative look ahead is not necessary because it doesn't matches `/-/[a-zA-Z0-9]+$` pattern.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -96,10 +96,6 @@ server {
     # Form fill link, public
     return 301 "/f/$enketoId$is_args$args";
   }
-  location ~ "^/-/edit/(?<enketoId>[a-zA-Z0-9]+)$" {
-    # Edit link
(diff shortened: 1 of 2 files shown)


lognaturel:
> Adding /-/edit/<enketo-id> in the negative look ahead is not necessary

I see! So `preview` can be removed as well, right?

lognaturel:
I have verified that this does what I expect. I continue not to see any risk -- it's reestablishing the prior behavior. Have deployed it to fix immediate issue and we can merge once that's validated.

lognaturel (approved):
Confirmed to have fixed the user-reported error!

We have one other commit merged to `next` for the point release so I think it makes most sense to merge there. @matthew-white?

matthew-white:
> We have one other commit merged to `next` for the point release so I think it makes most sense to merge there.

Agreed, let's merge into `next`!

PR #1110, opened 2025-05-29
Fix: don't check enketo keys size when ENV is DEV

### LINKED ISSUE #1109: Enketo doesn't start with development docker config
Running `make dev` as per docs should start all required services in docker. But Enketo doesn't start because secret keys generated for development ([here](https://github.com/getodk/central/blob/master/docker-compose.dev.yml#L40)) don't pass checks in `start-enketo.sh` ([here](https://github.com/getodk/central/blob/master/files/enketo/start-enketo.sh#L28)).

Logs of enketo container:

```
[start-enketo.sh] Checking secrets exist...
[start-enketo.sh] !!!
[start-enketo.sh] !!! Unexpected file size:
[start-enketo.sh] !!!   file: /etc/secrets/enketo-secret
[start-enketo.sh] !!!   expected: 64 b
[start-enketo.sh] !!!   actual:   18 b
[start-enketo.sh] !!!
```

yoursdearboy:
Closes #1109

Successfully started enketo container with `docker-compose.dev.yml`

`docker-compose.dev.yml` defines `ENV` variable equal to `DEV`, so it can be used in `start-enketo.sh` script, with a default value of `PROD`.

Another approach would be to use a predefined key with a proper size of 128 bytes instead of `enketorules`, but it must be done across configs and is a kind of breaking change (for development environment).

It doesn't affect users because if `ENV` variable isn't defined, keys will be validated as before.

#### Does this change require updates to documentation? If so, please file an issue [here](https://github.com/git odk/docs/issues/new) and include the link below.

No.

### DIFF
--- a/files/enketo/start-enketo.sh
+++ b/files/enketo/start-enketo.sh
@@ -25,9 +25,12 @@ assert_size() {
     exit 1
   fi
 }
-assert_size /etc/secrets/enketo-secret       64
-assert_size /etc/secrets/enketo-less-secret  32
-assert_size /etc/secrets/enketo-api-key     128
+
+if ! [[ "${ENV:-PROD}" = "DEV" ]]; then
+  assert_size /etc/secrets/enketo-secret       64
+  assert_size /etc/secrets/enketo-less-secret  32
+  assert_size /etc/secrets/enketo-api-key     128
+fi
 
 CONFIG_PATH=${ENKETO_SRC_DIR}/config/config.json
 log "Generating enketo configuration..."


lognaturel:
Thanks for filing issues and suggesting fixes as you get your dev environment set up! I want to let you know that we are working on a tight timeline currently and may not have a chance to get to all your PRs in a timely manner. We do see you and appreciate you!

This one looks like a fine approach to me and I've looped in @alxndrsn for a final review.

alxndrsn:
Thanks for the PR - the general idea seems OK.

> `docker-compose.dev.yml` defines `ENV` variable equal to `DEV`, so it can be used in `start-enketo.sh` script, with a default value of `PROD`.
> 
> Another approach would be to use a predefined key with a proper size of 128 bytes instead of `enketorules`, but it must be done across configs and is a kind of breaking change (for development environment).

I can't actually see this `ENV` is used anywhere else, and I would prefer either:

1. a more explicit env var/value combo, e.g. `ENKETO_SECRETS=insecure`, or
2. your alternate approach of using predefined secrets which actually fulfil expectations

My preference would be for approach **2**, as it means less complexity in the code, and dev environments aligning more closely with production environments.  I appreciate your caution around this, but IMO it would be worth the temporary incompatibilities it would introduce between different versions in development.

If going with approach **1**, the current PR will log `Checking secrets exist...`, even when it is not.  It would be preferable to either:

1. only log when secrets are checked:

```sh
if [[ "${ENKETO_SECRETS-}" = insecure ]]; then
  log "Skipping secrets check."
else
  log "Checking secrets..."
  ...
fi
```

or

2. check the new expectations for the secrets:

```sh
log "Checking secrets..."
if [[ "${ENKETO_SECRETS-}" = insecure ]]; then
  assert_size /etc/secrets/enketo-secret       17
  assert_size /etc/secrets/enketo-less-secret  28
  assert_size /etc/secrets/enketo-api-key      11
else
  assert_size /etc/secrets/enketo-secret       64
  assert_size /etc/secrets/enketo-less-secret  32
  assert_size /etc/secrets/enketo-api-key     128
fi
```

PR #1144, opened 2025-06-10
Update news for v2025.1.3

lognaturel:
News update only


sadiqkhoja (approved):
What a coincidence — we are releasing every 10 days 😂

PR #1148, opened 2025-06-10
Bump shm

### LINKED ISSUE #1147: Running vacuum on built-in DB fails due to Docker's low shared memory limit
After upgrading to v2025.1.3, I attempted to run a vacuum on the built-in DB and it faled with...

> "ERROR:  could not resize shared memory segment "/PostgreSQL.1712702934" to 67145184 bytes: No space left on device"

By default, Docker limits the size of /dev/shm to 64MB, which is small for PostgreSQL. After I bumped it up to 512MB in docker-compose.yml, the vacuum worked great.

```
services:
  postgres14:
    build:
      context: .
      dockerfile: postgres14.dockerfile
    shm_size: 512m
    volumes:
      - postgres14:/var/lib/odk/postgresql/14

```

I settled on 512MB because ChatGPT says that's what a real app with a light load would need. You can do more but you are likely overcommitting resources...

yanokwa:
Closes #1147 

Vacuum runs on at least two self-hosted instances with a lot of data.

ChatGPT says 512MB seems to be the most reasonable for real apps with light workloads

Probably makes for faster performance. I wonder if auto vacuums have been failing on a bunch of servers.
 
Nah.


matthew-white:
@yanokwa, I've merged `master` into `next` now that the patch release is out. Could you try rebasing onto `next`? Right now, it looks like there's a conflict.

yanokwa:
We don't have that many non-cloud instances where we could check if autovacuums are failing. The ones I've checked seem to, according to last_autovacuum pg_stat_user_tables, be running OK.

Yes, more SHM reduces available RAM on the host. Would you be happier with 128MB?

alxndrsn:
> Would you be happier with 128MB?

I don't have a preference, I just wanted to highlight the change as a potential extra answer to the template question:

> **How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?**

yanokwa:
https://docs.docker.com/reference/compose-file/build/
> shm_size sets the size of the shared memory (/dev/shm partition on Linux) **allocated** for building Docker images. 

I continue to think that 512MB is a reasonable default.

lognaturel:
> according to that short documentation blurb, shm_size would just be for building docker images

The link is to the docs for the shm_size directive in the build block but this is about the directive at the top level! Here are the docs: https://docs.docker.com/reference/compose-file/services/#shm_size I agree this one is likely a limit and not an allocation.

alxndrsn:
> I bet you could trigger the error at #1147 by changing smh_size to something small (default is 64MB) and try running a vacuum. @alxndrsn please give that a try when you have a chance.

Demonstrated at https://github.com/getodk/central/pull/1201.  Increase shm_size to `512m`, and the test will fail.

PR #1182, opened 2025-07-22
Fixes: Cache strategy for public offline Enketo Forms

sadiqkhoja:
Removed $ from the enketo cache strategy regex so that enketo offline line works with query parameter suffix as well

Closes [forum post](https://forum.getodk.org/t/after-upgrade-to-v-2025-2-1-public-access-link-of-forms-are-not-working-in-offline-mode/55810)

Added a test and manual verification on dev environment.

Simple solution. I was thinking about using regex with optional query parameters but that would make the regex complex without any benefits.

No

No


lognaturel (approved):
Thanks! Makes sense to me.

srujner:
@sadiqkhoja @lognaturel 
I have finished testing the main part of the issue and everything looks fine.
I haven't had time to look at the other query parameters yet.

If you are in a hurry and the answer to the comment above is yes, you can remove the ‘needs testing’ label. If not, I would check the other parameters tomorrow and then close the PR.

lognaturel:
Thanks! Yes, a public access link without the offline URL component should NOT work offline.

PR #1183, opened 2025-07-22
Use bash from the PATH in shebang

sadiqkhoja:
TBA


sadiqkhoja:
from the article: 

> The only reason to start your script with '#!/usr/bin/env <whatever>' is if you expect your script to run on a system where Bash or whatever else isn't where you expect (or when it has to run on systems that have '<whatever>' in different places, which is probably most common for third party packages). Broadly speaking this only happens if your script is portable and will run on many different sorts of systems.

`test/nginx/run-tests.sh` is certainly needs to be executable on multiple systems. 

Can I merge this PR @alxndrsn @brontolosone ? or keep making local change whenever I have to execute `run-tests.sh`?

Issue #1165, opened 2025-07-04
Process submission geometry for mapping


lognaturel:
[Brainstorming doc](https://docs.google.com/document/d/1G_ytR1c2yiMvmsFxJrgvjRE3c70bOve3MwoIO0SuQOk/edit?tab=t.0)

### Goal
Prepare submission data for mapping in a performant way. A submission will be mapped by first geometry field (geopoint/trace/shape) outside of any repeat, if any exists. This matches how Collect maps submissions although Collect currently only maps points.

### Related concepts
* Mapping Entities. We believe that using and indexing the well-known geometry json property if it exists will be performant enough
* Detecting self-intersecting polygons. Out of scope for this initial work but could affect the approach taken
* Detecting overlapping polygons in a submission set. We're not 100% sure we want to do this within Central but it's worth considering as an input for picking approach (e.g. does postgis help?)
* Compute/save centroids of traces and shapes
* Submission processing to create/update Entities, blob processing to send to S3
* Building a full read-only data replica of submissions for analysis/processing. We will use this smaller, more targeted processing task to explore ideas that could help support building a complete, normalized replica for analysis.

Form submissions are currently stored as XML blobs. Processing those on the fly is not performant enough to map submission sets of significant size so we will split out the first geometry field as part of async submission processing similar to Entity processing.

matthew-white:
Now that getodk/central-backend#1603 has been approved, I think this issue is ready to be closed. That PR contains the core of the Backend work for mapping that we'll be shipping with v2025.3.

@brontolosone, I see that you described that PR as "Towards #1165." Is there more work that you were thinking this issue entailed? I've filed the follow-up issues #1373 and #1377, but I think we can track those separately: we can close this issue before those are resolved.

> Form submissions are currently stored as XML blobs. Processing those on the fly is not performant enough to map submission sets of significant size so we will split out the first geometry field as part of async submission processing similar to Entity processing.

getodk/central-backend#1603 actually does end up processing XML on the fly. To address performance concerns, it caches processed geo data. The follow-up issues #1373 and #1377 are also related to performance.

PR #1186, opened 2025-07-24
use postgis/postgis:14-3.5 drop-in replacement for postgres:14.10 image

brontolosone:
Related: #1165
Related: getodk/central-backend#1555

With getodk/central-backend#1555 we'll need the PostGIS extensions in the DB container. This makes that happen.

Ran the previous setup, checked locale provider, ran the new setup with same DB files, checked locale provider (matched, unsurprisingly, as [postgis/postgis](https://hub.docker.com/r/postgis/postgis) base themselves on [_/postgres](https://registry.hub.docker.com/_/postgres/)), nothing weird in the logs, data still there, and I can run `create extension postgis`.

I can't imagine anything simpler/easier than using the drop-in.

Using the drop-in replacement by itself shouldn't really _do_ anything, other than allowing one to install the PostGIS extension. 

No getodk/docs update needed, but getodk/central-backend#1555 has a README.md update for those not using the Docker images.

### DIFF
--- a/postgres14.dockerfile
+++ b/postgres14.dockerfile
@@ -1,4 +1,4 @@
-FROM postgres:14.10
+FROM postgis/postgis:14-3.5
 
 COPY files/postgres14/start-postgres.sh /usr/local/bin/
 


matthew-white:
Just a note that the latest version of the Backend implementation, getodk/central-backend#1603, doesn't use PostGIS.

PR #1202, opened 2025-08-01
Set PostgreSQL application name

brontolosone:
Related: getodk/central-backend#1570

Ran the container constellation, connected to postgresql-14-1, witnessed `odkcentral` showing up in a `select distinct(application_name) from pg_stat_activity where datname = 'odk' order by application_name` query.

Well, you can set the application name in the connection string as well, but doing it through the environment inheritance is much easier, no code required. Perhaps we should do all connection parameters this way? And then we might automatically support all the things libpq supports, including client certs, local Unix sockets, etc.

Nothing unless some user's software depends on that application name not being set of course. And if they already had a PGAPPNAME env var, we're not overriding it.

No changes required.


files/service/scripts/start-odk.sh
```diff
@@ -72,6 +72,7 @@ determine_worker_count() {
 MEMTOT=$(get_memory_limit)
 WORKER_COUNT=$(determine_worker_count "$MEMTOT")
 export WORKER_COUNT
+export PGAPPNAME=${PGAPPNAME:-odkcentral}
```
alxndrsn: Would setting in `service.environment` in https://github.com/getodk/central/blob/master/docker-compose.yml#L49-L77 be more appropriate?  That way it will also be used for thing like `prune.js`.

brontolosone: Yes. :+1:

brontolosone: Please see 3f26dca.

matthew-white:
I might mention this change in release notes, so I'm going to add the PR to the project. (There's no associated issue, right? I usually only add a PR to the project when there's no GitHub issue.)

alxndrsn (changes requested):
All the other env vars are listed in alphabetical order, so this one should probably be too.

Apart from that, this PR looks great 👍

PR #1310, opened 2025-08-26
nginx: enable HTTP2

### LINKED ISSUE #1309: nginx: enable HTTP3
From Lighthouse:

# HTTP1

Image

Image

# HTTP2

Image

Image

alxndrsn:
Closes #1309

Manual testing on test instance.

The only other approach attempted was `listen 443 ssl http2`, but this seems to be deprecated syntax.

* it should improve performance for users, but on lossy networks it may have the opposite effect
* it might increase the number of concurrent connections to node, using more server RAM

No - any HTTP1-only clients should still continue to function as before; HTTP2-supporting clients like web browsers should get improved performance transparent to the user.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -61,6 +61,7 @@ map $cache_strategy $cache_header_vary {
 
 server {
   listen 443 ssl;
+  http2 on;
   server_name ${DOMAIN};
 
   ssl_certificate /etc/${SSL_TYPE}/live/${CERT_DOMAIN}/fullchain.pem;


docs/news.html
```diff
@@ -7,15 +7,15 @@
...
+        ODK Central v2025.2.2
       </a>
     </div>
     <div class="news-item">
-      <span class="item-date">2025/06/17</span>
-      <a href="https://forum.getodk.org/t/55316" target="_blank">
-        ODK Central v2025.2
+      <span class="item-date">2025/07/24</span>
+      <a href="https://forum.getodk.org/t/55316/3" target="_blank">
+        ODK Central v2025.2.1
```
alxndrsn: revert

client
alxndrsn, line 1: Revert

alxndrsn:
Enabling HTTP2 **feels** like it makes Central much less reliable (in Chromium).  This may be a client issue?

E.g. on page reload:

Screenshot_2025-08-28_10-20-42

lognaturel:
@getodk/testers @grzesiek2010 @seadowg `staging` should soon be using HTTP/2. I learned that HTTP/3 seems to be experimental in `nginx` and would like to get clarity on that before we explore any further (https://github.com/getodk/central/pull/1328#issuecomment-3378067472).

As @alxndrsn and I wrote, our current understanding suggests that there should be no negative impact for Collect. If you notice any change positive or negative, please do flag it so we can investigate. This might be something like much longer or shorter form update times, more/less frequent failures when trying to send, things like that. If something feels different to you, it's worth noting it. The most likely outcome is that you won't feel any difference at all!

PR #1314, opened 2025-08-27
Fix: redirection for Enketo based on st parameter

sadiqkhoja:
Builds on https://github.com/getodk/central-frontend/pull/1325.

It solves the issue mentioned in https://forum.getodk.org/t/public-access-links-with-last-saved-not-working-as-expected-in-enketo-in-central-v2025-2/56351 where old public with `/single` prefix removed is redirected to the login page.

Added tests and manually verified the change.

Besides checking existence of `st` query parameter we don't have anything to know the intention of the users.

None

https://github.com/getodk/docs/issues/1985

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -59,26 +59,19 @@ map $cache_strategy $cache_header_vary {
   default      "*";
 }
 
-map $args $args_without_single {
-  ~^single=[^&]*&?(.*)$ $1;
-  ~^(.*)&single=[^&]*&?(.*)$ $1&$2;
-  ~^(.*)&single=[^&]*$ $1;
-  default $args;
-}
-
-map $args_without_single $is_args_without_single {
+map $args $qp_deliminator {
   ~.+ "&";
-  default "";
+  default "?";
 }
 
 map $arg_st $redirect_non_single_prefix {
-  ~.+     "?single=false$is_args_without_single$args_without_single";
-  default "/new$is_args$args";
+  ~.+     "${is_args}${args}${qp_deliminator}single=false";
+  default "/new${is_args}${args}";
 }
 
 map $arg_st $redirect_single_prefix {
-  ~.+     "$is_args$args";
-  default "/new?single=true$is_args_without_single$args_without_single";
+  ~.+     "${is_args}${args}";
+  default "/new${is_args}${args}${qp_deliminator}single=true";
 }
 
 server {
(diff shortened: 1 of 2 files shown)


files/nginx/odk.conf.template
```diff
@@ -59,14 +59,26 @@ map $cache_strategy $cache_header_vary {
...
+}
+
+map $arg_st $redirect_non_single_prefix {
+  ~.+     "?single=false$is_args_without_single$args_without_single";
+  default "/new$is_args$args";
+}
+
+map $arg_st $redirect_single_prefix {
+  ~.+     "$is_args$args";
+  default "/new?single=true$is_args_without_single$args_without_single";
```
matthew-white: Do you think we could get by without stripping the `single` query parameter? If a user is using an old URL like /-/single/:enketoId, I wouldn't expect them to also specify a `single` query parameter. /-/single/:enketoId is an old-style URL, whereas `?single` is from the land of the new-style /f/:enketoId URLs.

The logic of `$args_without_single` strips out the first `single` query parameter, but if `single` is specified twice, the second `single` query parameter will still come through. So I feel like we aren't fully preventing the user from passing through `single` if they really want to. If we don't expect them to try to do so, I feel like we could simplify the logic and remove `$args_without_single`. If for some reason they do try to specify `single`, it will end up being an array query parameter value in Vue Router, and `EnketoIframe` will fall back to the default behavior (because the value is not the string `'true'` or `'false'`).

If it's important to keep the stripping of the query parameter, I think tests of that case would be helpful.

files/nginx/odk.conf.template
```diff
@@ -94,7 +116,9 @@ server {
   # Following are the locations that serve a Form and these are redirected to the frontend:
   location ~ "^/-/single/(?<enketoId>[a-zA-Z0-9]+)$" {
     # Form fill link, public
-    return 301 "/f/$enketoId$is_args$args";
+    # If 'st' query parameter is not present, redirect to protected route with single only
+    # end-of-form behavior (/new/single)
```
matthew-white:
```suggestion
    # end-of-form behavior (/new?single=true)
```

files/nginx/odk.conf.template
```diff
@@ -105,7 +129,8 @@ server {
   # we don't want them to be redirected to central-frontend
   location ~ "^/-/(?!thanks$|connection$|login$|logout$|api$|preview$)(?<enketoId>[a-zA-Z0-9]+)$" {
     # Form fill link (non-public), or Draft
-    return 301 "/f/$enketoId/new$is_args$args";
+    # If 'st' query parameter is present, redirect to /multiple route for public access
```
matthew-white:
```suggestion
    # If 'st' query parameter is present, add ?single=false for public access
```

files/nginx/odk.conf.template
sadiqkhoja: I didn't thought about someone passing double `single`, I agree with you that it would be a lot simpler not to strip out `single`.

matthew-white:
I think it's OK to merge this into the `next` branch, there haven't been any other commits there. We can release the patch off `next`.

PR #1321, opened 2025-08-29
test/nginx: pin node version with volta

alxndrsn:
Running with e.g. node v18 gives:

	Exception during run: Error [ERR_REQUIRE_ESM]: require() of ES Module node_modules/chai/chai.js from test-nginx.js not supported.

Volta is the standard used across other `odk-central` repos.

Node v22.16.0 is the current version used in CI by this test suite.

Closes #


alxndrsn:
> do you think it'd be possible and worth it to specify just the major version?

I don't know if it's possible, and it doesn't seem helpful as different major node versions have different support for core functionality.

Specifying patch version is done similarly in `volta` settings in `package.json` files in related projects:

* https://github.com/getodk/central-backend/blob/c4a684089b7f3913ce6115dbd958b9bf908e9960/package.json#L8-L10
* https://github.com/getodk/central-frontend/blob/c15cc54e6995af6f86edb63f883a12ce4ad13689/package.json#L14-L16
* https://github.com/getodk/web-forms/blob/6015ab6b3cb9ccc5b79ad4a453e95c6cae6c79f3/package.json#L14-L17

It is also what the tests are run against in CI:

* https://github.com/getodk/central/blob/master/.github/workflows/test.yml#L37-L39

And how CI is set up for other related projects:

* https://github.com/getodk/central-backend/blob/c4a684089b7f3913ce6115dbd958b9bf908e9960/.github/workflows/standard-suite.yml#L28-L31
* https://github.com/getodk/central-frontend/blob/c15cc54e6995af6f86edb63f883a12ce4ad13689/.github/workflows/tests.yml#L42-L47
* https://github.com/getodk/web-forms/blob/6015ab6b3cb9ccc5b79ad4a453e95c6cae6c79f3/.github/workflows/ci.yml#L93-L95

Happy to discuss specifying version numbers more generally in a separate issue, but I think for these tests:

1. it's helpful to have some version pinned, and
2. following established convention is a good starting point

matthew-white:
Just to make sure I understand the use case, this is when you're running tests locally, right? I agree that it makes sense to try to reuse the same local version of Node in that case.

matthew-white (approved):

PR #1329, opened 2025-09-02
Increase gzip level and types

yanokwa:
gzip_comp_level of 4-6 seems to be the sweet spot for compression and CPU usage. We don't often see huge CPU usage on servers so 5 seems pretty safe. 

gzip_proxied makes sure to gzip anything coming from upstream (reverse proxy mode)

gzip_min_length 1024 is a common default. 1280 suggests we've profiled the responses and we are making a sensible choice, but we haven't. Might as well use the common default.

I've added geojson and svg because we'll be serving those in the future.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -81,10 +81,12 @@ server {
   client_max_body_size 100m;
 
   gzip on;
+  gzip_comp_level 5;
   gzip_vary on;
-  gzip_min_length 1280;
+  gzip_min_length 1024;
   gzip_http_version 1.1;
-  gzip_types text/plain text/css application/json application/x-javascript application/javascript text/xml text/csv;
+  gzip_types text/plain text/css application/json application/geo+json application/x-javascript application/javascript text/xml text/csv image/svg+xml;
+  gzip_proxied any;
 
   # Enketo Configuration.
   # Enketo express is traditionally served at /- but with the introduction of ODK Web Forms


files/nginx/odk.conf.template
```diff
@@ -81,10 +81,12 @@ server {
...
 
   gzip on;
+  gzip_comp_level 5;
   gzip_vary on;
-  gzip_min_length 1280;
+  gzip_min_length 1024;
   gzip_http_version 1.1;
-  gzip_types text/plain text/css application/json application/x-javascript application/javascript text/xml text/csv;
+  gzip_types text/plain text/css application/json application/geo+json application/x-javascript application/javascript text/xml text/csv image/svg+xml;
+  gzip_proxied any;
```
alxndrsn, line 89: What does this change?  Requests from central seem to already be gzipped.

yanokwa: Wasn't sure if data coming from service and enketo were gzipped. This seemed like a no-cost way to guarantee that.

alxndrsn: Is there something in front of test.getodk.cloud which could be zipping things?

yanokwa: Oh, I thought this was downstream. Upstream, if you are running a proxy or load-balanced in front of Central that might zip things? I'm happy to drop gzip_proxied. I agree it's not necessary.

alxndrsn: I think it's potentially a good addition, I'm just trying to understand why it doesn't appear to be necessary.  Potentially all zipping could be deferred to whatever is zipping downstream...

yanokwa: Any reason why it shouldn't be added? Seems like the only risk seems to be double-compression (unlikely because most modern proxies respect `Content-Encoding: gzip`) and maybe minor increase in nginx CPU.

matthew-white:
I'm going to add this PR to the project, since I'll probably mention it in the detailed release notes.

alxndrsn (approved):

PR #1366, opened 2025-09-17
Add cert that passes SSL checks

yanokwa:
Currently, Central fails [SSL Labs'](https://www.ssllabs.com/ssltest/analyze.html?d=production.getodk.cloud) test with
> Assessment failed: Internal Server Error

People (including [us](https://docs.getodk.org/security/#hosting-considerations)) refer to this test to show that our configs are secure. The problem seems to be that the dummy cert we use is too dumb.  

ChatGPT told me we need X.509 v3, a subjectAltName and be reasonably valid to work with SSL scanners, so I made those changes and confirmed that everything works.

Screenshot 2025-09-16 at 18 58 40

The lack of SNI means we don't support these old browsers: Android 2.3.7, IE 6 / XP, IE 8 / XP, Java 6u45.

We could get rid of the dummy block that was added at https://github.com/getodk/central/pull/814 and make the real vhost the default. Not 100% sure why that was strictly necessary to begin with. 

We could also add do something like this...
```
server {
  listen 443 ssl default_server;
  server_name ${CERT_DOMAIN} _;

  ssl_certificate         /etc/${SSL_TYPE}/live/${CERT_DOMAIN}/fullchain.pem;
  ssl_certificate_key     /etc/${SSL_TYPE}/live/${CERT_DOMAIN}/privkey.pem;
  ssl_trusted_certificate /etc/${SSL_TYPE}/live/${CERT_DOMAIN}/fullchain.pem;

  ssl_protocols TLSv1.2 TLSv1.3;
  ssl_prefer_server_ciphers off;
  ssl_session_tickets off;

  # 🔒 Hard stop: only serve the real host
  if ($host != '${CERT_DOMAIN}') { return 421; }
```

It doesn't

No.

### DIFF
--- a/files/nginx/setup-odk.sh
+++ b/files/nginx/setup-odk.sh
@@ -12,15 +12,16 @@ fi
   < /usr/share/odk/nginx/client-config.json.template \
   > /usr/share/nginx/html/client-config.json
 
-# Generate self-signed keys for the incorrect (catch-all) HTTPS listener.  This
-# cert should never be seen by legitimate users, so it's not a big deal that
-# it's self-signed and won't expire for 1,000 years.
+# Generate self-signed keys for the incorrect (catch-all) HTTPS listener. This
+# cert should never be seen by legitimate users, but it needs to be X.509 v3
+# have subjectAltName and be reasonably valid to work with SSL scanners.
 mkdir -p /etc/nginx/ssl
 openssl req -x509 -nodes -newkey rsa:2048 \
-    -subj "/" \
+    -subj "/CN=invalid.local" \
+    -addext "subjectAltName=DNS:invalid.local" \
     -keyout /etc/nginx/ssl/nginx.default.key \
     -out    /etc/nginx/ssl/nginx.default.crt \
-    -days 365000
+    -days 3650
 
 DH_PATH=/etc/dh/nginx.pem
 if [ "$SSL_TYPE" != "upstream" ] && [ ! -s "$DH_PATH" ]; then


matthew-white:
Since this PR changes code, I think we should merge into `next` rather than `master`. I just updated the target branch.

I'm also going to add this PR to the project just to keep track of it in the lead-up to releasing v2025.3.

files/nginx/setup-odk.sh
```diff
@@ -12,15 +12,16 @@ fi
...
+# have subjectAltName and be reasonably valid to work with SSL scanners.
 mkdir -p /etc/nginx/ssl
 openssl req -x509 -nodes -newkey rsa:2048 \
-    -subj "/" \
+    -subj "/CN=invalid.local" \
+    -addext "subjectAltName=DNS:invalid.local" \
     -keyout /etc/nginx/ssl/nginx.default.key \
     -out    /etc/nginx/ssl/nginx.default.crt \
-    -days 365000
+    -days 3650
```
alxndrsn: Is 10 years the max that the checkers will accept?  Will we deal with the failures of outdated certs in 10 years time?

yanokwa: I did not test if 10 years was the max. Agreed we should bump it up to give us as much time while still passing the checks.

yanokwa: Confirmed that 1000 years works, so I've updated that.

yanokwa:
> Is HTTPS usage without SNI worth supporting?

I don't know, but I don't think so?

files/nginx/setup-odk.sh
alxndrsn: How are you testing?  It would be great to be able to check from here too.

yanokwa: I added this PR to dev.getodk.cloud, then ran a new non-cached test on ssl labs. 
https://www.ssllabs.com/ssltest/analyze.html?d=dev.getodk.cloud&clearCache=on

Certificate 2: RSA 2048 bits (SHA256withRSA)
> Valid until: Tue, 15 Feb 3025 03:40:33 UTC (expires in 999 years and 3 months)

files/nginx/setup-odk.sh
```diff
@@ -12,12 +12,13 @@ fi
...
-# cert should never be seen by legitimate users, so it's not a big deal that
-# it's self-signed and won't expire for 1,000 years.
+# Generate self-signed keys for the incorrect (catch-all) HTTPS listener. This
+# cert should never be seen by legitimate users, but it needs to be X.509 v3
+# have subjectAltName and be reasonably valid to work with SSL scanners.
 mkdir -p /etc/nginx/ssl
 openssl req -x509 -nodes -newkey rsa:2048 \
-    -subj "/" \
+    -subj "/CN=invalid.local" \
+    -addext "subjectAltName=DNS:invalid.local" \
```
alxndrsn: ~I seem to get the same test result (A+) without the line:~  edit: incorrect - I'm having trouble getting the regenerated cert to show up

alxndrsn: ...now tested, and this line does not seem to have any effect on SSLLabs' test result.

yanokwa: Can confirm that only subj with `/CN=invalid.local` is needed. I've updated the PR.

alxndrsn (approved):

PR #1422, opened 2025-10-21
make dev: ensure running code is up-to-date

alxndrsn:
Currently, `make dev` will not reflect local changes to docker config, files etc.

# This change:

The official `docker compose` docs ref the `--build` flag are scant: https://docs.docker.com/reference/cli/docker/compose/up/

However, behaviour seems to be that if `--build` is supplied, then containers with a `build:` section defined in any of the compose `yaml` files will be rebuilt.

This has the effect that changes in any of the following could trigger a rebuild:

* `files/`
* the relevant `.dockerfile`
* the compose `.yaml` file(s)

If no changes are detected, this "rebuild" seems to be just a fetch from the local cache.

Tested this locally while trying to use `make dev` to speed up dev env setup.

It seems intuitive that `make dev` would serve recent changes to docker compose container contents and configuration!

No change - dev only.

No change - dev only.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ docker-compose-dev := docker compose --profile central -f docker-compose.yml -f 
 
 .PHONY: dev
 dev:
-	$(docker-compose-dev) up -d
+	$(docker-compose-dev) up --detach --build
 
 .PHONY: stop
 stop:


sadiqkhoja (approved):

baibongbaibong50-ops:
alxndrsn:dev-rebuild-on-start

Issue #934, opened 2025-03-26
Add ability to edit submission using Web Forms


matthew-white:
This is a follow-up issue to #883.

If I have enabled ODK Web Forms for a form, I should be able to use Web Forms to edit submissions to the form.

srujner:
**Tested With Success!**

Issue #883, opened 2025-01-31
Integrate latest version of ODK Web Forms with Submission capabilities


sadiqkhoja:
[Figma file here](https://www.figma.com/design/ULpCuTKkIys6Y99YGyMcrF/Central?node-id=103-1880&t=rRu9HDziCOC2Bzrj-4)

- When a user lands on the Form Settings page, there is a new section: Web forms
  - Show BETA pill 
  - Use the specific form name in the copy
- User is defaulted to Enketo
  - If the user clicks on the ODK Web forms ration button, show the Web forms modal with CTA to confirm action
  - When the modal is closed, show ODK Web forms as the selected radio button
- If the user selects Enketo radio button to change back, show the Enketo confirmation modal 

## TODO
- [ ] H/N: finalize modal text https://getodk.slack.com/archives/C01TKJ7TG7J/p1741627886395949?thread_ts=1741627025.581929&cid=C01TKJ7TG7J

nicoleorlowski:
[Figma file here](https://www.figma.com/design/ULpCuTKkIys6Y99YGyMcrF/Central?node-id=103-1880&t=rRu9HDziCOC2Bzrj-4)

nicoleorlowski:
Designs and release criteria added to description

sadiqkhoja:
@getodk/testers this is ready for testing.

Few things to note:

- Confirmation dialog box doesn't show screenshot of Web Forms as shown in Figma, it will come after #801 
- Do not test draft submission and draft preview, that should be done after #932

srujner:
@sadiqkhoja I see that there is follow-up issue #934 but I need to make sure that at the moment editing a submission ends up creating a new submission (with new ID) and at this point this is the expected behavior?

nicoleorlowski:
Could this be a success toast message?

lognaturel:
The currently-planned dialog text is 

> Form successfully sent!

> You can fill this form out again, or close if you're done.
> Close Fill out again

https://www.figma.com/design/ULpCuTKkIys6Y99YGyMcrF/Central?node-id=468-978&t=4jigZPoCIoHPfVES-0

I think that makes the two paths really clear. A toast would makes the "fill out again" case the default and our experience with Enketo is that it can be confusing to know where we are and what the form status is at that point (has it been cleared? Is there any relationship with the form I just sent?)

srujner:
**Tested with Success!**

nicoleorlowski:
[...]

PR #1467, opened 2025-11-01
nginx: configure CSP for web-forms

### LINKED ISSUE #1130: Add Content-Security-Policy for WebForms
WebForms have different behaviour and different paths to enketo.

# Example paths

Are webforms paths identifiable?

* https://staging.getodk.cloud/projects/158/forms/MarkdownExamples/preview?webforms=true
* https://staging.getodk.cloud/projects/158/forms/mediaTest/preview?webforms=true

# Example violations

Image

alxndrsn:
Closes #1130

Added new tests.

* A more permissive content security policy could be created, but this one seems to address all current violations.
* It's unfortunate to need a new `location` block, but there wasn't another clear way to achieve this as without it, nginx will rewrite the `$uri` value to `/index.html`, and then re-run the `map` which calculates `$frontend_csp_header`

The alternative to the extra `location` block might be just to make the general frontend security policy more permissive to account for Web Forms requirements.

They may see less network traffic, or less warnings from e.g. uBlock Origin.

There is risk of regression due to the weird way that nginx matchers work (or at least, how I managed to get them working).

Is there any risk to adding an extra env var to `docker-compose.yml`?  Presumably the `environment` section should not be edited.  Risk also seems low as there are already env vars duplicated across the `environment` sections for different services in `docker-compose.yml`.

No.

### DIFF
--- a/test/nginx/test-nginx.js
+++ b/test/nginx/test-nginx.js
@@ -485,7 +485,7 @@ describe('nginx config', () => {
 
         // then
         assert.equal(res.status, 200);
-        // and
+        assert.equal(await res.text(), '<div id="app"></div>\n');
         assertSecurityHeaders(res, { csp:'central-frontend' });
       });
     });


matthew-white:
> The only way I know to differentiate between webforms and other frontend requests is a combination of the path (must end with `/preview`, and the query parameters (`webforms` must be set to `true`)

As of v2025.1, Web Forms can be used not just for form preview, but also to create submissions (#883) and to edit them (#934). So there are additional paths to account for, for example, .../submissions/new. I'm pretty sure that `?webforms=true` is only relevant to form preview, not submission creation or editing.

Note that a form can be configured to use either Web Forms or Enketo. That means that you can't tell from the path alone whether a page will render Web Forms or Enketo. .../preview can be either Web Forms or Enketo; same for .../submissions/new. The `webforms` query parameter is entirely optional: if it's not `true`, .../preview will fall back to the form setting. Note that Enketo is rendered in an iframe, while Web Forms is not.

Given that Enketo is rendered in an iframe, maybe I'm making too much of the distinction between Web Forms and Enketo. Maybe for this PR, the only thing that matters is "are you on a path that may render Web Forms?" vs. "are you elsewhere in Frontend?" So maybe the list of paths just needs to be expanded beyond .../preview.

@sadiqkhoja was the one who implemented this in .1, so he has the best understanding of the different pathways in this area. I seem to remember that he created a Google doc summarizing those pathways.

sadiqkhoja:
Here is the document that list all possible paths: https://docs.google.com/document/d/19DYSIYXUY8ouizn1MAKz0j_M_17IpVFQXXwIwHzBco4/edit?tab=t.0#heading=h.dg7ce09owlla (requires auth)

files/nginx/odk.conf.template
```diff
@@ -166,6 +166,16 @@ server {
     proxy_read_timeout 2m;
   }
 
+  # Custom handling for form previews.
+  # Added to allow webforms CSP header to be set correctly.
+  location ~^(/f/)|(/projects/\d+/forms/[^/]*/(((draft/)?(preview|submissions/new))|(submissions/[^/]*/edit))$) {
```
matthew-white: This looks pretty complete to me. There is also .../submissions/new/offline to account for though.

@sadiqkhoja, would you mind reviewing this regular expression and checking that it's complete?

alxndrsn: > There is also .../submissions/new/offline to account for though.

Added.

matthew-white:
@lognaturel is going to take a look at this PR!

files/nginx/odk.conf.template
```diff
@@ -165,6 +165,16 @@ server {
     proxy_read_timeout 2m;
   }
 
+  # Custom handling for form previews.
```
lognaturel: replace "form previews" with "Web Forms"? Or maybe remove this line since the next one stands alone.

files/nginx/odk.conf.template
```diff
@@ -165,6 +165,16 @@ server {
...
+  # Added to allow webforms CSP header to be set correctly.
+  location ~^(/f/)|(/projects/\d+/forms/[^/]*/(((draft/)?(preview|submissions/new(/offline)?))|(submissions/[^/]*/edit))$) {
+    alias /usr/share/nginx/html/index.html;
+
+    add_header Content-Security-Policy-Report-Only "default-src 'none'; connect-src 'self' https://translate.google.com https://translate.googleapis.com ${S3_SERVER}; font-src 'self' data:; frame-src 'self'; img-src * data:; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-hashes'; style-src-attr 'unsafe-inline'; report-uri /csp-report";
+
```
lognaturel, line 177: Edit: found a comment that explains why, let’s keep it 

I'd tend to remove the directives that fall back to the default (`manifest-src`, `media-src`, `object-src`) so the focus on the meaningful ones is clear. There are already a few others like `child-src`, `worker-src` that are not explicitly listed and I like symmetry with that. Don't feel very strongly though!

lognaturel: Edit: confirmed blob is needed. 

`img-src` needs `blob:` as well. I think `*` is specifically for map tiles, everything else would be a blob or from self, I believe. We could probably be more narrow there.

lognaturel: Edit: confirmed unsafe-hashes won’t work for us. Also found that there are style elements inline. 

My understanding of the `unsafe-hashes` value for `style-src` is that the hashes of all allowed inline styles have to be included in the CSP policy. I think that can be removed and the `style-src-attr 'unsafe-inline'` is enough for our needs.

lognaturel: Edit: looks like `wasm-unsafe-eval` is enough for modern browsers 

Check: is `wasm-unsafe-eval` enough or is `unsafe-eval` needed?

lognaturel (changes requested):
I've done a first pass reading through, will also try the changes I've suggested and make sure they work.

files/nginx/odk.conf.template
lognaturel, line 152: I see this comment is helpful and addresses  https://github.com/getodk/central/pull/1467/files#r2582684607. Is there any reason `frame-src`, `manifest-src`, `object-src` were included but others like `worker-src` were excluded?

I imagine `worker-src 'self'` is needed here for offline mode.

files/nginx/odk.conf.template
```diff
@@ -165,6 +165,16 @@ server {
     proxy_read_timeout 2m;
   }
 
+  # Custom handling for form previews.
+  # Added to allow webforms CSP header to be set correctly.
+  location ~^(/f/)|(/projects/\d+/forms/[^/]*/(((draft/)?(preview|submissions/new(/offline)?))|(submissions/[^/]*/edit))$) {
+    alias /usr/share/nginx/html/index.html;
```
lognaturel, line 174: Was this maybe for testing?

lognaturel: My understanding is that we can't have a `location` block with a regex just to add a header. We either need to also need to have a content handler because regex location matches win.

files/nginx/odk.conf.template
alxndrsn: @sadiqkhoja were you happy with this?

files/nginx/odk.conf.template
```diff
@@ -169,6 +169,15 @@ server {
     include /usr/share/odk/nginx/backend.conf;
   }
 
+  # Allow webforms CSP header to be set correctly.
+  location ~^(/f/)|(/projects/\d+/forms/[^/]*/(((draft/)?(preview|submissions/new(/offline)?))|(submissions/[^/]*/edit))$) {
```
sadiqkhoja, line 173:
```suggestion
  location ~^(/f/)|(/projects/\d+/forms/[^/]+/(((draft/)?(preview|submissions/new(/offline)?))|(submissions/[^/]+/edit))$) {
```

files/nginx/odk.conf.template
sadiqkhoja: yes, added a minor suggestion.

files/nginx/odk.conf.template
alxndrsn: :+1: it looks like this change was incorporated into https://github.com/getodk/central/pull/1526/files#diff-63e29b9f739885a03903f2c29d84eac681d12ad34411a41351ae202410646d7fR83

alxndrsn:
Superseded by https://github.com/getodk/central/pull/1526.

PR #1468, opened 2025-11-02
nginx: update CSP for frontend maps

### LINKED ISSUE #1403: Map violates Content Security Policy
When I view a map in Central Frontend, I see the following message in the console:

> [Report Only] Refused to create a worker from 'blob:foo' because it violates the following Content Security Policy directive: "script-src 'self'". Note that 'worker-src' was not explicitly set, so 'script-src' is used as a fallback.

Example URL: https://staging.getodk.cloud/projects/182/forms/building_footprints/submissions?map=true

@alxndrsn, I'll add this issue to #770.

alxndrsn:
Closes #1403

Fresh tests.

Simple.  Open to alternatives, but couldn't think of any!

* should reduce browser console noise
* may override current `script-src`, in which case `'self'` will need to be added.  This will be noticed in testing, or in prod (where it would replace the current CSP violation with a new one, so not a big deal)

No.

### DIFF
--- a/docs/news.html
+++ b/docs/news.html
@@ -7,15 +7,15 @@
   </head>
   <body>
     <div class="news-item">
-      <span class="item-date">2025/11/03</span>
-      <a href="https://forum.getodk.org/t/57029" target="_blank">
-        ODK Central v2025.3.0
+      <span class="item-date">2025/11/13</span>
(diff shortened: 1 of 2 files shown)


matthew-white (approved):
This one is a small change, so I think I can go ahead and approve. :+1:

PR #1478, opened 2025-11-05
csp: allow backend passthrough for /oidc/callback

### LINKED ISSUE #1235: oidc: csp: style-src-elem directive violated on login
> when I log in on dev, a request is sent to /csp-report. The `violated-directive` is `style-src-elem`.
> - _https://github.com/getodk/central-backend/pull/910#issuecomment-1696768080_

alxndrsn:
With https://github.com/getodk/central-backend/pull/1679, closes #1235

Updated tests.

This allows style content-hash in CSP set in https://github.com/getodk/central-backend/pull/1679 to pass through nginx.

Should reduce CSP violation reports; most users won't notice they were happening in the first place.

No.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -91,6 +91,9 @@ server {
 
   server_tokens off;
 
+  add_header Content-Security-Policy-Report-Only "default-src 'none'; connect-src https://translate.google.com https://translate.googleapis.com; img-src https://translate.google.com; report-uri /csp-report";
+  include /usr/share/odk/nginx/common-headers.conf;
+
   client_max_body_size 100m;
 
   gzip on;


test/nginx/test-nginx.js
```diff
@@ -6,7 +6,7 @@ const none = `'none'`;
 const self = `'self'`;
 const unsafeInline = `'unsafe-inline'`;
 const contentSecurityPolicies = {
-  'restrictive': {
+  'backend-default': {
```
alxndrsn, line 10: Renamed this to:

* be more accurate, and
* reflect association with new ruleset `backend-unmodified`

test/nginx/test-nginx.js
```diff
@@ -36,6 +39,9 @@ const contentSecurityPolicies = {
     'style-src-attr': unsafeInline,
     'report-uri':     '/csp-report',
   },
+  'disallow-all': {
+    'default-src': none,
+  },
```
alxndrsn, line 46: Newly-introduced to better reflect what `blank.html` should receive.  Happy to separate as a prior PR if helpful.

alxndrsn: PR at https://github.com/getodk/central/pull/1479

test/nginx/test-nginx.js
alxndrsn: Also renamed in https://github.com/getodk/central/pull/1479?

files/nginx/odk.conf.template
```diff
@@ -91,9 +91,6 @@ server {
 
   server_tokens off;
 
-  add_header Content-Security-Policy-Report-Only "default-src 'none'; connect-src https://translate.google.com https://translate.googleapis.com; img-src https://translate.google.com; report-uri /csp-report";
-  include /usr/share/odk/nginx/common-headers.conf;
```
matthew-white, line 95: Is it definitely safe to remove this `include`? Do all `location` blocks that need it already individually `include` it?

Similarly for the `Content-Security-Policy-Report-Only` header, is the reason for removing it here because it's already been added to all the individual `location` blocks that need it?

matthew-white (approved):

files/nginx/odk.conf.template
alxndrsn: > Is it definitely safe to remove this `include`?

Good question!

> Do all `location` blocks that need it already individually include it?

I think so, as do the current tests.  But that's not a guarantee!

The most conservative approach might be to lock down all headers here as much as possible.  Should we try that?  That could definitely catch some cases, but still won't protect from `add_header` directives in lower-level blocks from blowing away _all_ headers set here.

matthew-white: I actually kind of prefer removing these lines if possible. I feel like it's easier to reason about the headers if they're all on one level. But I thought I'd check about it first.

I did a scan of the `location` blocks in this PR, and there seem to be three categories:

1. `location` blocks that individually `include` common-headers.conf and add the `Content-Security-Policy-Report-Only` header (including `location` blocks that `try_files ... @blank.html;`).
2. `location` blocks that `return 301`
    - These don't need any of the normal response headers, right?
3. /csp-report
    - This is a specific use case, so maybe it doesn't need any of the normal response headers. Users shouldn't be navigating to this URL directly. What do you think?
    - This is also a proxy to Sentry.

matthew-white: I'm also open to locking down all headers here if you think that's better. :+1: Or at least locking down `Content-Security-Policy-Report-Only`. I think my first choice would be to remove these lines if possible, but it's not a strong preference (I'm happy either way).

alxndrsn: > seem to be three categories

:+1: looks accurate to me

> These don't need any of the normal response headers, right?

I think this is a slightly complicated question.  I'll reinstate these lines, and merge the PR.  We can try to delete or refine these default headers down the line.

files/nginx/odk.conf.template
```diff
@@ -154,15 +155,17 @@ server {
...
+    include /usr/share/odk/nginx/common-headers.conf;
+    include /usr/share/odk/nginx/backend.conf;
+  }
+
   location ~ ^/v\d {
-    proxy_set_header X-Forwarded-Proto $scheme;
-    proxy_pass http://service:8383;
-    proxy_redirect off;
+    proxy_hide_header Content-Security-Policy-Report-Only;
+    add_header Content-Security-Policy-Report-Only "default-src 'none'; connect-src https://translate.google.com https://translate.googleapis.com; img-src https://translate.google.com; report-uri /csp-report";
```
matthew-white, line 165: This is kind of an aside/tangent, but why do we include these Google Translate URLs for the Backend CSP? Backend almost never returns HTML, so I feel like the Google Translate bit isn't needed. Google Translate isn't going to translate JSON returned from Backend. I know this has been the default value of this header, but I wonder whether it could be further simplified for Backend.

Or is the attitude more like, "it never hurts to include Google Translate, so let's just always include it"?

alxndrsn: I think this is a mistake - API services should not be loaded in the browser as the main document, so it should be safe to completely lock down their CSPs.  On the other hand, the `blank.html` file **may** be loaded as the main document, so should probably be allowing google translate.

alxndrsn: Filed at:

* https://github.com/getodk/central/issues/1516
* https://github.com/getodk/central/issues/1517

PR #1479, opened 2025-11-05
nginx/csp: make blank.html CSP completely restrictive

alxndrsn:
Noted while working on https://github.com/getodk/central/issues/770

Updated tests.

Not really.

Low risk, low impact.

No.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -172,7 +172,6 @@ server {
     add_header Content-Security-Policy-Report-Only "default-src 'none'";
     include /usr/share/odk/nginx/common-headers.conf;
   }
-
   location = /blank.html {
     try_files $uri @blank.html;
   }


files/nginx/odk.conf.template
matthew-white, line 1: We also load blank.html via `try_files $uri $uri/ /index.html;` in `location / {` below. The `SubmissionDownload` modal in Frontend sometimes renders blank.html in an iframe.

Given that, I'm wondering whether we need an addition `location` block (`location /blank.html {`) to ensure that the restrictive CSP in this PR is applied when /blank.html is loaded within `SubmissionDownload`.

matthew-white:
I'm happy to review this PR rather than @brontolosone, but I'm still wondering about my question above. Maybe that question doesn't need to block merge though. We could address the case of /blank.html being loaded within the modal iframe in a separate PR.

files/nginx/odk.conf.template
alxndrsn: 👍 done

files/nginx/odk.conf.template
```diff
@@ -130,8 +130,7 @@ server {
   }
   # To read single submission cookies
   location = /-/single/check-submitted {
-    alias /usr/share/nginx/html/blank.html;
-    default_type text/html;
+    try_files $uri @blank.html;
```
matthew-white, line 133: Is it possible to remove `$uri` somehow? I don't think so, but I wanted to ask. It feels a little funny knowing that nginx will be looking for a file at file path `$uri`, which we know doesn't exist.

From the docs, it looks like `try_files` is expecting at least two arguments. It seems like only the last argument can be a named location. So I don't see an obvious alternative to what's here.

alxndrsn: It looks like it should be possible; let's test and find out :rocket:

alxndrsn: Not possible:

> invalid number of arguments in "try_files" directive in /etc/nginx/conf.d/odk.conf:133

PR #1508, opened 2025-12-02
ci: run gixy to detect nginx config issues

alxndrsn:
Implemented coincidentally while trying to answer the query at https://github.com/getodk/central/pull/1478/files/30b1cf420e7ea4cd74ed63b4347be6929ab37fef#r2579682481.  I was looking for a static analysis tool which might report on paths not covered by existing location directives.  No luck there, but this PR came out of it easily.

#### What has been done to verify that this works as intended?

New test, initially failing, then passing when changes to config are made.

#### Why is this the best possible solution? Were any other approaches considered?

* `gixy-ng` seems to be a maintained, popular fork of `gixy`
  * `gixy-ng`: https://github.com/dvershinin/gixy/
  * original `gixy` repo: https://github.com/yandex/gixy/
* currently only considering HIGH-level reports
  * there are some HIGH reports that can be fixed without much effort or controversy
  * there are some MEDIUM reports which are more complicated; this PR ignores those

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

No change.

#### Does this change require updates to documentation? If so, please file an issue [here](https://github.com/getodk/docs/issues/new) and include the link below.

No.

#### Before submitting this PR, please make sure you have:

- [x] branched off and targeted the `next` branch OR only changed documentation/infrastructure (`master` is stable and used in production)
- [x] verified that any code or assets from external sources are properly credited in comments or that everything is internally sourced

PR #1509, opened 2025-12-02
nginx/redirector: add warning re editing the file

alxndrsn:
Add a warning to `redirector.conf` because the file is _sometimes_ modified in container startup.  The modification is quite naive, and also only occurs in some situations, so it would be easy to break that code without realising.

This PR will also serve as a helpful reference for how to handle changes to `redirector.conf` with corresponding changes in `setup-odk.sh`.

CI actually catches (some) bad modification right now!

A better approach might be to make the modification code more robust, e.g. using a patchfile with large line context.  However, this PR seems like a good step to improve safety with very low effort.

No change.

No.

### DIFF
--- a/files/nginx/redirector.conf
+++ b/files/nginx/redirector.conf
@@ -1,4 +1,4 @@
-# Be VERY careful modifying this file - it is edited BY LINE NUMBER in setup-odk.sh
+# Be VERY careful modifying this file - it is modified BY LINE NUMBER in setup-odk.sh
 server {
     # Listen on plain old HTTP and catch all requests so they can be redirected
     # to HTTPS instead.


files/nginx/redirector.conf
```diff
@@ -1,3 +1,4 @@
+# Be VERY careful modifying this file - it is edited BY LINE NUMBER in setup-odk.sh
```
alxndrsn:
```suggestion
# Be VERY careful modifying this file - it is modified BY LINE NUMBER in setup-odk.sh
```

files/nginx/setup-odk.sh
```diff
@@ -63,7 +63,7 @@ else
     echo "starting nginx for upstream ssl..."
   else
     # remove letsencrypt challenge reply, but keep 80 to 443 redirection
-    perl -i -ne 'print if $. < 7 || $. > 14' /etc/nginx/conf.d/redirector.conf
+    perl -i -ne 'print if $. < 8 || $. > 15' /etc/nginx/conf.d/redirector.conf
```
matthew-white, line 66: I guess this will need to become 9 and 16 once #1508 is merged, since it too adds a line to redirector.conf.

alxndrsn: :+1: thanks for the reminder.

I merged this PR first, and once updated #1508 is failing CI with error:

```
...
writing fresh nginx templates...
starting nginx for custom ssl and self-signed certs...
2025/12/03 07:54:43 [emerg] 1#1: "location" directive is not allowed here in /etc/nginx/conf.d/redirector.conf:12
nginx: [emerg] "location" directive is not allowed here in /etc/nginx/conf.d/redirector.conf:12
```

See: https://github.com/getodk/central/actions/runs/19886477705/job/56994705734?pr=1508

So _currently_ the build will detect changes to `redirector.conf` without corresponding updates to `setup-odk.sh` iff they result in broken nginx config.

PR #1526, opened 2025-12-04
Add CSP headers specifically for Web Forms

### LINKED ISSUE #1130: Add Content-Security-Policy for WebForms

lognaturel:
Closes #1130
Alternative to https://github.com/getodk/central/pull/1467/files

Put on dev server, loaded All Question Types form as new submission, Public Link, and for submission edit. Looked in console for any violations.

I did NOT review reported violations extensively.

The most obvious alternative is to have the same policy for all of frontend. I don't think that would be so bad but the increased granularity is nice in some cases. The regex approach to matching routes feels a little unsatisfying/brittle but the stakes are low if there's a mistake because the policies are not very different.

One challenge for this behavior is where to define the Web Forms-specific headers which apply to several different routes. This PR uses a map to go from the request URI to the intended CSP headers. That feels like a clean, easy-to-read solution to me. I mostly took the regex from #1467 and trusted that the tests are complete enough.

I tried mapping on `$uri` to allow query parameters but realized that gets rewritten so mapped on `$request_uri` and explicitly allowed query params in the regex.

The directives are intended to be in alphabetical order with `default-src: 'none'` listed first for readability.

Two fetch directives are omitted: `child-src` and `prefetch-src`. I think that's ok because they're not likely to be the cause of violations. The more specific directives like `script-elem-src` are also omitted and I think that's ok too.

TODO:
- [x] remove 'self' from frame-src for Web Forms

I also reviewed each existing directive:
| Directive | Default | Web Forms | Explanation |
|---|---|---|---|
| `default-src` | `none` | `none`| Want restrictive baseline |
| `connect-src` | `'self' Google translate` | `'self' https:` | Web Forms may need to show images hosted on some arbitrary blob storage server. The blob storage URLs may not exactly match the domain specified in the config.  |
| `font-src` | `'self'` | `'self' data:` | `data:` is because Roboto gets downloaded (note: seeing it on test. but not dev.) |
| `frame-src` | `'self' https://getodk.github.io/central/news.html` | `'none'` | At some point we'll need to figure out how we want to allow Web Forms to be used in frames but we don't need to do that now |
| `img-src` | `https: data:` | `https: blob:` | I did see a `data:` image request from frontend but ⚠️ couldn't figure out what it is; `https:` for frontend is because users can link to arbitrary images in markdown. `https:` for Web Forms is currently for map tiles. We could pretty easily be more restrictive here I think. |
| `manifest-src` | `'none'` | `'none'` | Listed just for reports |
| `media-src` | `'none'` | `'none'` | Listed just for reports |
| `object-src` | `'none'` | `'none'` | Listed just for reports |
| `script-src` | `'self'` | `'self' 'wasm-unsafe-eval'` | WASM used by XPath evaluator |
| `style-src` | `'self'` | `'self' 'unsafe-inline'` | PrimeVue uses inline styles for at least `DatePicker` |
| `style-src-attr` | `'unsafe-inline'` | (not set) | ⚠️ review why this is needed for frontend |
| `worker-src` | `blob:` | `blob:` | Used by OpenLayers |

The most important intended behavior change is that we start getting more meaningful / less noisy reports and that users don't see warnings in their consoles.

No immediate regression risk because we're still doing report only. We'd like to turn on actual protection as soon as possible, though, so if we made mistakes we run the risk of further delaying that.

No

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -81,7 +81,7 @@ map $arg_st $redirect_single_prefix {
 map $request_uri $central_frontend_csp {
   # Web Forms CSP for /f/... and /projects/.../forms/... routes
   ~^/(?:f/[^/]+(?:/.*)?|projects/\d+/forms/[^/]+/(?:(?:draft/)?(?:preview|submissions/new(?:/offline)?)|submissions/[^/]+/edit)(?:/)?)(?:\?.*)?$
-    "default-src 'none'; connect-src 'self' https:; font-src 'self' data:; frame-src 'self'; img-src blob: https:; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; worker-src blob:; report-uri /csp-report";
+    "default-src 'none'; connect-src 'self' https:; font-src 'self' data:; frame-src 'none'; img-src blob: https:; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; worker-src blob:; report-uri /csp-report";
 
   default
     "default-src 'none'; connect-src 'self' https://translate.google.com https://translate.googleapis.com; font-src 'self'; frame-src 'self' https://getodk.github.io/central/news.html; img-src data: https:; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'self'; style-src 'self'; style-src-attr 'unsafe-inline'; worker-src blob:; report-uri /csp-report";
--- a/test/nginx/test-nginx.js
+++ b/test/nginx/test-nginx.js
@@ -125,7 +125,7 @@ const contentSecurityPolicies = {
       self,
       'data:',
     ],
-    'frame-src': self,
+    'frame-src': none,
     'img-src': [
       'blob:',
       'https:',


test/nginx/test-nginx.js
```diff
@@ -18,15 +19,15 @@ const asArray = val => {
...
+    ? asArray(connectSrc)
+    : [...asArray(connectSrc),
     'https://translate.google.com',
     'https://translate.googleapis.com',
   ],
-  'img-src': [
-    ...asArray(imgSrc),
-    'https://translate.google.com',
-  ],
+  'img-src': (imgSrc ?? '').includes('https:')
```
alxndrsn:
```suggestion
  'img-src': imgSrc?.includes('https:')
```

test/nginx/test-nginx.js
```diff
@@ -18,15 +19,15 @@ const asArray = val => {
 };
 const allowGoogleTranslate = ({ 'connect-src':connectSrc, 'img-src':imgSrc, ...others }) => ({
   ...others,
-  'connect-src': [
-    ...asArray(connectSrc),
+  'connect-src': (connectSrc ?? '').includes('https:')
```
alxndrsn:
```suggestion
  'connect-src': connectSrc?.includes('https:')
```

test/nginx/test-nginx.js
```diff
@@ -109,6 +110,35 @@ const contentSecurityPolicies = {
     'style-src-attr': unsafeInline,
     'report-uri': '/csp-report',
   }),
+
```
alxndrsn:
```suggestion
```

files/nginx/odk.conf.template
```diff
@@ -75,6 +75,16 @@ map $arg_st $redirect_single_prefix {
...
 }
 
+# Use 'none' per directive instead of falling back to default-src to make CSP violation reports more specific
+map $request_uri $central_frontend_csp {
+  # Web Forms CSP for /f/... and /projects/.../forms/... routes
+  ~^/(?:f/[^/]+(?:/.*)?|projects/\d+/forms/[^/]+/(?:(?:draft/)?(?:preview|submissions/new(?:/offline)?)|submissions/[^/]+/edit))(?:\?.*)?$
+    "default-src 'none'; connect-src 'self' https:; font-src 'self' data:; frame-src 'self'; img-src https: blob:; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; worker-src blob:; report-uri /csp-report";
+
+  default
+    "default-src 'none'; connect-src 'self' https://translate.google.com https://translate.googleapis.com; font-src 'self'; frame-src 'self' https://getodk.github.io/central/news.html; img-src https: data:; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'self'; style-src 'self'; style-src-attr 'unsafe-inline'; worker-src blob:; report-uri /csp-report";
```
alxndrsn: This has changed as well as moved.

files/nginx/odk.conf.template
```diff
@@ -146,11 +156,9 @@ server {
...
     proxy_hide_header Content-Security-Policy-Report-Only;
     add_header Content-Security-Policy-Report-Only "default-src 'none'; connect-src 'self' blob: https://maps.googleapis.com/ https://maps.google.com/ https://maps.gstatic.com/mapfiles/ https://fonts.gstatic.com/ https://fonts.googleapis.com/ https://translate.google.com https://translate.googleapis.com; font-src 'self' https://fonts.gstatic.com/; frame-src 'none'; img-src data: blob: jr: 'self' https://maps.google.com/maps/ https://maps.gstatic.com/mapfiles/ https://maps.googleapis.com/maps/ https://tile.openstreetmap.org/ https://translate.google.com; manifest-src 'none'; media-src blob: jr: 'self'; object-src 'none'; script-src 'unsafe-inline' 'self' https://maps.googleapis.com/maps/api/js/ https://maps.google.com/maps/ https://maps.google.com/maps-api-v3/api/js/; style-src 'unsafe-inline' 'self' https://fonts.googleapis.com/css; style-src-attr 'unsafe-inline'; report-uri /csp-report";
-    #
-    # Rules set to 'none' here would fallback to default-src if excluded.
-    # They are included here to ease interpretation of violation reports.
```
alxndrsn: Comment moved up and clarified; intention is the same.

files/nginx/odk.conf.template
```diff
@@ -75,6 +75,16 @@ map $arg_st $redirect_single_prefix {
   default "/new${is_args}${args}${qp_deliminator}single=true";
 }
 
+# Use 'none' per directive instead of falling back to default-src to make CSP violation reports more specific
+map $request_uri $central_frontend_csp {
```
alxndrsn, line 81: 1. is it safe to use `$request_uri` here, rather than `$uri`?
2. why not just use `$uri`, and then no matcher is needed below for presence/absence of querystring

files/nginx/odk.conf.template
alxndrsn, line 11: Ref comment below, this should probably just be `$uri`.  File separately.

test/nginx/test-nginx.js
```diff
@@ -43,14 +44,14 @@ const contentSecurityPolicies = {
       self,
       'https://getodk.github.io/central/news.html',
     ],
-    'img-src':        '* data:',
+    'img-src':        'https: data:',
```
alxndrsn: I think it would be helpful for commit history to track these changes separately from Web Forms-related changes.  Especially given the difficulty of understanding historic CSP changes when reports are no longer available.

test/nginx/test-nginx.js
```diff
@@ -43,14 +44,14 @@ const contentSecurityPolicies = {
...
     'manifest-src':   none,
     'media-src':      none,
     'object-src':     none,
     'script-src':     self,
     'style-src':      self,
     'style-src-attr': unsafeInline,
-    'worker-src':     'data:',
+    'worker-src':     'blob:',
```
alxndrsn, line 59: This is a specific bugfix in the existing frontend scheme - the worker-src here is to support OpenLayers maps, which require `blob:` and not `data:`.

alxndrsn: Separate PR, for the same reasoning as above:

>  it would be helpful for commit history to track these changes separately from Web Forms-related changes. Especially given the difficulty of understanding historic CSP changes when reports are no longer available.

test/nginx/test-nginx.js
```diff
@@ -18,15 +19,15 @@ const asArray = val => {
...
     'https://translate.google.com',
     'https://translate.googleapis.com',
   ],
-  'img-src': [
-    ...asArray(imgSrc),
-    'https://translate.google.com',
-  ],
+  'img-src': imgSrc?.includes('https:')
+    ? asArray(imgSrc)
+    : [...asArray(imgSrc), 'https://translate.google.com'],
```
alxndrsn: Struggling to read this, and its implications for test expectations.  I suspect it will confuse future readers in its current form.

test/nginx/test-nginx.js
```diff
@@ -18,15 +19,15 @@ const asArray = val => {
...
+    ? asArray(connectSrc)
+    : [...asArray(connectSrc),
     'https://translate.google.com',
     'https://translate.googleapis.com',
   ],
-  'img-src': [
-    ...asArray(imgSrc),
-    'https://translate.google.com',
-  ],
+  'img-src': imgSrc?.includes('https:')
```
alxndrsn: `includes` isn't a safe way to test this, as an `img-src` of e.g. `https://s3.amazon.com/my-special-bucket` would pass this check and cause google translate-specific URLs from being included.

alxndrsn: The intention is to check explicitly for the `https:` directive.

files/nginx/odk.conf.template
lognaturel: Addressed in PR description — uri is index.html

test/nginx/test-nginx.js
lognaturel: My thinking was that the table in the PR description would provide a good overview for the future. But if you prefer separate PRs, please make the change and update the table to match.

test/nginx/test-nginx.js
lognaturel: Ah yes, I had an array in mind, darn.

test/nginx/test-nginx.js
alxndrsn: Filed at https://github.com/getodk/central/pull/1530

alxndrsn, line 59:
```suggestion
    'worker-src':     'data:',
```

files/nginx/odk.conf.template
```diff
@@ -75,6 +75,16 @@ map $arg_st $redirect_single_prefix {
...
+  # Web Forms CSP for /f/... and /projects/.../forms/... routes
+  ~^/(?:f/[^/]+(?:/.*)?|projects/\d+/forms/[^/]+/(?:(?:draft/)?(?:preview|submissions/new(?:/offline)?)|submissions/[^/]+/edit))(?:\?.*)?$
```
alxndrsn: Can be bypassed e.g.

* ✅ https://dev.getodk.cloud/projects/11/forms/all-question-types/submissions/new
* ❌ https://dev.getodk.cloud/projects/11/forms/all-question-types/submissions/new/

Add a new test case?

test/nginx/test-nginx.js
alxndrsn: Separate PR at https://github.com/getodk/central/pull/1531

test/nginx/test-nginx.js
alxndrsn: Would be resolved by approach in https://github.com/getodk/central/pull/1531

files/nginx/odk.conf.template
alxndrsn: May be required to make fonts immutable by version.  Might have implications for query strings on patterns matched below with final `$`.

files/nginx/odk.conf.template
alxndrsn: This doesn't _seem_ safe, because percent-encoding parts of the request URL will bypass these checks due to `$request_uri` not being normalised.  However, in practice it's ok because of https://github.com/getodk/central/issues/1532.

files/nginx/odk.conf.template
alxndrsn: Added multiple new test cases.

test/nginx/test-nginx.js
alxndrsn: Would be resolved by approach in #1531?

test/nginx/test-nginx.js
```diff
@@ -411,6 +440,94 @@ describe('nginx config', () => {
...
+      '/projects/1/forms//preview',
+      '/projects/1/forms//preview/',
+
+      // missing submission ID
+      '/projects/1/forms/some_xml_form_id/submissions//edit',
+      '/projects/1/forms/some_xml_form_id/submissions//edit/',
+
+      // all /f/* should be valid
+      '/f',
+      '/f/',
```
alxndrsn, line 525: Fresh test cases.  These did not require any change to the regex.

test/nginx/test-nginx.js
```diff
@@ -411,6 +440,94 @@ describe('nginx config', () => {
...
+      '/projects/1-not-just-a-number-1/forms/some_xml_form_id/submissions/new',
+      '/projects/1-not-just-a-number-1/forms/some_xml_form_id/submissions/new/',
+      '/projects/1-not-just-a-number-1/forms/some_xml_form_id/submissions/00000000-0000-0000-0000-000000000000/edit',
+      '/projects/1-not-just-a-number-1/forms/some_xml_form_id/submissions/00000000-0000-0000-0000-000000000000/edit/',
+      '/projects/1-not-just-a-number-1/forms/some_xml_form_id/preview',
+      '/projects/1-not-just-a-number-1/forms/some_xml_form_id/preview/',
+      '/projects/1-not-just-a-number-1/forms/some_xml_form_id/draft/submissions/new',
+      '/projects/1-not-just-a-number-1/forms/some_xml_form_id/draft/submissions/new/',
+      '/projects/1-not-just-a-number-1/forms/some_xml_form_id/draft/preview',
+      '/projects/1-not-just-a-number-1/forms/some_xml_form_id/draft/preview/',
```
alxndrsn, line 509: Loads (too many?!) fresh test cases for trailing slashes.

test/nginx/test-nginx.js
alxndrsn: Reverted the change here.

test/nginx/test-nginx.js
alxndrsn: Imported approach from #1531

test/nginx/test-nginx.js
alxndrsn: Imported approach from #1531

files/nginx/odk.conf.template
```diff
@@ -75,6 +75,16 @@ map $arg_st $redirect_single_prefix {
   default "/new${is_args}${args}${qp_deliminator}single=true";
 }
 
+# Use 'none' per directive instead of falling back to default-src to make CSP violation reports more specific
```
alxndrsn, line 78:
```suggestion
# Use 'none' per directive instead of falling back to default-src to make CSP violation reports more specific
# Note: using $request_uri here remains reliable while percent-encodings are not
# normalised in frontend URLs.  Tracked at https://github.com/getodk/central/issues/1532
```

files/nginx/odk.conf.template
alxndrsn: Added a comment to this effect.

files/nginx/odk.conf.template
alxndrsn: Change reverted.  Will cause conflict when introduced.

lognaturel:
I'm happy with where this is. It's my PR so I can't approve it. @alxndrsn please squash and merge when you're happy with it!

alxndrsn:
> TODO:
> 
> - [ ] remove 'self' from frame-src for Web Forms

@lognaturel does this need to be done?

I haven't merged due to this - please feel free to merge, either with or without `frame-src: 'self'` set.

test/nginx/test-nginx.js
```diff
@@ -114,6 +115,37 @@ const contentSecurityPolicies = {
...
+    'default-src': none,
+    'connect-src': [
+      self,
+      'https:',
+    ],
+    'font-src': [
+      self,
+      'data:',
+    ],
+    'frame-src': self,
```
alxndrsn: This is still in question!

lognaturel: Whoops, thought that had been done for some reason. I fixed it.

Issue #672, opened 2024-05-23
Check for large files in repos


matthew-white:
This issue is to prevent large files from being accidentally committed to a Central repo. One reason why that's important is that end users clone the `central` repo when installing Central, then pull from it when upgrading. We should check for large files in each of the three repos. Let's use a limit of 1 MB.

- [ ] `central`
- [x] `central-backend`
- [ ] `central-frontend`

Issue #1572, opened 2025-12-16
[backend] drop circle CI build?


alxndrsn:
`odk-central-backend` is tested in both GitHub Actions and CircleCI.

Almost all tests run in CircleCI are also run in the `standard-suite` workflow in GitHub Actions.  The only difference seems to be that CircleCI runs two extra tests:

1. "Check for large files"
2. `make check-file-headers`

# Benefits

## Maintenance

Removing support for CircleCI would simplify maintenance of CI jobs.  See e.g.:

* https://github.com/getodk/central/issues/1570
* https://github.com/getodk/central/issues/1571

## Points of failure

Currently CircleCI can block builds due to unreliability.  E.g. https://github.com/getodk/central-backend/pull/1719 has been waiting ~40 minutes for a CircleCI job to run:

<kbd>
<img width="372" alt="Image" src="https://github.com/user-attachments/assets/c7bd2972-6c71-4381-a05f-a5739e532ac5" />

PR #1574, opened 2025-12-16
ci: reject if unexpected large files are committed

alxndrsn:
Noted while looking into https://github.com/getodk/central/issues/1572

Ref https://github.com/getodk/central/issues/672

Run locally, run in frontend and backend repos.  Only fails in backend, which has a big `package-lock.json`.

An alternative approach would be that used in `central-backend`'s CircleCI job:

https://github.com/getodk/central-backend/blob/97981e9694f392e5b3df90a2d28c6f7d3aefef17/.circleci/config.yml#L16-L19

Advantages of this version:

* only checks committed files, so
  * much faster (<10ms vs 1+ second)
  * no false-positives - running the `central-backend` script locally fails for `node_modules` and generated test files
* hopefully more readable (even if `awk` may not be familiar)

No effect.

No.

#### Follow-up

If this is accepted, it could be copied to `odk-central-frontend` to close https://github.com/getodk/central/issues/672.

### DIFF
--- a/test/check-for-large-files.sh
+++ b/test/check-for-large-files.sh
@@ -3,7 +3,7 @@ set -o pipefail
 shopt -s inherit_errexit
 
 git ls-files -z \
-| xargs -0 ls -l \
+| xargs -0 ls -l -- \
 | awk '
   BEGIN {
     print "[check-for-large-files] Checking for large files...";


test/check-for-large-files.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash -eu
+set -o pipefail
+shopt -s inherit_errexit
+
+git ls-files -z \
+| xargs -0 ls -l \
```
matthew-white:
```suggestion
| xargs -0 ls -l -- \
```

How about adding `--` in case a filename begins with a hyphen? So that the filename isn't interpreted as an option. Seems unlikely, but it feels consistent with how `-z` and `-0` enable the script to handle all kinds of filenames.

PR #1580, opened 2025-12-18
Chores getodk/central#1555: Update Nodejs to v24

sadiqkhoja:
Closes https://github.com/getodk/central/issues/1555

Usually we do dependency updates closer to the release but since this is major upgrade, it is better to do it at the start of the release cycle so that we have longer runway to catch subtle issues/bugs.

Related: https://github.com/getodk/central-frontend/pull/1456 and https://github.com/getodk/central-backend/pull/1722/


matthew-white:
Let's wait to merge until the patch release is out, but otherwise this looks good to go!

PR #1586, opened 2025-12-27
Update X-Forwarded-Proto directive in backend.conf

jniles:
This PR fixes a regression in behavior reported on the ODK forum [here](https://forum.getodk.org/t/setting-up-central-behind-a-proxy-this-authentication-method-is-only-available-over-https/57236/10).  The regression was introduced in https://github.com/getodk/central/commit/924d320112b1e5d5929803a51f3bad7056d1295e during a code refactor.  The code refactor moved the `X-Forwarded-Proto` header to a new file but did not update the regexp which pins this header to `https` in `setup-odk.sh` script to point to that file.  The result is that HTTP authentication breaks for users hosting ODK behind a reverse proxy.

I have made the changes on my local installation, rerun the docker compose build, and confirmed that the issues reported in [this forum post](https://forum.getodk.org/t/setting-up-central-behind-a-proxy-this-authentication-method-is-only-available-over-https/57236/10) are resolved.  I can log into my server without throwing the `httpsOnly()` error.

I believe this is a minor regression introduced during code refactoring in error.  This emulates the behavior of the build script before https://github.com/getodk/central/commit/924d320112b1e5d5929803a51f3bad7056d1295e.

This fixes a regression in behavior, and should not introduce more breakages.

No.

### DIFF
--- a/files/nginx/setup-odk.sh
+++ b/files/nginx/setup-odk.sh
@@ -59,7 +59,7 @@ else
     # strip out all ssl_* directives
     perl -i -ne 's/listen 443.*/listen 80;/; print if ! /ssl_/' /etc/nginx/conf.d/odk.conf
     # force https because we expect SSL upstream
-    perl -i -pe 's/X-Forwarded-Proto \$scheme/X-Forwarded-Proto https/;' /etc/nginx/conf.d/odk.conf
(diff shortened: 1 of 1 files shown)


alxndrsn:
@jniles I've added tests relating to `SSL_TYPE=upstream` at https://github.com/getodk/central/pull/1589, and fixed another minor bug with that setup (relating to Sentry CSP reporting - it should not have been impacting your users).

PR #1599, opened 2026-01-12
ci/close-issues: support more issue ref notations

### LINKED ISSUE #1592: Merging PRs to `next` branch doesn't close associated Issues
Sometimes?  Always?

# Examples

- 
  - issue: https://github.com/getodk/central/issues/1588
  - closed by: https://github.com/getodk/central/pull/1589
- 
  - issue: https://github.com/getodk/central/issues/1597
  - closed by: https://github.com/getodk/central/issues/1627

But... not closed.

# Root cause

Possibly this regex?

https://github.com/getodk/central/blob/242416c17ad316ee818a140f7602bc1d19db77a9/.github/workflows/close-issues.yml#L23

alxndrsn:
Closes #1592

- [x] added some tests for the regex
- [ ] forked this repo, merged the PR to `master` in the fork, and tested with various new issues and different structures of commit message.  REVIEW: is this necessary?

Another approach would be to extract the JS of the `script` section of `close-issues.yml` into a `.js` file, and write more traditional tests.  However, this would be a much more significant change to the overhead of the workflow, as `actions/checkout` would have to be added, and NodeJS would need to be installed.

No effect.

No.

### DIFF
--- a/test/test-images.sh
+++ b/test/test-images.sh
@@ -54,7 +54,7 @@ check_path 180 / 'ODK Central'
 log "  Frontend started OK."
 
 log "Verifying backend..."
-check_path 2 /v1/projects '{"message":"Could not authenticate with the provided credentials.","code":401.2}'
+check_path 20 /v1/projects '{"message":"Could not authenticate with the provided credentials.","code":401.2}'
 log "  Backend started OK."
 
 log "Verifying enketo..."


.github/workflows/close-issues.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           script: |
             const body = context.payload.pull_request.body || "";
             const closedIssues = new Set();
-            const issueRegex = /(Closes|Fixes|Resolves)\s*:?\s+#(\d+)/gi;
+            const issueRegex = /(Closes|Fixes|Resolves)\s*:?\s+(?:(?:https:\/\/)?github.com\/getodk\/central\/issues\/|getodk\/central#|#)(\d+)/gi;
```
latin-panda, line 23: Optional: 
It could support variants of the same words, e.g., `Close`, `Closes`, `Closed` 
```suggestion
            const issueRegex = /((?:close|fix|resolve)(?:s|d|es|ed)?)\s*:?\s+(?:(?:https:\/\/)?github.com\/getodk\/central\/issues\/|getodk\/central#|#)(\d+)/gi;
```

alxndrsn: Good idea, I'll follow up :+1:

PR #1619, opened 2026-01-22
chore: upgrade node to v22.22

sadiqkhoja:
I guess we don't need to make corresponding change in `central-backend` and `central-frontend` repos.

All tests are passing.

NA

NA

NA


lognaturel:
> don't need to make corresponding change in central-backend and central-frontend repos

Agreed

Issue #1347, opened 2025-09-05
start-odk.sh and cron jobs do not complete when Sentry is blocked


yanokwa:
> The scripts (both start-odk.sh and the nightly cron jobs) hang when the service container can't reach the ODK Sentry instance on the internet. It would be great if this could be handled gracefully with an error or with a log message.

https://forum.getodk.org/t/memory-use-issue-using-docker-compose-stack-odkcentral-service-container/56473

yanokwa:
I've added documentation at https://github.com/getodk/docs/pull/2022/files. We'll be working on a hot fix to ensure Central can start even with Sentry is disabled or down.

PR #1624, opened 2026-01-27
Update submodules for v2025.4.2

ktuite:
This PR prepares the patch release v2025.4.1. 

- It updates the `client` submodule, matching the frontend `hotfix/v2025.4.x` branch with an update to Web Forms. 
- It updates the `server` submodule, matching the backend branch of the same name `hotfix/v2025.4.x` with a Sentry-related workaround for issue https://github.com/getodk/central/issues/1347. 

The thing to verify is that the new commit hashes matches the the hotfix branches for both frontend and backend.

I believe LN tried this configuration on dev to give it a final test.


lognaturel (approved):
It's what's on dev and the update files match what I expect!

PR #1636, opened 2026-02-05
Release v2026.1

matthew-white:
This PR prepares the release of v2026.1. It should only contain changes from other PRs that have already been approved and merged (and possibly merge commits from the `master` branch).


matthew-white:
I'm noticing that there's a conflict between `next` and `master` that would prevent `next` from being merged. It looks related to CSP things. @alxndrsn, would you mind taking a look? CC @lognaturel, who I think is familiar with what's gone into recent patch releases.

lognaturel:
The conflict is in the Web Forms CSP. https://github.com/getodk/central/pull/1636/commits/74fea1b5ab3d99bd710432afcfe038ce1c6ac12e adds `data:` to `img-src`. I think it's not a clean merge because https://github.com/getodk/central/pull/1636/commits/edf03bfab57a791be49d51b319cd3cd3ac47cccd was cherry-picked? I'm not sure what the preferred way of resolving the conflict is, maybe you're ok doing it as part of the merge into master, @matthew-white, so we don't have to first merge `master` into this one? You should be able to just accept the change from `next`, the diff is straightforward.

matthew-white:
I think my main preference is that we resolve the conflict ahead of time. There's enough going on during the release process that I don't think it's an ideal time to try to resolve conflicts; mistakes would be more likely.

How about we merge `master` into this branch? Or I think the "Resolve conflicts" button below would be another way to resolve the conflict. It provides an option to "Accept current change". Would it be alright if you or @alxndrsn resolved the conflict? I think I understand the change you're describing, but I'm also not familiar with these commits.

lognaturel:
`Resolve conflicts` is disabled for me saying `next` is protected. Not for you? I'll try to see whether I can push from CLI.

matthew-white:
It looks like I'm able to do it. I'd be happy to try it that way if you wouldn't mind checking my work.

lognaturel:
Huh! Yes, happy to check. Otherwise https://github.com/getodk/central/pull/1719 feels even uglier.

matthew-white:
The latest commit reflects the result of "Resolve conflicts". I saw the following conflict:

```
<<<<<<< next
    "default-src 'none'; connect-src 'self' https:; font-src 'self' data:; frame-src 'self'; img-src blob: data: https:; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; worker-src blob:; report-uri /csp-report";
=======
    "default-src 'none'; connect-src 'self' https:; font-src 'self' data:; frame-src 'self'; img-src blob: https:; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; worker-src blob:; report-uri /csp-report";
>>>>>>> master
```

I chose the top version of the line, since I saw that it included `data:` in `img-src`.

matthew-white:
Not sure if you were trying to avoid a merge commit. 😬 I feel like some level of merge inception is expected/normal when there are patch releases, depending on the content of the patch.

lognaturel:
It’s ok! Much better than having to have another level of merge commit anyway. Agree this looks right. 👍

matthew-white:
The only thing that's still needed for this PR is updates to submodules, so I think the PR is ready for review.

ktuite (approved):

Issue #1245, opened 2021-11-17
Additional SSL configuration for database


matthew-white:
Right now, the Backend database config provides limited support around SSL. The config may include an optional `ssl` property, which if specified, must be `true`. When `ssl` is `true`, `rejectUnauthorized` is automatically specified as `false`. We set it up this way because Slonik provided limited configuration around SSL and seemed to set `rejectUnauthorized` to `false`. We also wanted to ensure that Slonik and Knex connect to the database in the same way. (For additional background, see getodk/central-backend#377 and getodk/central-backend#394.)

That said, the latest release of Slonik (v25.1.0) provides additional options around SSL: see gajus/slonik#159 and gajus/slonik@6c3b178c9b52418b348fb0d8a7a8e75fb31047f1. It looks like it's now possible to use SSL without setting `rejectUnauthorized` to `false`.

I think this is a useful change, but I also think it might be a breaking change for us. Slonink seems to no longer use `pg-connection-string` to parse the connection string. Relatedly, it seems to no longer support `ssl` as a query parameter: it wants SSL options to be specified in a different way.

I think we'll need to look more into this before upgrading to v25.1.0 or later. It's also nice that this would provide a way for us to support additional SSL configuration.

CC also @florianm, who I know is interested in these options.

matthew-white:
Update: getodk/central-backend#1748 will allow users to specify `PG*` environment variables, including those related to SSL. We don't have an answer yet for environment variables that specify a filename, e.g., `PGSSLCERT`, but we're going to discuss that. Related: https://github.com/getodk/central-backend/pull/1748#discussion_r2837181324

alxndrsn:
@matthew-white can this be closed?

matthew-white:
Are you thinking we should close it because it's not a priority or for another reason? I think this is still something we may want to do at some point. Definitely not a high priority though.

Or is this issue superseded by #2109 and #2115?

alxndrsn:
> Are you thinking we should close it because it's not a priority or for another reason? I think this is still something we may want to do at some point. Definitely not a high priority though.
> 
> Or is this issue superseded by [#2109](https://github.com/getodk/central/issues/2109) and [#2115](https://github.com/getodk/central/issues/2115)?

Good question!  I actually think this issue was significantly improved by https://github.com/getodk/central-backend/pull/1748, and **would be** closed if https://github.com/getodk/central/issues/2109 is resolved.  So it can be left open until that ticket is concluded.

PR #1647, opened 2026-02-14
DB config, including port, through libpq env vars

brontolosone:
Alternative to this contributor PR: #915

Companion PR to getodk/central-backend#1748

This demonstrates using the backend's new DB configuration approach (through libpq's PG* environment variables) in the Central dockerized setup.

In a nutshell:
- Backwards compatible with existing .env files
- Though users should be encouraged to head to our docs to learn about the [PostgreSQL libpq env vars](https://www.postgresql.org/docs/current/libpq-envars.html) in order to set `PG*` environment variables in the `.env` to configure their database access to the full extent now made possible.

Deployed to dev.getodk.cloud .

Not being backwards compatible would have been easier and more clear. But then users would be required to do a (trivial? :man_shrugging: ) migration of their `.env` database directives to the new variables.

Users may have workarounds for our previous incapability of exposing the full configuration surface (for instance, [one couldn't configure the port number](https://github.com/getodk/central/pull/915#issuecomment-3855342646)).

In particular for the port number, users have been tacking it onto the host variable, which would work in a number of places (but also not, in some others, eg probably not for `wait-for-it`) as we would concatenate host and port number anyway to form a connection URI, and had hardcoded the port number there). But this doesn't work with the environment variables.

If we want to pursue backwards compatibility even for THAT kind of workaround, then I would suggest we modify `service/scripts/start-odk.sh` some more, see if the port is unset yet the host ends with a port specification, and then mangle the variables appropriately.

The additional comments I made in `.env.example` suffice to explain, in my opinion.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,14 +53,6 @@ services:
       - SYSADMIN_EMAIL=${SYSADMIN_EMAIL}
       - HTTPS_PORT=${HTTPS_PORT:-443}
       - NODE_OPTIONS=${SERVICE_NODE_OPTIONS:-}
-      - PGAPPNAME=${PGAPPNAME:-odkcentral}
-      # Legacy compatibility for DB_* environment variables
-      - PGHOST=${PGHOST:-${DB_HOST:-postgres14}}
-      - PGUSER=${PGUSER:-${DB_USER:-odk}}
-      - PGPASSWORD=${PGPASSWORD:-${DB_PASSWORD:-odk}}
-      - PGDATABASE=${PGDATABASE:-${DB_NAME:-odk}}
-      # Note that DB_SSL is handled through start-odk.sh
-      # End of legacy compatibility for DB_* environment variables
       - DB_POOL_SIZE=${DB_POOL_SIZE:-10}
       - EMAIL_FROM=${EMAIL_FROM:-no-reply@$DOMAIN}
       - EMAIL_HOST=${EMAIL_HOST:-mail}
@@ -73,6 +65,7 @@ services:
       - OIDC_ISSUER_URL=${OIDC_ISSUER_URL:-}
       - OIDC_CLIENT_ID=${OIDC_CLIENT_ID:-}
       - OIDC_CLIENT_SECRET=${OIDC_CLIENT_SECRET:-}
+      - PGAPPNAME=${PGAPPNAME:-odkcentral}
       - SENTRY_ORG_SUBDOMAIN=${SENTRY_ORG_SUBDOMAIN:-o130137}
       - SENTRY_KEY=${SENTRY_KEY:-3cf75f54983e473da6bd07daddf0d2ee}
       - SENTRY_PROJECT=${SENTRY_PROJECT:-1298632}
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -18,14 +18,22 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
 export SENTRY_TAGS
 
 # handle legacy DB_SSL configuration, as we couldn't conditionally template that in docker-compose.yml
-if [ "${DB_SSL}" = "true" ]; then
-  if [ -z "${PGSSLMODE}" ] && [ -z "${PGREQUIRESSL}" ]; then
+if [[ -v DB_SSL && "${DB_SSL}" = "true" ]]; then
+  if [[ -v PGSSLMODE && -v PGREQUIRESSL ]]; then
     export PGSSLMODE="require"
   else
     echo "Fatal: legacy 'DB_SSL=true' specified, but PGSSLMODE or PGREQUIRESSL has already been set. To resolve ambiguity, remove 'DB_SSL=true' from your .env file." > /dev/stderr
     exit 100
   fi
 fi
+
+# handle other legacy DB_* configuration, and fill in our defaults if nothing is specified
+# When these PG* variables are _defined_ (even if _empty_), they will be left as-is.
+[[ -v PGHOST ]] || export PGHOST=${DB_HOST:-postgres14}
+[[ -v PGUSER ]] || export PGUSER=${DB_USER:-odk}
+[[ -v PGPASSWORD ]] || export PGPASSWORD=${DB_PASSWORD:-odk}
+[[ -v PGDATABASE ]] || export PGDATABASE=${DB_NAME:-odk}
+[[ -v PGAPPNAME ]] || export PGAPPNAME=odkcentral
 
 echo "waiting for PostgreSQL to become connectable to..."
 while ! (psql --no-password --quiet --command "" > /dev/null 2>&1 || (echo "sleeping 1 second waiting for a database connection"; false)); do sleep 1; done


.env.template
```diff
@@ -14,11 +14,21 @@ HTTPS_PORT=443
 # Optional: configure Node
 # SERVICE_NODE_OPTIONS=
 
-# Optional: connect to a custom database server
-# DB_HOST=
-# DB_USER=
-# DB_PASSWORD=
-# DB_NAME=
+# Optional: a custom DB connection, for instance to one provisioned by your organization.
```
lognaturel: Wanted to give this some thought since it's what users will be interacting with. It might make sense to first get the backend work merged and then come back to finalize this!

It's very unlikely that a user would need to change a db connection for an existing server so only documenting PG* vars moving forward sounds good to me.

I'd rather we continue the pattern of having only a section header here and putting details in docs (so keep the line above and move the comments below to docs)

How about something like this:
```
# PGHOST=
# PGUSER=
# PGPASSWORD=
# PGDATABASE=
# Can also use other libpq environment variables, see docs
```

I think we can be very confident the first 4 won't change between postgres versions. That gives us the same configurable items we've had so far still in the template and a hint of where to go to learn about the rest.

matthew-white: > putting details in docs

That sounds great to me. :+1: I've added this topic to getodk/docs#2034 so we remember to update user docs. Both you and @brontolosone should feel welcome to edit that issue to capture any details or other thoughts about this topic.

files/service/scripts/start-odk.sh
```diff
@@ -17,6 +17,19 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
 # shellcheck disable=SC2090
 export SENTRY_TAGS
 
+# handle legacy DB_SSL configuration, as we couldn't conditionally template that in docker-compose.yml
+if [ "${DB_SSL}" = "true" ]; then
+  if [ -z "${PGSSLMODE}" ] && [ -z "${PGREQUIRESSL}" ]; then
+    export PGSSLMODE="require"
+    printf "Setting PGSSLMODE to \"require\", as DB_SSL=true.\nThis protects against eavesdropping on the database connection, but not against impersonation of the database server.\nRefer to the PostgreSQL manual to learn how to enable host authentication:\nhttps://www.postgresql.org/docs/current/libpq-ssl.html#LIBQ-SSL-CERTIFICATES\n"
+  else
+    echo "Not setting PGSSLMODE, as it (or PGREQUIRESSL) has already has been set."
```
matthew-white: Because we instruct users to `docker compose up` with `-d`, I think users will often miss this sort of logging. If the user has specified conflicting configuration, I think we should error in that case. E.g., if `DB_SSL` is `true`, but `PGSSLMODE` is defined, but something other than `require`, I think that should be an error so the user knows to fix up their configuration.

files/service/scripts/start-odk.sh
```diff
@@ -17,6 +17,19 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
...
+  fi
+fi
+
+echo "waiting for PostgreSQL to become connectable to..."
+while ! (psql --no-password --quiet --no-align --tuples --command "" > /dev/null 2>&1 || (echo "sleeping 1 second waiting for a database connection"; false)); do sleep 1; done
```
matthew-white: I don't see `--tuples` documented at https://www.postgresql.org/docs/14/app-psql.html. Is it the same as `--tuples-only`?

matthew-white: I reckon that `--no-align` and `--tuples` aren't doing much here given `--command ""`. If so, I think it'd be nice to streamline the command by removing those flags.

files/service/config.json.template
```diff
@@ -1,11 +1,10 @@
...
     "database": {
-      "host": "${DB_HOST}",
-      "user": "${DB_USER}",
-      "password": "${DB_PASSWORD}",
-      "database": "${DB_NAME}",
-      "ssl": ${DB_SSL},
+      "host": "",
+      "user": "",
+      "password": "",
+      "database": "",
```
matthew-white, line 7: What do you think about removing these entirely? I saw that `ssl` has been removed, and that made me wonder, why not just remove them all?

docker-compose.yml
```diff
@@ -47,17 +47,25 @@ services:
     volumes:
       - secrets:/etc/secrets
       - /data/transfer:/data/transfer
+    env_file: .env
```
matthew-white, line 50: This line is here in order to ensure that all `PG*` env vars in .env end up in the environment of the `service` container, right? I'm seeing that it will add some extra env vars as well:

- `SSL_TYPE`
- `HTTP_PORT`
- `SERVICE_NODE_OPTIONS`

That doesn't seem like a problem to me, but I wanted to make sure it was on your radar and to see if you had any general thoughts about this. Is there any risk stemming from env vars intended for other containers being passed through to `service`?

files/service/scripts/start-odk.sh
lognaturel: Good catch

server
alxndrsn, line 1: This PR probably shouldn't be merged while it updates the `server` submodule to point to a different repo.

matthew-white: That makes sense to me. I'm thinking that @brontolosone should merge getodk/central-backend#1748 first, then point the `server` submodule to the `master` branch.

brontolosone: I did that in order to have the tests succeed. Of course I wouldn't just merge it with that altered submodule... unless I would forget that I did that, the probability of which is now lowered ;-)

files/service/scripts/start-odk.sh
```diff
@@ -17,6 +17,19 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
 # shellcheck disable=SC2090
 export SENTRY_TAGS
 
+# handle legacy DB_SSL configuration, as we couldn't conditionally template that in docker-compose.yml
+if [ "${DB_SSL}" = "true" ]; then
+  if [ -z "${PGSSLMODE}" ] && [ -z "${PGREQUIRESSL}" ]; then
+    export PGSSLMODE="require"
+    printf "Setting PGSSLMODE to \"require\", as DB_SSL=true.\nThis protects against eavesdropping on the database connection, but not against impersonation of the database server.\nRefer to the PostgreSQL manual to learn how to enable host authentication:\nhttps://www.postgresql.org/docs/current/libpq-ssl.html#LIBQ-SSL-CERTIFICATES\n"
```
matthew-white: I think this information would be helpful in user docs, but I don't think it needs to be logged here. As I mention elsewhere, users aren't usually going to see this sort of logging.

Also, I feel like we need to finish #1245 first, otherwise this could end up pointing users to configuration that isn't available quite yet.

.env.template
brontolosone: fixed in 73e8d2c

files/service/scripts/start-odk.sh
brontolosone: Recap:

I think we have now landed on not exposing SSL config explicitly in the template, just referring to our documentation (in turn referring to libpq documentation) and letting it be the conversation (via `PG*` env vars) between user and libpq that it should be.

Yet for backwards compat with `DB_SSL=true`, we'll set `PGSSLMODE=require`, **unless** `PGSSLMODE` or `PGREQUIRESSL` are already defined (new-style config), in which case we'll error out.

files/service/config.json.template
brontolosone: because then the vars defined in less-specific config take over; you'd get the jubilant stuff from `default.json`, and then those'll be set as `PG*` env vars by the load-db-env compat shim...

docker-compose.yml
brontolosone: Discussed: if this happens (if we have env vars in `.env` that should not end up in the service container's env) we could for instance consider splitting off a separate `.env-$foo` file for that.

files/service/scripts/start-odk.sh
brontolosone: Fixed in 5cb5696

files/service/scripts/start-odk.sh
brontolosone: Effected in 5cb5696

files/service/scripts/start-odk.sh
brontolosone: removed in 5fc8e06

brontolosone: I co-reckon this, removed in 5fc8e06

docker-compose.yml
```diff
@@ -47,17 +47,25 @@ services:
...
+      # Legacy compatibility for DB_* environment variables
+      - PGHOST=${PGHOST:-${DB_HOST:-postgres14}}
+      - PGUSER=${PGUSER:-${DB_USER:-odk}}
+      - PGPASSWORD=${PGPASSWORD:-${DB_PASSWORD:-odk}}
+      - PGDATABASE=${PGDATABASE:-${DB_NAME:-odk}}
+      # Special case: DB_SSL. It is interpreted as "PGSSLMODE=require" (which provides transport encryption, but no authentication).
+      # Ideally we'd set the corresponding PG* variable here, but we can't, because we can't translate values here.
+      # So to maintain backwards compatibility with existing .env files, this one is handled in `start-odk.sh`:
+      # if `true`; and PGSSLMODE nor PGREQUIRESSL are defined, then we set PGSSLMODE=require.
       - DB_SSL=${DB_SSL:-null}
```
matthew-white:
```suggestion
      - DB_SSL=${DB_SSL:-}
```

If `DB_SSL` is only relevant to shell stuff now, maybe an empty string reads better than `null`? `null` made sense when this value was being passed off to JSON/JS, but maybe it's kind of historical baggage now.

No strong feelings, also fine to keep.

docker-compose.yml
```diff
@@ -47,17 +47,25 @@ services:
...
-      - DB_USER=${DB_USER:-odk}
-      - DB_PASSWORD=${DB_PASSWORD:-odk}
-      - DB_POOL_SIZE=${DB_POOL_SIZE:-10}
-      - DB_NAME=${DB_NAME:-odk}
+      - PGAPPNAME=${PGAPPNAME:-odkcentral}
+      # Legacy compatibility for DB_* environment variables
+      - PGHOST=${PGHOST:-${DB_HOST:-postgres14}}
+      - PGUSER=${PGUSER:-${DB_USER:-odk}}
+      - PGPASSWORD=${PGPASSWORD:-${DB_PASSWORD:-odk}}
+      - PGDATABASE=${PGDATABASE:-${DB_NAME:-odk}}
```
matthew-white: In practice, will users ever need one or more of these env vars to be empty strings? Seems like that's not possible right now: they will always fall back to some default value.

docker-compose.yml
brontolosone: Actually, we can now drop this line (and comment) altogether. If `DB_SSL` is set in the .env, it will now end up in the container's env (and thus `start-odk.sh`) anyways, because of that newly added `env_file: ".env"` in the dockerfile. Thus: 860b29d

docker-compose.yml
brontolosone: Ugh.

On my machine if I empty all these `PG*`s, I get connected to the database of the same name as my unix user, over a unix socket, using peer auth (no password).
So that's a scenario in which it matters (that these vars are empty), but that's not a scenario you'd encounter in our docker setup; you'd always connect over the network, not over a unix socket, thus no peer auth and no visibility of your unix uid on the serverside, and thus also no default database.

But I guess you could have an empty password. Plus PostgreSQL has a 'trust' auth mode which does what it says and works over the network. And in that case you'd also be annoyed because you can't express "no or empty password". You couldn't before this change either, though.

So what we really need is to be able to distinguish "variable is undefined" from "variable is empty". Bash can do that in a clean way apparently, so if we move this logic out of this yaml into `start-odk.sh` (where, anyway, the DB_SSL logic already exclusively lives now, since 860b29d), then we could behave properly.

brontolosone: 40c0b52 implements the above idea, it's running on dev.getodk.cloud .

docker-compose.yml
matthew-white: I actually think the explictness is helpful. It documents which environment variables in .env are relevant to the `service` container and which aren't. E.g., `DOMAIN=${DOMAIN}` and `SYSADMIN_EMAIL=${SYSADMIN_EMAIL}` could now go away given `env_file: "env"`, but I'm not totally convinced we want to do that yet.

So I'd say let's keep it `DB_SSL=${DB_SSL:-null}` or make it `DB_SSL=${DB_SSL:-}`, but I think it's worth it to continue listing the variable. Otherwise, there's no single list of environment variables anymore.

matthew-white: It makes sense to get the `PG*` env vars out of docker-compose.yml so that you can distinguish between empty vs. unset. It's necessary to move them out, and there's also no point in trying to list all the `PG*` vars, since we don't know which `PG*` vars will actually be used in .env. But I think `DB_SSL` is a different case, as we aren't passing it to Postgres. It's our own variable, and it'd be good to document its existence. So for `DB_SSL`, I'd prefer leaving it in docker-compose.yml for now. We can always reevaluate that later, just in terms of getting this PR merged.

matthew-white: Though I see you added a comment about `DB_SSL` to docker-compose.yml, so maybe that's enough. 🤔 But then it feels kind of inconsistent with our treatment of vars like `DOMAIN` and `SYSADMIN_EMAIL`.

brontolosone:
I added support for specifying empty `PG*` vars, we won't override them, and they'll end up in the container env.
@matthew-white Please re-review!

docker-compose.yml
matthew-white: I can see how it's nice that start-odk.sh handles all DB-related vars now without docker-compose.yml getting in the way. E.g., start-odk.sh is also now the only place where `DB_HOST` appears. So I think I'm talking myself into `DB_SSL` staying just in start-odk.sh and not docker-compose.yml. Sorry for the back and forth, I'm happy either way at this point.

brontolosone: BackwardsCompat and TidyBranchless are seldom seen walking hand in hand. :man_shrugging:

Issue #1640, opened 2026-02-09
[central] intermittent CI failure in test-images: No names found, cannot describe anything.


alxndrsn:
I'm not sure yet if this affects `next`/`master`, but it was seen for https://github.com/getodk/central/pull/1112.  I can't see an obvious reason why the changes in that PR should have broken `test-images`.

Note that the error was observed in a 3rd-party repo from which a PR was made:

https://github.com/yoursdearboy/central/actions/runs/21819260531/job/62948189917

```
#39 [nginx intermediate 4/5] RUN files/prebuild/write-version.sh
#39 0.243 fatal: No names found, cannot describe anything.
#39 DONE 0.4s

#41 [service intermediate 5/9] RUN git describe --tags --dirty > /tmp/sentry-versions/central
#41 0.149 fatal: No names found, cannot describe anything.
#41 ERROR: process "/bin/sh -c git describe --tags --dirty > /tmp/sentry-versions/central" did not complete successfully: exit code: 128
```

PR #1666, opened 2026-02-25
ci: reject PRs/commits which change submodules

alxndrsn:
Related: #1640

Tested on `master` of a forked repo, with changes to `master`, and a PR: https://github.com/alxndrsn/odk-central/pull/2

None.

No effect.

No.

### DIFF
--- a/test/check-submodules.sh
+++ b/test/check-submodules.sh
@@ -13,14 +13,12 @@ not_rel() {
 check_submodules() {
   local actualClientSubmodule
   local actualServerSubmodule
-  local expectedClientSubmodule
-  local expectedServerSubmodule
 
   actualClientSubmodule="$(git config --file .gitmodules --get submodule.client.url)"
   actualServerSubmodule="$(git config --file .gitmodules --get submodule.server.url)"
 
-  expectedClientSubmodule=https://github.com/getodk/central-frontend.git
-  expectedServerSubmodule=https://github.com/getodk/central-backend.git
+  local expectedClientSubmodule=https://github.com/getodk/central-frontend.git
+  local expectedServerSubmodule=https://github.com/getodk/central-backend.git
 
   if ! [[ "$actualClientSubmodule" = "$expectedClientSubmodule" ]]; then
     log "!!!"


test/check-submodules.sh
```diff
@@ -0,0 +1,77 @@
...
+if ! [[ "$GITHUB_REPOSITORY" = getodk/central ]]; then
+  not_rel "not running in canonical repository."
+fi
+
+log "Checking current branch/PR target..."
+log "  GITHUB_BASE_REF=$GITHUB_BASE_REF"
+log "  GITHUB_REF_NAME=$GITHUB_REF_NAME"
+if   [[ $GITHUB_BASE_REF = master ]] || [[ $GITHUB_BASE_REF = next ]]; then
+  log "PR to merge to '$GITHUB_BASE_REF'; checking submodules..."
+  check_submodules
```
matthew-white: Can't someone change the target branch after creating the PR? Would doing so rerun this script? If not, I think someone could start off by targeting one branch (something other than `master`/`next`), then switch to `master`/`next` and circumvent the check that way. Maybe that's a sequence that this PR isn't trying to prevent, I just wanted to mention it.

I also see below that the check would rerun anyway after the PR is merged to `master`/`next`.

test/check-submodules.sh
```diff
@@ -0,0 +1,77 @@
...
+elif [[ $GITHUB_REF_NAME = master ]] || [[ $GITHUB_REF_NAME = next ]]; then
```
matthew-white: Should `$GITHUB_REF_NAME` be enclosed in double quotes? I'm pretty sure it doesn't matter in `[[`, but I see that other `[[`s above use quotes (e.g., for `$GITHUB_REPOSITORY`), so I thought I'd ask.

matthew-white (approved):

test/check-submodules.sh
alxndrsn: > Can't someone change the target branch after creating the PR?

They can.

> Would doing so rerun this script?

I think so.

> I also see below that the check would rerun anyway after the PR is merged to `master`/`next`.

:+1: 

The main aim of the PR is to prevent merge accidents, rather than determined attackers with malicious PRs.

test/check-submodules.sh
alxndrsn: Incorrect/ambiguous quoting should be caught by `shellcheck` in CI:

https://github.com/getodk/central/blob/4c30fb3999a4de89802f2fdebdd6937a734bba94/.github/workflows/test.yml#L15-L16

In this case they're unnecessary, but I'll make the file consistent.

PR #1671, opened 2026-03-03
PG* DB env preparation

brontolosone:
Problem:
A bit of circling around in #1647 eventually resulted in a solution that would use Bash's expressiveness to deal with the `PG*` environment variables and in particular, backward compatibility for legacy `DB_*` variables. Defaults for the `PG*` env vars would not be defined via the docker environment (as we can't express what we want in Dockerese, to start with, it cannot distinguish between variable-undefined and variable-empty).
Thus in `start-odk.sh` we would do the necessary work. But, the environment created there would not be set for any random `docker compose exec service [...]` command. Only the env vars set through Docker's mechanisms are applied for those.
The central-related commands we run are run through wrappers in `files/service/scripts`, in addition to the `odk-cmd` wrapper (for creating users etc) that we steer users towards via the docs.
So if we set the environment variables when those wrappers are executed, we'll be good.
This PR implements that idea.

As a recap, the alternatives are:
1. Set PG* vars through Docker.
    - Upside: Consistent env for anything executed in the service container
    - Downside: Due to lack of expressiveness in Dockerese, this makes [certain DB configurations](https://github.com/getodk/central/pull/1647#discussion_r2866991204) impossible, and *again* we'd be in the way for the user for configuring their database connection to the full libpq extent possible, while the idea we started on this journey for was to *get out of the way*.
2. Run some preprocessor to generate an env file for Docker to source. The classic undesirable "use some build system to create inputs for some other build system".
    - Upside: Consistent env
    - Downside: There is no obvious canonical way of doing this, it's kinda antithetical to the idea of using Docker (as the win95 "Start" button), users won't like it if they have to run Make or something; it's throwing out the baby with the bath water; if we're going to create friction we might as well ask users to migrate their legacy `DB_*` env vars and be done with it.
3. This here solution.
    - Upside: Backwards compatible while keeping it possible for the user to use any `PG*` configuration.
    - Downside: We need to remember to use the wrappers. This PR takes care of the obvious entry points (`odk-cmd` and the scripts in `scripts/`), but of course since previously there weren't any obvious advantages to using those wrappers, we haven't been dogfooding them. For instance [here in Frontend](https://github.com/getodk/central-frontend/blob/v2025.4.2/e2e-tests/run-tests.sh#L63-L64) we have been doing variants `docker exec node [...].js`, which then ends in tears as the DB defaults haven't been set. So, a bit of cognitive overhead for us devs.

See discussion above.

See discussion above.

No changes necessary. In the docs we steer users to `odk-cmd`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -65,7 +65,7 @@ services:
       - PGDATABASE=${PGDATABASE-${DB_NAME:-odk}}
       - PGUSER=${PGUSER-${DB_USER:-odk}}
       - PGPASSWORD=${PGPASSWORD-${DB_PASSWORD:-odk}}
-      - PGAPPNAME=${PGAPPNAME:-odkcentral}
+      - PGAPPNAME=${PGAPPNAME-odkcentral}
       # End of libpq connection env var preparation.
       - DB_POOL_SIZE=${DB_POOL_SIZE:-10}
       - EMAIL_FROM=${EMAIL_FROM:-no-reply@$DOMAIN}
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -17,7 +17,7 @@ RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODEN
 
 
 ARG DB_SSL
-RUN [ "${DB_SSL}" = "true" ] && (echo 'You have "DB_SSL=true" defined (in your .env file, probably). This is no longer supported from Central 2026.1 onwards, but you can replace it with "PGSSLMODE=require". Please refer to the 2026.1.0 release notes for more information.'; exit 13) || true
+RUN [[ -v DB_SSL ]] && (echo '\n\n\n\n\nYou have the "DB_SSL" variable defined (in your .env file, probably).\nThis variable is no longer supported from Central 2026.1 onwards.\nThere is a new way of configuring SSL for your database, please see:\n\nhttps://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server\n\nPlease refer to the Central 2026.1.0 release notes for more information on this change.\n\n\n\n\n'; exit 13) || true
 
 FROM node:${node_version}-slim AS intermediate
 RUN apt-get update \


files/service/env.d/000_DB-env-defaults
```diff
@@ -0,0 +1,23 @@
+# if the PGSERVICE mechanism — libpq named configurations (specified through this env var)
+# sourced from a file (referenced through PGSERVICEFILE env var) is used, leave everything be.
+if ! [[ -v PGSERVICE ]]; then
```
matthew-white: What's the reason for these new lines? Feels like this should maybe go in its own PR.

brontolosone: It was an omission, this proviso should've been in #1647. I want dodge a bit of PR-busywork. It's all on-topic :-)

brontolosone:
> > #### Does this change require updates to documentation? If so, please file an issue [here](https://github.com/getodk/docs/issues/new) and include the link below.
> > No changes necessary. In the docs we steer users to `odk-cmd`.
> 
> Not always, unfortunately. E.g., at https://docs.getodk.org/central-backup/ we recommend:
> 
> ```shell
> docker compose exec service node /usr/odk/lib/bin/restore.js ...
> ```

We can provide a wrapper for that. I'm starting to think we might want a generic wrapper so that you can go `docker compose exec thewrapper node /some/js/file`, this would also help you and your `repl.js` use case.

files/service/env.d/000_DB-env-defaults
matthew-white: I definitely get not wanting to open a million different PRs, but I'd encourage a separate PR in this case. I promise to review it quickly. 😅 I have specific questions about these lines, but asking them here feels a tad off-topic. I think having that conversation in its own PR would make the change more transparent — both to those following this work now and to our future selves doing Git archeology.

brontolosone: The current state of this PR doesn't have this file (or equivalent code elsewhere) anymore, so you probably don't mind if I just press "Resolve conversation" :-)

docker-compose.yml
```diff
@@ -47,12 +47,24 @@ services:
...
       - DOMAIN=${DOMAIN}
       - SYSADMIN_EMAIL=${SYSADMIN_EMAIL}
       - HTTPS_PORT=${HTTPS_PORT:-443}
       - NODE_OPTIONS=${SERVICE_NODE_OPTIONS:-}
+      # Prepare libpq connection env vars, while handling legacy DB_* vars.
+      # Resolution order:
+      # 1. PGVAR specified in .env. This includes declarations with empty values, for instance, to use an empty password (PGPASSWORD=).
+      # 2. Corresponding legacy DB_VAR in .env. These are legacy overrides for custom database connections.
+      # 3. ODK default value.
+      - PGHOST=${PGHOST-${DB_HOST-postgres14}}
```
matthew-white:
```suggestion
      - PGHOST=${PGHOST-${DB_HOST:-postgres14}}
```

We've been using `:-` for `DB_*` variables up til the current release, so I think we should continue to do so. Empty values are meaningful for `PG*` variables, but they haven't been special for `DB_*` variables — they just fall back.

It's probably not a big deal either way, because the `DB_*` variables were commented out by default in .env.template, i.e., they were unset. So `-` and `:-` should behave the same unless someone has uncommented a `DB_*` variable and left it empty. But I think we might as well try to be as backwards-compatible as possible: "If you have `PG*` variables, we'll use those, but if you have `DB_*` variables, they'll behave exactly as before — including if they're set to empty."

docker-compose.yml
```diff
@@ -47,12 +47,24 @@ services:
     volumes:
       - secrets:/etc/secrets
       - /data/transfer:/data/transfer
-    env_file: .env
+    env_file:
+      - ".env"
```
matthew-white, line 53: Will this change be reverted or do you prefer it this way?

files/service/scripts/start-odk.sh
```diff
@@ -17,26 +17,8 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
...
 
-# handle legacy DB_SSL configuration, as we couldn't conditionally template that in docker-compose.yml
-if [[ -v DB_SSL && "${DB_SSL}" = "true" ]]; then
-  if [[ -v PGSSLMODE && -v PGREQUIRESSL ]]; then
-    export PGSSLMODE="require"
-  else
-    echo "Fatal: legacy 'DB_SSL=true' specified, but PGSSLMODE or PGREQUIRESSL has already been set. To resolve ambiguity, remove 'DB_SSL=true' from your .env file." > /dev/stderr
-    exit 100
-  fi
-fi
```
matthew-white, line 28: Starting a comment thread about something you wrote on the PR:

<blockquote>

what we haven't accomplished is backward compat for `DB_SSL=true`.
We'll have to handle that one through erroring out in the build scripts, pointing the user to the docs, explaining they need to add `PGSSLMODE=require` to their `.env`. But this is presumably a rare case
</blockquote>

I like this idea of erroring out in the build script. :+1: Or even erroring out in start-odk.sh if that's easier for some reason. You can miss a note in the docs, but you can't miss an error.

files/service/scripts/start-odk.sh
```diff
@@ -17,26 +17,8 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
...
-# When these PG* variables are _defined_ (even if _empty_), they will be left as-is.
-[[ -v PGHOST ]] || export PGHOST=${DB_HOST:-postgres14}
-[[ -v PGUSER ]] || export PGUSER=${DB_USER:-odk}
-[[ -v PGPASSWORD ]] || export PGPASSWORD=${DB_PASSWORD:-odk}
-[[ -v PGDATABASE ]] || export PGDATABASE=${DB_NAME:-odk}
-[[ -v PGAPPNAME ]] || export PGAPPNAME=odkcentral
-
 echo "waiting for PostgreSQL to become connectable to..."
-while ! (psql --no-password --quiet --command "" > /dev/null 2>&1 || (echo "sleeping 1 second waiting for a database connection"; false)); do sleep 1; done
+while ! (pg_isready --quiet || (echo "sleeping 1 second waiting for a database connection"; false)); do sleep 1; done
```
matthew-white: It probably won't come as a surprise that I'm tempted to ask that this go in a separate PR. 😅 It's just one line, so I'm not opposed to keeping it here, but could you say a little more about it? Is `pg_isready` preferred just because it's a specialized tool for this task, or are there functional changes that are happening here? (Is this a refactor or an enhancement?)

docker-compose.yml
brontolosone: A compelling argument! Fixed for all 5 of them in b4aa5c0.

docker-compose.yml
brontolosone: I prefer it this way because it's clearer that this is not where you define what "the" env file is (that was already .env, and will still be even if you change this), instead it's where you define potentially multiple env files to load (it's a list), but we just happen to want to load everything from only `.env`.

files/service/scripts/start-odk.sh
brontolosone: Yes. It's the tool for the job. The psql way is just very roundabout and imposes more cognitive load to figure out how I'm (ab)using it to do something that for `pg_isready` **is in the name**.

brontolosone: as discussed on Slack: while `pg_isready` is a better approximation than `psql` of the `wait-for-it` TCP connection tester we used before, the former `psql` approach is nicer in that it doesn't just check whether we can connect to the dabase server, but also whether we can connect to the actual database.
That's nice, because it means that upon misconfiguration, the container doesn't bootloop, just stays in that waiting state instead. And not bootlooping makes other things (such as `docker exec`) easier.
`pg_isready` is out, `psql` is in.

brontolosone: with 707290a.

files/service/scripts/start-odk.sh
brontolosone: With 13faa6e it looks a bit like this. The first line is in red (not for everyone, depends on your terminal).

```
 => ERROR [service pgdg 4/4] RUN [ "true" = "true" ] && (echo "You have "DB_SSL=true" defined (in your .env file, probably). This is no longer supported, but you can r  0.6s
------
 > [service pgdg 4/4] RUN [ "true" = "true" ] && (echo "You have "DB_SSL=true" defined (in your .env file, probably). This is no longer supported, but you can replace it with "PGSSLMODE=require". Please refer to the 2026.1.0 release notes for more information."; exit 13):
0.525 You have DB_SSL=true defined (in your .env file, probably). This is no longer supported, but you can replace it with PGSSLMODE=require. Please refer to the 2026.1.0 release notes for more information.
------
failed to solve: process "/bin/sh -c [ \"${DB_SSL}\" = \"true\" ] && (echo \"You have \"DB_SSL=true\" defined (in your .env file, probably). This is no longer supported, but you can replace it with \"PGSSLMODE=require\". Please refer to the 2026.1 release notes for more information.\"; exit 13)" did not complete successfully: exit code: 13
```

So not the prettiest, but it'll do.

service.dockerfile
```diff
@@ -16,6 +16,8 @@ RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODEN
       | gpg --dearmor > /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg
 
 
+ARG DB_SSL
+RUN [ "${DB_SSL}" = "true" ] && (echo 'You have "DB_SSL=true" defined (in your .env file, probably). This is no longer supported from Central 2026.1 onwards, but you can replace it with "PGSSLMODE=require". Please refer to the 2026.1.0 release notes for more information.'; exit 13) || true
```
matthew-white: Could we error out if `DB_SSL` is set to any value? I saw what you wrote in another comment on the PR:

> central itself used to bomb out on anything that wasn't literally `true`, so I think it's a safe assumption that if there's any value, then it's `true`, by now

I think that's very likely as well, so if we need to go that way, I'd be OK with that. But it feels just as easy to check for any value. At this point, any value would be unexpected and in my opinion worth erroring out for. I'm thinking of something like:

```suggestion
RUN [[ -v ${DB_SSL} ]] && (echo 'You have DB_SSL defined (in your .env file, probably). This is no longer supported from Central 2026.1 onwards. Please refer to the 2026.1.0 release notes for more information.'; exit 13) || true
```

server
matthew-white, line 1: Does the submodule need to change? I guess there's no harm, but looking at the four new commits, they don't really seem necessary for this PR.

docker-compose.yml
```diff
@@ -47,12 +49,24 @@ services:
...
+      # Prepare libpq connection env vars, while handling legacy DB_* vars.
+      # Resolution order:
+      # 1. PGVAR specified in .env. This includes declarations with empty values, for instance, to use an empty password (PGPASSWORD=).
+      # 2. Corresponding legacy DB_VAR in .env. These are legacy overrides for custom database connections.
+      # 3. ODK default value.
+      - PGHOST=${PGHOST-${DB_HOST:-postgres14}}
+      - PGDATABASE=${PGDATABASE-${DB_NAME:-odk}}
+      - PGUSER=${PGUSER-${DB_USER:-odk}}
+      - PGPASSWORD=${PGPASSWORD-${DB_PASSWORD:-odk}}
+      - PGAPPNAME=${PGAPPNAME:-odkcentral}
```
matthew-white: Should this be `-` rather than `:-`? Seems like we're using `-` for all the other `PG*` variables. Or is the idea that `PGAPPNAME` should never be empty?

service.dockerfile
brontolosone: (We probably want `[[ -v DB_SSL ]]`)
Sure, but I'll just drop a light hint that what they want is still possible. The way you frame it the impatient might conclude there's no out and they shouldn't upgrade because there's no SSL support anymore whereas the contrary is the case, we have gobs of it now in fact.

Fixed in d5bb49f.

server
brontolosone: d5bb49f

docker-compose.yml
brontolosone: We have released code where it worked like this... either empty or unset would result in falling back to the default.
I'm fine with a narrative in which that was a bug (only unset should result in filling in the default), which we're now fixing :-) except if you trying to trap me into making a separate bugfix PR, in which case I'll prefer `:-` after all :clown_face: 

Fixed in d5bb49f.

PR #1679, opened 2026-03-05
Update news for v2025.4.3

lognaturel:
I verified that there have been no additional posts made to the release thread and I'm ready to add the new post there in a sec!


sadiqkhoja (approved):
verified that the links open the right messages.

PR #1685, opened 2026-03-13
csp/blank.html: prevent favicon violation

alxndrsn:
* allow favicon
* rename policy, as it is no longer "disallowing all"

# Resolves:

## Firefox:

In `FaviconLoader.sys.mjs:224:20`:

```
Content-Security-Policy: (Report-Only policy) The page’s settings would block the loading of a resource (img-src) at https://dev.getodk.cloud/favicon.ico because it violates the following directive: “img-src https://translate.google.com”
```

## Chrome:

In `blank.html`:

```
Loading the image 'https://dev.getodk.cloud/favicon.ico' violates the following Content Security Policy directive: "img-src https://translate.google.com". The policy is report-only, so the violation has been logged but no further action has been taken.
```

* updated tests

Could just ignore this - seems to currently be filtered on the Sentry side.

No effect.

No.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -183,7 +183,7 @@ server {
     root /usr/share/nginx/html;
     try_files /blank.html =404;
 
-    add_header Content-Security-Policy-Report-Only "default-src 'none'; connect-src https://translate.google.com https://translate.googleapis.com; img-src https://translate.google.com; report-uri /csp-report" always;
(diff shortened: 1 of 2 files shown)


matthew-white:
@lognaturel, do you have time to take a look at this? I'm happy to take a look as well.

lognaturel:
@matthew-white Yes, can do!

lognaturel (approved):
Makes sense!

PR #1691, opened 2026-03-15
csp: include report samples where supported

alxndrsn:
This should help to understand future violations more easily.

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#report-sample

Related: https://github.com/getodk/security/issues/99

- [x] tests/CI
  - [x] unit tests - is `'report-sample'` included?
  - [x] integration tests - are samples actually included?
- [x] tested on dev.getodk.cloud & reviewed Sentry reports

Alternative approach would be to only include in directives relevant to getodk/security#99

However, there might be future cases where seeing samples would be useful.

No risk perceived:

* directives which are not understood should be ignored by browsers
* CSP-Report-Only issues shouldn't have any effects on the user anyway

No.


.github/workflows/test.yml
```diff
@@ -43,10 +43,11 @@ jobs:
       with:
         node-version: 24.13.0
     - run: cd test/nginx && npm clean-install
-    - run: cd test/nginx && npm run lint
```
alxndrsn: revert once passing

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -34,10 +36,15 @@ const allowGoogleTranslate = ({ 'connect-src':connectSrc, 'img-src':imgSrc, ...o
 
 const contentSecurityPolicies = {
   'backend-unmodified': {
-    'default-src': 'NOTE:FROM-BACKEND',
+    'default-src': [
+      'NOTE:FROM-BACKEND',
```
alxndrsn: why change?

test/nginx/src/lib.js
```diff
@@ -0,0 +1,51 @@
...
+  assert,
+  assertSentryReceived,
+  requestSentryMock,
+  resetSentryMock,
+};
+
+async function assertSentryReceived(...expectedRequests) {
+  const { status, body } = await requestSentryMock({ path:'/event-log' });
+  assert.equal(status, 200);
+  console.log(JSON.stringify({ expectedRequests, body:JSON.parse(body) }, null, 2));
```
alxndrsn: remove before merge

test/nginx/nginx.test.docker-compose.yml
```diff
@@ -41,3 +51,8 @@ services:
     ports:
       - "10000:80"
       - "10001:443"
+    networks:
+      - odk-central-nginx-test-network
+networks:
+  odk-central-nginx-test-network:
+    driver: bridge
```
alxndrsn: is this helpful?

test/playwright.config.js
```diff
@@ -0,0 +1,34 @@
+const { defineConfig, devices } = require('@playwright/test');
+
+const dockerHosts = [
+  'odk-nginx.example.test',
+  'o-fake-dsn.ingest.sentry.io',
```
alxndrsn: chromium shouldn't need to connect directly to the sentry DSN, so this should be removed

alxndrsn: if it can be removed, the whole setting can be inlined

.github/workflows/test.yml
```diff
@@ -43,10 +43,12 @@ jobs:
...
+    #- run: cd test/nginx && npm run lint
     - run: cd test/nginx && ./setup-tests.sh
-    - run: cd test/nginx && npm run test:nginx
-    - run: cd test/nginx && ./lint-config.sh
+    - run: cd test/nginx && npm run test:nginx:mocha
+    #- run: cd test/nginx && ./lint-config.sh
```
alxndrsn: revert before merge

test/nginx/nginx.test.docker-compose.yml
alxndrsn: no, removed

test/playwright.config.js
alxndrsn: Removed.

test/nginx/src/mocha/nginx.spec.js
alxndrsn: Unnecessary.  Removed.

matthew-white:
@lognaturel, I tagged you for review, since you're familiar with how the CSP is set up. I'm sure someone else could review as well if that would work better.

matthew-white:
I don't think this PR is going to get merged before regression testing begins, as it's going to start shortly. I don't think that matters a ton though; I think it'd be safe to merge this PR during regression testing. It's hard to imagine that changes to the `Content-Security-Policy-Report-Only` header would lead to a user-facing regression.

I'm still thinking that @lognaturel is best positioned to review, but I'm sure Sadiq or I could also review this week if that'd work better and help ship the PR with v2026.1.

lognaturel (approved):
I've verified that the changes to production code are limited to adding `report-sample` to CSP headers, that this is applied to directives we are likely to care about, and that the tests likely test what they intend to.

PR #1692, opened 2026-03-15
test/eslint: enable rule: comma-dangle

alxndrsn:
Including trailing commas in multiline arrays simplifies git diffs over time, which eases understanding of changes & code history.

This commit also resolves existing violations.

* ran locally
* CI

Including trailing commas in multiline arrays simplifies git diffs over time, which eases understanding of changes & code history.

An alternative might be to add eslint-ignore directives to existing violations, but these are noisy in their own way.

### DIFF
--- a/test/.eslintrc.js
+++ b/test/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
     ecmaVersion: 'latest',
   },
   rules: {
+    'comma-dangle': [ 'error', 'always-multiline' ],
     'eol-last': 'error',
     'no-tabs': 'error',
     'no-trailing-spaces': 'error',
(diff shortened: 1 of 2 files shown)


matthew-white (approved):
We're a bit inconsistent on this `comma-dangle` rule. In Frontend, it's set to `only-multiline`, while in Backend, it's turned off completely. In general, I feel like consistency across codebases on this rule would be nice (whether changing `central` to match Frontend or changing Frontend to match `central`). At the same time, I feel like `always-multiline` makes a lot of sense in the context of testing, where you often have lists of test cases that change over time. I've seen most others on the team use dangling commas at some point or another, so I don't feel like anyone would object to this rule being applied to tests here.

PR #1737, opened 2026-03-31
Enable HTTP/1.1 for nginx <-> backend

### LINKED ISSUE #1736: enable HTTP/1.1 for nginx <-> backend
Related: https://github.com/getodk/central-backend/pull/1760/commits/fa85d3e6

Nginx' default for proxied upstreams is to use HTTP/1.0.

In several places we use streaming responses, and in some of those (/v1/backup for sure) we want to be able to have it clientside detectable if stream production at the server crashes (as that can then inform further action clientside, for instance, toss the downloaded backup file if the download was incomplete and try anew).
Absent a way to instruct Node to send a TCP RST, we can use chunked transport encoding to achieve that, as those need to be finalized with a 0-length chunk, and if it's absent yet the TCP connection has been closed (eg with FIN), then the http client can tell. And in fact, at least `curl` does tell, it even has a dedicated exit code for incomplete transfers, and I presume HTTP libraries worth their salt also have some way of detecting this case.

So we want to use chunked transport encoding, and for that we need nginx to talk HTTP/1.1 to the backend server. There's a directive for that: `proxy_http_version`.

brontolosone:
Closes #1736

watched CI not explode, plus [a test](https://github.com/getodk/central/pull/1737/changes#diff-fdf0396ee2376b0619bd9e8a66d186b1d20db1b887b9d200386ab58f5926f326) that fails in when nginx <-> backend is HTTP/1.0 .

See #1736 for background.

N/A

N/A

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,6 +46,7 @@ jobs:
     - run: cd test/nginx && npm clean-install
     - run: cd test/nginx && ./setup-tests.sh
     - run: cd test/nginx && npm run test:nginx:mocha
+    - run: cd test/nginx && npm run test:nginx-chunked-stream-error-passthrough
     - run: cd test/nginx && ./lint-config.sh
 
     - run: cd test/nginx && npx playwright install --with-deps chromium-headless-shell
--- a/files/nginx/backend.conf
+++ b/files/nginx/backend.conf
@@ -6,3 +6,4 @@ proxy_redirect off;
 proxy_request_buffering on;
 proxy_buffering off;
 proxy_read_timeout 2m;
+proxy_http_version 1.1;  # because https://github.com/getodk/central/issues/1736
--- a/test/nginx/mock-http-server/index.js
+++ b/test/nginx/mock-http-server/index.js
@@ -9,7 +9,23 @@ const app = express();
 
 app.use((req, res, next) => {
   console.log(new Date(), req.method, req.originalUrl);
+  next();
+});
 
+app.get('/v1/chunked', (req, res) => {
+  // See https://github.com/getodk/central/issues/1736
+  // We don't have to set the transfer encoding to chunked explicitly; by default,
+  // node will produce a chunked response when we treat the response as a stream.
+  res.flushHeaders();
+  res.cork();
+  res.write('Pack it up, ');
+  res.uncork();
+  res.write('pack it in, ');
+  if (req.query.crash) throw new Error("let's pretend-play a bad thing happened and now we couldn't call .end() and thus we have an unterminated chunked stream on our hands.");
+  res.end('let me begin\n');
+});
+
+app.use((req, res, next) => {
   // always set CSP header to detect (or allow) leaks from backend through to the client
   res.set('Content-Security-Policy',             'default-src NOTE:FROM-BACKEND:block');
   res.set('Content-Security-Policy-Report-Only', 'default-src NOTE:FROM-BACKEND:reportOnly');
--- a/test/nginx/test-nginx-chunked-stream-error-passthrough.sh
+++ b/test/nginx/test-nginx-chunked-stream-error-passthrough.sh
@@ -1,0 +1,32 @@
+#!/bin/bash -eu
+set -o pipefail
+shopt -s inherit_errexit
+
+set -e
+STREAM_URL="https://odk-nginx.example.test:9001/v1/chunked"
+BROKEN_STREAM_URL="${STREAM_URL}?crash=1"
+CURL="curl --resolve odk-nginx.example.test:9001:127.0.0.1 --insecure --fail --verbose --silent --show-error"
+
+# check completed stream on HTTP/1.1
+$CURL --http1.1 ${STREAM_URL}
+
+# check completed stream on HTTP/2
+$CURL --http2-prior-knowledge ${STREAM_URL}
+
+
+function check_broken_stream_detectable() {
+    set +e
+    $CURL "${1}" "${BROKEN_STREAM_URL}"
+    local curl_exitcode=${?}
+    set -e
+    if [[ $curl_exitcode -ne "${2}" ]]; then
+        printf "\n\n\nCrashed stream production should have been detected:\n\t— for http/1.1, with exitcode 18: \"Partial file. Only a part of the file was transferred.\"\n\t— for http/2, with exitcode 92: \"HTTP/2 stream 1 was not closed cleanly\"\n\n\n"
+        exit 1
+    fi
+}
+
+# check interrupted stream on HTTP/1.1
+check_broken_stream_detectable --http1.1 18
+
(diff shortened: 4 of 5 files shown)


matthew-white:
I'm thinking that it probably makes sense for either @alxndrsn or @sadiqkhoja to review.

Since regression testing has already begun for v2026.1, I'm thinking that we shouldn't merge until .1 has been released.

alxndrsn (dismissed):
Any chance of a test?

From #1736:

> we want to use chunked transport encoding, and for that we need nginx to talk HTTP/1.1 to the backend server. 

E.g. a test showing that chunked transport encoding now works.

matthew-white:
Just checking my understanding. I thought we were using HTTP/2 these days in nginx. Does this change mean that nginx and Backend will communicate over HTTP/1.1, then nginx will communicate with the end user over HTTP/2? Would it help at all for nginx and Backend to communicate over HTTP/2?

brontolosone:
> Just checking my understanding. I thought we were using HTTP/2 these days in nginx. 

Not quite. We've enabled HTTP/2 for clients who express a desire to engage in HTTP/2, for the client <-> nginx part. The nginx <-> backend part has been HTTP/1.0, up to this point.
Clients could and can also still speak HTTP/1.1 or HTTP/1.0 to nginx.

How it works is that clients speak whatever to Nginx, currently including HTTP/1.0, HTTP/1.1, HTTP/2. The connection from Nginx to its backends is a different story.

With this here change stream interruption detection works when the client speaks HTTP/1.1 or HTTP/2 to Nginx. Nginx translates an interrupted HTTP/1.1 chunked-encoding stream into a HTTP/2 stream error (HTTP/2 doesn't have a chunked transfer encoding). I tested this with curl.

> Would it help at all for nginx and Backend to communicate over HTTP/2?

It's possible to do it, but I think letting our backend speak HTTP/2 could be a change with some ramifications, and I'm not sure it's worth it. Note that a setup such as ours, in which backend application servers are on HTTP/1.x while frontend reverse proxies speak fancier protocols with clients, are very common, not weird, nothing to be ashamed of :smile:

brontolosone:
> Any chance of a test?
> 
> From #1736:
> 
> > we want to use chunked transport encoding, and for that we need nginx to talk HTTP/1.1 to the backend server.
> 
> E.g. a test showing that chunked transport encoding now works.

f7dc21fe contains a test that fails in absence of f045490b.

alxndrsn:
> [f7dc21f](https://github.com/getodk/central/commit/f7dc21fe78116c75d894cc7ea85f8d4da3100f58) contains a test that fails in absence of [f045490](https://github.com/getodk/central/commit/f045490bd1bfd90572e91f956245a0883bc63bfd).

I tried reverting f045490, and the test still passes: https://github.com/alxndrsn/odk-central/actions/runs/23903112144.

brontolosone:
> > [f7dc21f](https://github.com/getodk/central/commit/f7dc21fe78116c75d894cc7ea85f8d4da3100f58) contains a test that fails in absence of [f045490](https://github.com/getodk/central/commit/f045490bd1bfd90572e91f956245a0883bc63bfd).
> 
> I tried reverting [f045490](https://github.com/getodk/central/commit/f045490bd1bfd90572e91f956245a0883bc63bfd), and the test still passes: https://github.com/alxndrsn/odk-central/actions/runs/23903112144.

Oh, I see - [http/1.1 for backends became the default since nginx v1.29.7](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version). That version [came out just over a week ago](https://nginx.org/en/CHANGES), just a little bit before I opened the issue :laughing: .

To see the test fail in absence of f045490, you can:
1. set the server's `proxy_http_version` to 1.0, or
2. travel a couple of days back in time so that you're on nginx < v1.29.7

Nevertheless, this PR still makes sense, as
1. it makes our dependency on http/1.1 explicit
2. documents why that is
3. and makes sure things keep working if we'd downgrade nginx
4. and provides warning if we'd swap out nginx for something else (where that something else speaks HTTP/1.0 to the backend).

alxndrsn:
> Nevertheless, this PR still makes sense, as
> 
>     1. it makes our dependency on http/1.1 explicit
>     2. documents why that is

This seems reasonable :+1:  Would this be testable with the existing mocha test framework?

brontolosone:
> > Nevertheless, this PR still makes sense, as
> > ```
> > 1. it makes our dependency on http/1.1 explicit
> > 2. documents why that is
> > ```
> 
> This seems reasonable 👍 Would this be testable with the existing mocha test framework?

Is there anything wrong with the `curl`-based tests? They're very to-the-point, in fact I expect we can put a curl oneliner in the docs with respect to how to backup safely. Backing up safely (not unnoticably saving interrupted streams) was the departure point for this PR. So the test I added is also testing whether `curl` keeps living up to expectations (because in case not, we'd have to think about how users can detect interrupted streams, and change the docs accordingly).

Curl is a fair standard for HTTP reproducability for this simple use case. After all, the network inspector panels of Chrome and Firefox feature "Copy as curl", they don't feature "Copy as Mocha test". So unless you have a compelling argument that trumps those benefits, *and* can justify the $-cost of rewriting the curl-based test into mocha, ...

Maybe you're just curious as to what an interrupted stream looks like as an XMLHttpRequest or in a fetch()? I am. I hope they'll show up as a NetworkError. But that's outside the scope of this PR.

alxndrsn:
> Backing up safely (not unnoticably saving interrupted streams) was the departure point for this PR.

:+1: starting to make sense

So is the aim to test something like "the recommended backup command should detect if a downloaded file is incomplete"?

brontolosone:
> > Backing up safely (not unnoticably saving interrupted streams) was the departure point for this PR.
> 
> 👍 starting to make sense
> 
> So is the aim to test something like "the recommended backup command should detect if a downloaded file is incomplete"?

The aim of the test is to assert that unfinalized HTTP/1.1 chunck-encoded streams as produced by the backend, are apparent as such from the perspective of a HTTP client, even when nginx interposes.

So an HTTP client is required. `libcurl` is a good model HTTP client library, it's widely used and extensive, and the `curl` utility is a prime choice of instrumentable client that uses that library, especially since one of the use cases for broken stream detection is downloading DB backups, for which we don't write the client✝, yet basing documentation on how to safely backup around a `curl` incantation is very reasonable and is what I plan to do.

✝) otherwise we could have considered some other form of signaling, such as adding a specific HTTP footer to signal that the stream is complete

brontolosone:
Intended "how to download backups without letting interrupted streams go unnoticed" documentation, which indeed advocates using `curl`, is in PR getodk/docs#2052 .

lognaturel:
It turned out to be an odd coincidence that the default changed part way through! We've decided not to add tests at this level for now. We will rely on the manual end-to-end testing of backups to alert us if somehow functionality in this area changes. Thank you!

PR #1738, opened 2026-03-31
fix #1728 && fix sh compat

### LINKED ISSUE #1728: `docker compose` logs warning about `DB_SSL`
Any `docker compose` command now seems to log the following warning:

```
WARN[0000] The "DB_SSL" variable is not set. Defaulting to a blank string.
```

[Slack discussion](https://getodk.slack.com/archives/C01TKJ7TG7J/p1774554683385649)

brontolosone:
Closes #1728

Also fixes `sh` compatibility.

Manual testing, checking whether `docker compose build` displays the desired behaviour in presence and absence of a DB_SSL .env var (of nonzero length).

N/A

N/A

N/A

### DIFF
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -15,9 +15,8 @@ RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODEN
     && curl https://www.postgresql.org/media/keys/ACCC4CF8.asc \
       | gpg --dearmor > /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg
 
-
 ARG DB_SSL
-RUN [[ -v DB_SSL ]] && (echo '\n\n\n\n\nYou have the "DB_SSL" variable defined (in your .env file, probably).\nThis variable is no longer supported from Central 2026.1 onwards.\nThere is a new way of configuring SSL for your database, please see:\n\nhttps://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server\n\nPlease refer to the Central 2026.1.0 release notes for more information on this change.\n\n\n\n\n'; exit 13) || true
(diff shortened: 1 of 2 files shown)


service.dockerfile
```diff
@@ -15,9 +15,8 @@ RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODEN
     && curl https://www.postgresql.org/media/keys/ACCC4CF8.asc \
       | gpg --dearmor > /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg
 
-
 ARG DB_SSL
-RUN [[ -v DB_SSL ]] && (echo '\n\n\n\n\nYou have the "DB_SSL" variable defined (in your .env file, probably).\nThis variable is no longer supported from Central 2026.1 onwards.\nThere is a new way of configuring SSL for your database, please see:\n\nhttps://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server\n\nPlease refer to the Central 2026.1.0 release notes for more information on this change.\n\n\n\n\n'; exit 13) || true
+RUN [ -z ${DB_SSL} ] || (/bin/echo -e '\n\n\n\n\nYou have the DB_SSL variable defined (in your .env file, probably).\nThis variable is no longer supported from Central 2026.1 onwards.\nThere is a new way of configuring SSL for your database, please see:\n\nhttps://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server\n\nPlease refer to the Central 2026.1.0 release notes for more information on this change.\n\n\n\n\n'; exit 13)
```
matthew-white: I'm thinking that it'd be a good idea to enclose `${DB_SSL}` in double quotes, since it's now `[`, not `[[`. We don't expect `DB_SSL` to be anything other than `true` or empty string, but I think it's nice to handle unexpected values gracefully, and no quotes could be a problem if `DB_SSL` contains white space.

matthew-white: Could you say more about the difference between `echo` and `/bin/echo`? Does it have to do with `-e`? What do you think about using `printf` instead?

brontolosone: Yes it appeared RUN is executed with `sh` of which the shell builtin `echo` doesn't honor `-e` (doesn't complain about it either? maybe it does something else?). Hence use of the external `echo` which does honor `-e`.

Yes `printf` is another alternative.

brontolosone: fixed

matthew-white: I feel like `printf` would be a little clearer, but I also don't think it's a big deal. When I saw `/bin` before `echo`, I wasn't really sure what the intention was at first. Just my two cents, no need to change if you disagree.

matthew-white (approved):

PR #1740, opened 2026-03-31
npm: harden default config

alxndrsn:
* ignore scripts
* add 14 day cooldown period for dependencies

See:

* https://docs.npmjs.com/cli/v11/configuring-npm/npmrc
* https://docs.npmjs.com/cli/v11/using-npm/config#ignore-scripts
* https://docs.npmjs.com/cli/v11/using-npm/config#min-release-age

CI.

Other approaches not considered.

Might affect devs or CI if there is a dep which requires scripts, or is too new.  This should be detected quickly and would not affect users.

No.


test/nginx/mock-http-server/.npmrc
```diff
@@ -0,0 +1,2 @@
+ignore-scripts=true
+min-release-age=7
```
matthew-white:
```suggestion
min-release-age=14
```

test/nginx/mock-sentry/.npmrc
```diff
@@ -0,0 +1,2 @@
+ignore-scripts=true
+min-release-age=7
```
alxndrsn:
```suggestion
min-release-age=14
```

test/.npmrc
```diff
@@ -0,0 +1,2 @@
+ignore-scripts=true
+min-release-age=7
```
alxndrsn:
```suggestion
min-release-age=14
```

matthew-white (approved):

PR #1765, opened 2026-04-04
feat: Support self-hosted Sentry via DSN

### LINKED ISSUE #1489: Support for self-hosted Sentry
Hello,

I don't understand why the Sentry host is set to “ingest.sentry.io.” Why not provide the Sentry DSN directly, which would allow us to use self-hosted sentry?

aryanjsingh:
## Summary

This change updates the Sentry configuration to use a DSN (Data Source Name) instead of separate fields for organization, key, and project. This allows users to configure Central with self-hosted Sentry instances. The Nginx configuration has been updated to parse the DSN to construct the CSP reporting endpoint, and the setup is now more robust when Sentry is not configured.

## Changes

- **.env.template**: Replaced SENTRY_ORG_SUBDOMAIN, SENTRY_KEY, and SENTRY_PROJECT with SENTRY_DSN to support self-hosted Sentry instances, which is the standard way to configure Sentry clients.
- **files/nginx/odk.conf.template**: Updated the `/csp-report` location block to use a single `SENTRY_CSP_ENDPOINT` variable. This variable will be constructed from the `SENTRY_DSN` in the `setup-odk.sh` script, allowing for flexible Sentry endpoint configuration.
- **files/nginx/setup-odk.sh**: Added logic to parse the SENTRY_DSN environment variable to construct the Sentry CSP endpoint URL. Also added a fallback for when SENTRY_DSN is not set, which makes the Nginx configuration robust by replacing the Sentry proxy block with a simple `return 204` to gracefully handle CSP reports.
- **files/service/config.json.template**: Updated the backend service configuration to accept a Sentry `dsn` instead of separate `orgSubdomain`, `key`, and `project` fields. This aligns with the move to a single DSN for Sentry configuration.

## Related Issue

Closes #1489

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -38,9 +38,7 @@ HTTPS_PORT=443
 # OIDC_CLIENT_SECRET=
 
 # Optional: configure error reporting
-# SENTRY_ORG_SUBDOMAIN=
-# SENTRY_KEY=
-# SENTRY_PROJECT=
+# SENTRY_DSN=
 # SENTRY_TRACE_RATE=
 
 # Optional: configure S3-compatible storage for binary files
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -200,7 +200,7 @@ server {
   }
 
   location /csp-report {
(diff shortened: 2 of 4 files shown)


brontolosone:
@aryanjsingh, as you closed your PR: are you no longer interested in support for self-hosted sentry? Or are you thinking about some other way to accomplish same?

Issue #1770, opened 2026-04-06
Validate docker-compose.yml for duplicate env variables


sadiqkhoja:
Originally reported at https://forum.getodk.org/t/odk-central-v2026-1-update-error/57948

I think it is a good idea to validate/lint the docker-compose.yml for duplicate env variables. Most of the docker composer versions allow duplicate env var but some have regression which prevents running any docker compose command if yml file has duplicate env var.

matthew-white:
Just wanted to clarify that this issue is different from #1771. #1771 was about removing the one duplicate that already existed. This issue is about preventing future duplicates.

PR #1772, opened 2026-04-06
Fixes: remove duplicate PGAPPNAME

### LINKED ISSUE #1771: Duplicate PGAPPNAME values in docker-compose.yml
We have two declarations of `PGAPPNAME` in `docker-compose.yml`
- https://github.com/getodk/central/blob/master/docker-compose.yml#L69
- https://github.com/getodk/central/blob/master/docker-compose.yml#L83

I would vote for removing the second one (line 83) since the first one is where all the `PG*` variables are.

Relatedly, I'd also likely vote for a different default value for consistency. That is instead of `odkcentral`, feels like `odk` would be more aligned with the other PG variables.

Original report: See https://forum.getodk.org/t/odk-central-v2026-1-update-error/57948

sadiqkhoja:
Closes https://github.com/getodk/central/issues/1771

Addresses https://forum.getodk.org/t/odk-central-v2026-1-update-error/57948

All tests are passing.

In this PR I have just removed the duplicate env var PGAPPNAME but I have created a corresponding issue to validate yml file for catching such issues proactively https://github.com/getodk/central/issues/1770

No

No


docker-compose.yml
```diff
@@ -66,7 +66,7 @@ services:
       - PGDATABASE=${PGDATABASE-${DB_NAME:-odk}}
       - PGUSER=${PGUSER-${DB_USER:-odk}}
       - PGPASSWORD=${PGPASSWORD-${DB_PASSWORD:-odk}}
-      - PGAPPNAME=${PGAPPNAME-odkcentral}
+      - PGAPPNAME=${PGAPPNAME-odk-central}
```
matthew-white: Are there any side effects to expect from this rename? Could it be a problem for some users? Or is it just for display/logging purposes?

matthew-white (approved):

docker-compose.yml
lognaturel: Excellent question. Even if it's just for logging, it's been set to `odkcentral` forever so could certainly be annoying. As we've discussed in Slack, let's keep it without the dash after all. Thank you!

lognaturel:
Looks like we're going to have a proper point release so should this target `next`?

docker-compose.yml
matthew-white: It was added relatively recently, in #1202, but it's been like that for a couple of releases now.

brontolosone: [It is possible people are already relying on a value of `odkcentral`.](https://github.com/getodk/central/issues/1771#issuecomment-4195627526).

sadiqkhoja:
removed the last commit that changed the default value of PGAPPNAME

PR #1776, opened 2026-04-08
csp: allow data: URLs in worker-src for web-forms

### LINKED ISSUE #1775: CSP violation: worker-src data: in web-forms image size widget
https://getodk.sentry.io/issues/7379561532/?project=5724763

```json
{
  "csp-report": {
    "effective_directive": "worker-src",
    "blocked_uri": "data",
    "document_uri": "https://staging.getodk.cloud/projects/224/forms/image_widget_size_pixels/submissions/new",
    "original_policy": "default-src 'none' 'report-sample'; connect-src 'self' https:; font-src 'self' data:; frame-src 'self'; img-src blob: data: https:; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'report-sample' 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; worker-src 'report-sample' blob:; report-uri https://staging.getodk.cloud/csp-report",
    "referrer": "https://staging.getodk.cloud/projects/224/forms/image_widget_size_pixels/submissions",
    "status_code": 200,
    "violated_directive": "worker-src",
    "column_number": 1,
    "disposition": "report"
  }
}
```

Seems to relate to the web-forms image size widget.

The triggering data-URL code decodes to:

```js
(function lA(uA) {
    if (typeof createImageBitmap > "u") {
        uA.postMessage(!1);
        return
    }
    Promise.resolve().then(function() {
        var gA = new OffscreenCanvas(10,10)
          , dA = gA.getContext("2d");
        return dA.rect(0, 0, 1, 1),
        createImageBitmap(gA, 0, 0, 1, 1)
    }).then(function() {
        return uA.postMessage(!0)
    }, function() {
        return uA.postMessage(!1)
    })
}
)(self);
```

alxndrsn:
Closes #1775

Added a test.

Maybe not a great solution - it seems unfortunate to need to loosen the CSP for feature detection.

No effect.

No.


lognaturel:
Do you remember what we have `worker-src blob:` for? If we already need to have that, adding `data:` doesn't seem worse but ideally we wouldn't need either. CC @garethbowen in case you have other ideas!

files/nginx/odk.conf.template
```diff
@@ -91,7 +91,7 @@ map $arg_st $redirect_single_prefix {
 map $request_uri $central_frontend_csp {
   # Web Forms CSP for /f/... and /projects/.../forms/... routes
   ~^/(?:f/[^/]+(?:/.*)?|projects/\d+/forms/[^/]+/(?:(?:draft/)?(?:preview|submissions/new(?:/offline)?)|submissions/[^/]+/edit)(?:/)?)(?:\?.*)?$
-    "default-src 'report-sample' 'none'; connect-src 'self' https:; font-src 'self' data:; form-action 'self'; frame-ancestors 'self'; frame-src 'self' https://getodk.github.io/central/; img-src blob: data: https:; manifest-src 'self'; media-src blob:; object-src 'none'; script-src 'report-sample' 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; worker-src 'report-sample' blob:; report-uri /csp-report";
+    "default-src 'report-sample' 'none'; connect-src 'self' https:; font-src 'self' data:; form-action 'self'; frame-ancestors 'self'; frame-src 'self' https://getodk.github.io/central/; img-src blob: data: https:; manifest-src 'none'; media-src 'none'; object-src 'none'; script-src 'report-sample' 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; worker-src 'report-sample' blob: data:; report-uri /csp-report";
```
```suggestion
    "default-src 'report-sample' 'none'; connect-src 'self' https:; font-src 'self' data:; form-action 'self'; frame-ancestors 'self'; frame-src 'self' https://getodk.github.io/central/; img-src blob: data: https:; manifest-src 'self'; media-src blob:; object-src 'none'; script-src 'report-sample' 'self' 'wasm-unsafe-eval'; style-src 'self' 'unsafe-inline'; worker-src 'report-sample' blob: data:; report-uri /csp-report";
```

alxndrsn:
> Do you remember what we have `worker-src blob:` for?

It was originally introduced for OpenLayers maps in:

* https://github.com/getodk/central/pull/1530

It was then duplicated when enketo & web-forms CSPs were split in:

* https://github.com/getodk/central/pull/1526

> If we already need to have that, adding `data:` doesn't seem worse but ideally we wouldn't need either.

I think technically `data:` is less safe than `blob:`, as a data URL (`data:`) can be injected with only HTML or DOM manipulation, whereas an object URL (`blob:`) requires script injection.

Ideally we'd avoid both; adding `blob:` when `data:` is already listed should not be a significant increase in risk.

Ref:

* https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data
* https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/blob
* https://github.com/openlayers/openlayers/issues/16046

lognaturel (approved):
We could consider inlining pica since it doesn't seem to be actively maintained. But I think this is ok for now.

PR #1778, opened 2026-04-08
csp: allow central's outdated-version page

### LINKED ISSUE #1777: CSP: frame-src should allow for outdated version iframe
https://github.com/getodk/central-frontend/blob/788563a129b9e02393d8881e87023b48e8c258ff/src/components/outdated-version.vue#L49 can make iframes with `src=https://getodk.github.io/central/outdated-version.html?...`.

It looks like this can be shown on any page in the app, including enketo-passthrough.

alxndrsn:
Closes #1777

CI.

Could be more specific, and list each specific page.  Adds header size overhead, and annoying to maintain as it might be forgotten if new content is added in https://github.com/getodk/central/tree/master/docs.

No effect.

No.

### DIFF
--- a/test/nginx/src/mocha/nginx.spec.js
+++ b/test/nginx/src/mocha/nginx.spec.js
@@ -14,7 +14,7 @@ const self = `'self'`;
 const unsafeInline = `'unsafe-inline'`;
 const wasmUnsafeEval = `'wasm-unsafe-eval'`;
 
-// Central has notifications shared from GitHub Pages including:
+// Central has notifications defined in https://github.com/getodk/central/tree/master/docs, and served from GitHub Pages.  These include:
 //
 // * https://getodk.github.io/central/news.html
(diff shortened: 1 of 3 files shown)


test/nginx/src/mocha/nginx.spec.js
```diff
@@ -14,6 +14,12 @@ const self = `'self'`;
 const unsafeInline = `'unsafe-inline'`;
 const wasmUnsafeEval = `'wasm-unsafe-eval'`;
 
+// Central has notifications shared from GitHub Pages including:
```
```suggestion
// Central has notifications defined in https://github.com/getodk/central/tree/master/docs, and served from GitHub Pages.  These include:
```

matthew-white (approved):
I don't really remember why we decided to host outdated-version.html from central/docs/ rather than render it in Frontend. But given that that's how it's set up, it makes sense to me that the CSP has to change in this way.

PR #1814, opened 2026-04-15
csp: embed policy name in reports

alxndrsn:
This should significantly simplify understanding of reports.

Updated tests.

It makes every header bigger, so I've aimed to keep the additional data small.

Should not have a noticeable effect.

No.


files/nginx/odk.conf.template
```diff
@@ -114,7 +114,7 @@ server {
 
   server_tokens off;
 
-  add_header Content-Security-Policy-Report-Only "default-src 'report-sample' 'none'; connect-src https://translate.google.com https://translate.googleapis.com; img-src https://translate.google.com; report-uri /csp-report" always;
+  add_header Content-Security-Policy-Report-Only "default-src 'report-sample' 'none'; connect-src https://translate.google.com https://translate.googleapis.com; img-src https://translate.google.com; report-uri /csp-report?p=fu" always;
```
```suggestion
  add_header Content-Security-Policy-Report-Only "default-src 'report-sample' 'none'; connect-src https://translate.google.com https://translate.googleapis.com; img-src https://translate.google.com; report-uri /csp-report?p=err" always;
```

nginx.dockerfile
```diff
@@ -30,12 +30,11 @@ RUN mkdir -p /usr/share/odk/nginx/
...
-COPY files/nginx/common-headers.conf /usr/share/odk/nginx/
-COPY files/nginx/robots.txt /usr/share/nginx/html
-COPY --from=intermediate client/dist/ /usr/share/nginx/html
-COPY --from=intermediate /tmp/version.txt /usr/share/nginx/html
+COPY files/nginx/nginx-conf/* /usr/share/odk/nginx/
+COPY --from=intermediate \
+    /tmp/version.txt \
+    client/dist/ \
+    /usr/share/nginx/html/
+COPY files/nginx/robots.txt /usr/share/nginx/html/
```
alxndrsn: upstream

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -101,31 +109,11 @@ const contentSecurityPolicies = {
         reportSample,
         'blob:',
       ],
-      'report-uri':     '/csp-report',
     }),
   },
-  'disallow-all': {
```
alxndrsn: upstream rename

test/nginx/mock-sentry/index.js
```diff
@@ -102,6 +103,7 @@ const server = (() => {
       }
       cb(null, createSecureContext(goodCreds));
     },
+    secureOptions: crypto.constants.SSL_OP_NO_TICKET,
```
alxndrsn, line 107:
```suggestion
    // Without this setting, nginx pools connections to mock-sentry which are then missing SNI info when re-used.
    secureOptions: crypto.constants.SSL_OP_NO_TICKET,
```

test/nginx/src/mocha/nginx.spec.js
alxndrsn: https://github.com/getodk/central/pull/1823

nginx.dockerfile
alxndrsn: or just revert

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -1000,7 +1018,7 @@ function standardTestSuite({ fetchHttp, fetchHttp6, apiFetch, apiFetch6, forward
         }
 
         // then
-        assert.isOk(caught);
+        assert.isOk(caught, 'An error should have been thrown.');
```
```suggestion
        assert.isOk(caught);
```

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -958,19 +961,34 @@ function standardTestSuite({ fetchHttp, fetchHttp6, apiFetch, apiFetch6, forward
...
-      });
+    describe('report URLs', () => {
+      Object.values(contentSecurityPolicies)
+          .map(csp => csp.codename)
+          .flatMap(codename => [
+            `/csp/r/${codename}`,
+            `/csp/b/${codename}`,
+            `/csp/b/${codename}`,
+            `/csp/r/${codename}`,
+          ])
```
alxndrsn, line 970:
```suggestion
          .flatMap(codename => [
            `/csp/b/${codename}`,
            `/csp/r/${codename}`,
          ])
```

files/nginx/odk.conf.template
```diff
@@ -209,8 +209,18 @@ server {
     include /usr/share/odk/nginx/common-headers.conf;
   }
 
-  location /csp-report {
```
matthew-white, line 212: I think the removal of this line probably means that there's something to change about user docs. At https://docs.getodk.org/central-install-digital-ocean/#disabling-or-customizing-sentry, when describing how to disable Sentry, this line is referenced.

files/nginx/odk.conf.template
```diff
@@ -209,8 +209,18 @@ server {
...
+  location = /csp/b/ek  { include /usr/share/odk/nginx/csp.conf; }
+  location = /csp/r/ek  { include /usr/share/odk/nginx/csp.conf; }
+  location = /csp/b/err { include /usr/share/odk/nginx/csp.conf; }
+  location = /csp/r/err { include /usr/share/odk/nginx/csp.conf; }
+  location = /csp/b/fe  { include /usr/share/odk/nginx/csp.conf; }
+  location = /csp/r/fe  { include /usr/share/odk/nginx/csp.conf; }
+  location = /csp/b/wf  { include /usr/share/odk/nginx/csp.conf; }
+  location = /csp/r/wf  { include /usr/share/odk/nginx/csp.conf; }
+  location = /csp/b/XX  { include /usr/share/odk/nginx/csp.conf; }
+  location = /csp/r/XX  { include /usr/share/odk/nginx/csp.conf; }
```
matthew-white, line 225: Most of these make sense to me, but there are two that I don't fully understand:

- `err`. I think this means: "The default Content Security Policy. It's supposed to be overridden, so if it's seen in Sentry, something unexpected has happened."
- `XX`. In tests, I see this is related to `backend-unmodified`. I don't totally remember what `backend-unmodified` is for though. Would you mind jogging my memory? Is `XX` just for testing? Could it be renamed to `test`?

matthew-white:
> This should significantly simplify understanding of reports.

Just to make sure I understand this, I think it helps because it will surface the policy name in Sentry, because Sentry shows the full Content Security Policy, which includes the `report-uri`. Is that right?

alxndrsn:
Consider for the future.  For now, other approaches have helped identify which policy is active.

Issue #1822, opened 2026-04-17
No browser caching for login logo and hero images on S3


lognaturel:
When I refresh, I always see:

<img width="572" height="96" alt="Image" src="https://github.com/user-attachments/assets/d5f9341a-d011-4c73-a61c-ff9bc42a638a" />

And it takes a while for each to load.

I have not tried with non-S3 images yet.

matthew-white:
> I have not tried with non-S3 images yet.

There may be a significant difference between S3 and non-S3 images in this area. It looks to me like an S3 blob is always a redirect, whereas non-S3 blobs get an ETag and possible caching:

- The responses above are 307s, whereas I was expecting 304s.
- [Related Backend code](https://github.com/getodk/central-backend/blob/b490cba1f58bcdb77e0c16516d4fef082e0d2254/lib/util/blob.js#L54-L62)

One way to test that would be to upload an entirely new image, one that hasn't been uploaded to S3 yet. I'd expect that to get 304 responses up until the point when it gets uploaded to S3.

CC @alxndrsn, who's definitely more familiar than me about our S3 setup.

> it takes a while for each to load.

I'm wondering to what extent that's a function of image size. If the image is reasonably small, does it still take a while to load?

---

Just to connect issues, initially I thought this issue was similar to #1718. However, now it seems to me like it's probably something different. #1718 was about server revalidation, whereas I see this issue as probably being more about redirects (304 vs. 307).

alxndrsn:
I think caching is only likely to be possible if either:

1. central's nginx masks the redirect to s3, so to the user the response comes from the ODK Central server, or
2. s3 buckets are expected to be public

lognaturel:
Thank you, that makes sense! Could we possibly mask the redirect just for those well-known assets? 

We’ve decided not to do it by default for all assets because of the overhead it adds but in this case they’re going to be loaded really frequently and likely not change much so overhead at first request is worth it if we can get ongoing caching.

alxndrsn:
> Could we possibly mask the redirect just for those well-known assets?

Do they have predictable URLs?

lognaturel:
Not on S3 but the Central routes are `v1/config/public/logo` and `v1/config/public/hero-image`

alxndrsn:
> Not on S3 but the Central routes are `v1/config/public/logo` and `v1/config/public/hero-image`

Very do-able :+1:



> I think caching is only likely to be possible if either:
> 
[...]

PR #1825, opened 2026-04-21
backend: cache special login images

alxndrsn:
Set long-lived cache headers for special config blobs shown as images on the login page:

* hero image
* logo

Ref https://github.com/getodk/central/issues/1822

Added tests.

It's done in conjunction with preventing clients from seeing the S3 URLs for these images (https://github.com/getodk/central-backend/pull/1810).

The long caching header set here means that the `ETag`/`If-Not-Modified` header is probably never even revalidated in practice.

An alternative would be to leave `odk-central-backend` responsible for this caching header.

That would:

✅ keep the nginx config more consistent
✅ would allow for simpler logic around changing header depending if `ts` is set in the query string
✅ mean this repo does not need to keep track of changes to the config blobs list
❌ would mean re-defining immutable cache headers in the backend repo

Leverages caching to streamline their authentication experience.

No.


test/nginx/src/mocha/nginx.spec.js
```diff
@@ -828,6 +828,14 @@ function standardTestSuite({ fetchHttp, fetchHttp6, apiFetch, apiFetch6, forward
       [ '/fonts/icomoon.ttf',                              'revalidate' ],
       [ '/fonts/icomoon.ttf?',                             'revalidate' ],
       [ '/fonts/icomoon.ttf?ohpk4j',                       'immutable' ],
+
+      // central-backend public images - always requested with timestamp
+      // REVIEW it would be great to ensure that immutable headers are only included when a valid `ts` param is included.  Maybe not realistic tho...
```
matthew-white: One idea is that we could mention in API docs that these endpoints have more caching than most endpoints.

matthew-white: Just thinking to reduce potential nginx complexity while giving users a heads-up about the caching.

alxndrsn: Actually we do the same for fonts already, so I'll update the PR to serve as immutable iff `ts` param is supplied.

alxndrsn:
@matthew-white on reflection I think `odk-central-backend` should be responsible for these headers - I'll add to https://github.com/getodk/central-backend/pull/1810.

matthew-white:
Sounds good, that makes sense. :+1: Would you consider doing it in a follow-up PR after getodk/central-backend#1810 is merged? Since I reviewed that PR already, it feels slightly easier from my perspective for significant additional work to come in a follow-up PR. Though perhaps I'm overestimating how much code will be involved.

matthew-white:
(Not a strong preference, just wanted to throw out this idea of a follow-up PR.)

PR #1857, opened 2026-04-28
nginx/csp: blank.html: allow form-action 'self'

### LINKED ISSUE #1856: allow form-action: 'self' for blank.html
blank.html is specifically used for submitting forms to the backend:

https://github.com/getodk/central-frontend/blob/93c621efafee6c3cd425c021309ee31529798961/src/components/submission/download.vue#L221-L239

alxndrsn:
Closes https://github.com/getodk/central/issues/1856

Updated tests.

Violations were reported to Sentry.  An alternative would be to make the decrypt downloading in central-frontend less arcane.  But that's not likely to be on the cards.

Should prevent incorrect reports.

No.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -193,7 +193,7 @@ server {
     root /usr/share/nginx/html;
     try_files /blank.html =404;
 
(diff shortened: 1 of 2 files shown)


matthew-white (approved):

matthew-white:
> An alternative would be to make the decrypt downloading in central-frontend less arcane.  But that's not likely to be on the cards.

I'm certainly open to it! It doesn't strike me as a priority, but I definitely agree that it's arcane.

PR #1859, opened 2026-04-28
nginx/csp: enforce policy for central-backend

alxndrsn:
Switch from `Content-Security-Policy-Report-Only` to `Content-Security-Policy`.

* Long-term monitoring of CSP reports.
* updated tests

Could leave it off?

Should be low-impact for backend paths - almost none of them except OIDC-related paths are viewed in-browser or expected to have HTML content.

No.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -97,6 +97,11 @@ map $request_uri $central_frontend_csp {
     "default-src 'report-sample' 'none'; connect-src 'self' https://translate.google.com https://translate.googleapis.com; font-src 'self'; form-action 'self'; frame-ancestors 'none'; frame-src 'self' https://getodk.github.io/central/; img-src data: https:; manifest-src 'self'; media-src 'none'; object-src 'none'; script-src 'report-sample' 'self'; style-src 'report-sample' 'self'; style-src-attr 'unsafe-inline'; worker-src 'report-sample' blob:; report-uri /csp-report";
 }
 
+map $upstream_http_content_security_policy $central_backend_csp {
+  ""      "default-src 'report-sample' 'none'; form-action 'none'; frame-ancestors 'none'; report-uri /csp-report";
+  default $upstream_http_content_security_policy;
+}
+
 server {
   listen 443 ssl;
   http2 on;
@@ -176,17 +181,20 @@ server {
   }
   # End of Enketo Configuration.
 
-  location ~ ^/v\d+/oidc/callback$ {
-    include /usr/share/odk/nginx/common-headers.conf;
-    include /usr/share/odk/nginx/backend.conf;
-  }
-
   location ~ ^/v\d {
-    proxy_hide_header Content-Security-Policy-Report-Only;
-    add_header Content-Security-Policy-Report-Only "default-src 'report-sample' 'none'; form-action 'none'; frame-ancestors 'none'; report-uri /csp-report" always;
+    proxy_hide_header Content-Security-Policy;
+    add_header Content-Security-Policy $central_backend_csp always;
 
     include /usr/share/odk/nginx/common-headers.conf;
-    include /usr/share/odk/nginx/backend.conf;
+
+    proxy_set_header X-Forwarded-Proto $scheme;
+    proxy_pass http://service:8383;
+    proxy_redirect off;
+
+    # buffer requests, but not responses, so streaming out works.
+    proxy_request_buffering on;
+    proxy_buffering off;
+    proxy_read_timeout 2m;
   }
 
   location @blank.html {
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -98,6 +98,7 @@ map $request_uri $central_frontend_csp {
 }
 
 map $upstream_http_content_security_policy $central_backend_csp {
+  // pass through any Content-Security-Policy received from upstream services (central-backend, enketo)
   ""      "default-src 'report-sample' 'none'; form-action 'none'; frame-ancestors 'none'; report-uri /csp-report";
   default $upstream_http_content_security_policy;
 }
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -98,7 +98,7 @@ map $request_uri $central_frontend_csp {
 }
 
 map $upstream_http_content_security_policy $central_backend_csp {
-  // pass through any Content-Security-Policy received from upstream services (central-backend, enketo)
+  # pass through any Content-Security-Policy received from upstream services (central-backend, enketo)
   ""      "default-src 'report-sample' 'none'; form-action 'none'; frame-ancestors 'none'; report-uri /csp-report";
   default $upstream_http_content_security_policy;
 }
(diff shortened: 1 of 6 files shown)


files/nginx/odk.conf.template
```diff
@@ -97,6 +97,11 @@ map $request_uri $central_frontend_csp {
     "default-src 'report-sample' 'none'; connect-src 'self' https://translate.google.com https://translate.googleapis.com; font-src 'self'; form-action 'self'; frame-ancestors 'none'; frame-src 'self' https://getodk.github.io/central/; img-src data: https:; manifest-src 'self'; media-src 'none'; object-src 'none'; script-src 'report-sample' 'self'; style-src 'report-sample' 'self'; style-src-attr 'unsafe-inline'; worker-src 'report-sample' blob:; report-uri /csp-report";
 }
 
+map $upstream_http_content_security_policy $central_backend_csp {
```
alxndrsn, line 100:
```suggestion
map $upstream_http_content_security_policy $central_backend_csp {
  // pass through any Content-Security-Policy received from upstream services (central-backend, enketo)
```

test/nginx/mock-http-server/index.js
```diff
@@ -9,23 +9,15 @@ const app = express();
...
-  res.set('Content-Security-Policy-Report-Only', topLevelDirectives.map(d => `${d} NOTE:FROM-BACKEND:reportOnly`).join('; '));
-
   next();
 });
 
 // Enketo express returns response with Vary and Cache-Control headers
 app.use('/-/', (req, res, next) => {
   res.set('Vary', 'Accept-Encoding');
   res.set('Cache-Control', 'public, max-age=0');
+  res.set('Content-Security-Policy',             `NOTE:FROM-BACKEND:block`);
```
alxndrsn, line 21:
```suggestion

  // Set both CSP headers from enketo.  Eventually nginx should be confident to override both.
  res.set('Content-Security-Policy',             `NOTE:FROM-BACKEND:block`);
```

test/nginx/mock-http-server/index.js
```diff
@@ -45,6 +39,13 @@ app.get('/v1/projects', (_, res) => {
   res.send('OK');
 });
 
+app.get('/v1/oidc/callback', (req, res) => {
+  res.set('Content-Security-Policy',             `NOTE:FROM-BACKEND:block`);
```
alxndrsn, line 44:
```suggestion
  // This endpoint is 100% responsible for its own headers.  Set both, and test they both get through.
  res.set('Content-Security-Policy',             `NOTE:FROM-BACKEND:block`);
```

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -260,13 +243,15 @@ describe('Content-Security-Policy definitions', () => {
...
-          it(`should have required directives: ${requiredDirectives}`, () => {
-            assert.containsAllKeys(policy, requiredDirectives);
-          });
+          if(!(typeof policy === 'string' && policy.startsWith('NOTE:FROM-BACKEND:'))) {
+            it(`should have required directives: ${requiredDirectives}`, () => {
+              assert.containsAllKeys(policy, requiredDirectives);
+            });
+          }
 
           Object.entries(policy)
```
matthew-white: Should this right brace be moved further below? If `policy` is a string, is `Object.entries()` called on it right now? I'm thinking that `Object.entries()` should be skipped in the case where `policy` is a string.

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -260,13 +243,15 @@ describe('Content-Security-Policy definitions', () => {
...
               .map    (([ key, directive ]) => [ key, asArray(directive) ])
-              .filter (([ key, directive ]) => !(directive.length === 1 && directive[0] === `NOTE:FROM-BACKEND:${headerType}`)) // eslint-disable-line no-unused-vars
+              .filter (([ key, directive ]) => directive.length !== 1) // eslint-disable-line no-unused-vars
```
matthew-white: This may be related to the comment right above, but I don't totally understand the significance of this change. It seems like before, this line was trying to filter out `NOTE:FROM-BACKEND` values. But are such values even possible here anymore? If not, could this entire line be removed?

matthew-white (approved):

test/nginx/src/mocha/nginx.spec.js
alxndrsn: It shouldn't be a string unless it's a blocked-from-backend string.  Maybe more correct to make them symbols?  Either way, I think it's reasonable for this code to fail if an invalid value gets set in the policies map above.  Or do you think an explicit check would be helpful?

test/nginx/src/mocha/nginx.spec.js
alxndrsn: Good catch - I agree the line should be removed.

```suggestion
```

test/nginx/src/mocha/nginx.spec.js
alxndrsn: Note that symbols would complicate `assertCsp()` slightly.

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -262,13 +245,14 @@ describe('Content-Security-Policy definitions', () => {
...
 
         describe(`header: ${headerNames[headerType]}`, () => {
-          it(`should have required directives: ${requiredDirectives}`, () => {
-            assert.containsAllKeys(policy, requiredDirectives);
-          });
+          if(!(typeof policy === 'string' && policy.startsWith('NOTE:FROM-BACKEND:'))) {
+            it(`should have required directives: ${requiredDirectives}`, () => {
+              assert.containsAllKeys(policy, requiredDirectives);
+            });
+          }
```
```suggestion
          if(typeof policy === 'string') {
            if(!policy.startsWith('NOTE:FROM-BACKEND:')) throw new Error(`Unexpected policy string: '${policy}'`);
          } else {
            it(`should have required directives: ${requiredDirectives}`, () => {
              assert.containsAllKeys(policy, requiredDirectives);
            });
          }
```

test/nginx/src/mocha/nginx.spec.js
alxndrsn: Added explicit check for unexpected strings.

matthew-white: Maybe I'm misreading the code, but even for a `NOTE:FROM-BACKEND:` string, I wouldn't expect `Object.entries()` to be called on that string. The result would be an array like `[[0, 'N'], [1, 'O'], [2, 'T'], [3, 'E'], ...]`. Which I guess doesn't fail, but seems unintended. I'm thinking that `Object.entries()` should be wrapped within the `else` and only run if `policy` is an object.

matthew-white: E.g., currently in the tests I'm seeing things like:

```
    policy: backend-unmodified
      header: Content-Security-Policy
        directive: 0
          ✔ should not include 'report-sample' in directive '0'
        directive: 1
          ✔ should not include 'report-sample' in directive '1'
        directive: 2
          ✔ should not include 'report-sample' in directive '2'
        directive: 3
          ✔ should not include 'report-sample' in directive '3'
```

Which points to `Object.entries()` being called on a string.

alxndrsn: 🤦  now fixed.

Issue #1810, opened 2026-04-14
check-docker-context file count is unstable


alxndrsn:
The differences appear to come from changes in the `.git` dir.

Examples - repeated job on same CI run:

* https://github.com/alxndrsn/odk-central/actions/runs/24387253287/job/71234193262 - 1732 files
* https://github.com/alxndrsn/odk-central/actions/runs/24387253287/job/71224176658 - 1729 files

Differences:

```diff
638,643c638,640
< ./.git/objects/pack/pack-9965b4e17949f2cabc5d30242619a69bc52001c6.idx
< ./.git/objects/pack/pack-9965b4e17949f2cabc5d30242619a69bc52001c6.rev
< ./.git/objects/pack/pack-b6f01a515c37419d293c2192984f5ab33646d880.idx
< ./.git/objects/pack/pack-b6f01a515c37419d293c2192984f5ab33646d880.rev
< ./.git/objects/pack/pack-b6f01a515c37419d293c2192984f5ab33646d880.pack
< ./.git/objects/pack/pack-9965b4e17949f2cabc5d30242619a69bc52001c6.pack
---
> ./.git/objects/pack/pack-3d098af647a8b0a7b34bc9e1175918670d9eeb76.pack
> ./.git/objects/pack/pack-3d098af647a8b0a7b34bc9e1175918670d9eeb76.rev
> ./.git/objects/pack/pack-3d098af647a8b0a7b34bc9e1175918670d9eeb76.idx
791,793c788,790
< ./.git/modules/server/objects/pack/pack-669d6cf2000e949aa7d9b7bc83639f87d899ee90.pack
< ./.git/modules/server/objects/pack/pack-669d6cf2000e949aa7d9b7bc83639f87d899ee90.rev
< ./.git/modules/server/objects/pack/pack-669d6cf2000e949aa7d9b7bc83639f87d899ee90.idx
---
> ./.git/modules/server/objects/pack/pack-dc07faa2e1965d910e769ffd5fbd61ea59079ee1.pack
> ./.git/modules/server/objects/pack/pack-dc07faa2e1965d910e769ffd5fbd61ea59079ee1.rev
> ./.git/modules/server/objects/pack/pack-dc07faa2e1965d910e769ffd5fbd61ea59079ee1.idx
```

alxndrsn:
Some recent examples:

repo | file count | build
---|---|---
this | 1603 | https://github.com/getodk/central/actions/runs/24678708932/job/72170087989
[...]

PR #1863, opened 2026-05-01
nginx: fix escaping in /fonts/ matcher

alxndrsn:
Noted while working on #1810

The question mark is intended as a literal `?` rather than making the `.*` optional.

Seems like a simple change.

No likely effect - such a minor detail, expected URLs would have matched anyway.

No.

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -40,7 +40,7 @@ map "$request_method::$uri$is_args$args" $cache_strategy {
 
   # enketo
   ~^(GET|HEAD)::/-(/x)?/css/                 "revalidate";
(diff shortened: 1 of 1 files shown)


files/nginx/odk.conf.template
```diff
@@ -40,7 +40,7 @@ map "$request_method::$uri$is_args$args" $cache_strategy {
 
   # enketo
   ~^(GET|HEAD)::/-(/x)?/css/                 "revalidate";
-  ~^(GET|HEAD)::/-(/x)?/fonts/.*?v=          "immutable";
+  ~^(GET|HEAD)::/-(/x)?/fonts/.*\?v=          "immutable";
```
```suggestion
  ~^(GET|HEAD)::/-(/x)?/fonts/.*\?v=         "immutable";
```

matthew-white (approved):

PR #1889, opened 2026-05-08
 service: move DB_SSL check back to runtime

alxndrsn:
The `DB_SSL` env var was made illegal in https://github.com/getodk/central/pull/1647.

The check was then moved from runtime to build time in https://github.com/getodk/central/pull/1671.

Checking at build-time allows for faster failure and clearer feedback to sysadmins who are upgrading, and previously depended on this env var.  However, the downside is that if container images are pre-built centrally, this check will be skipped.

With this commit, the check will move to container startup.  However, it will now be skipped if the container is started with a non-standard CMD/command/COMMAND.

- [x] tests
- [x] Checked locally with `docker compose up service`.  Logs as follows:

```
service-1  | 2026-05-08T09:17:20.120152694Z !!!
service-1  | 2026-05-08T09:17:20.120169026Z !!! You have the DB_SSL variable defined (in your .env file, probably).
service-1  | 2026-05-08T09:17:20.120172803Z !!! This variable is no longer supported from Central 2026.1 onwards.
service-1  | 2026-05-08T09:17:20.120183784Z !!! There is a new way of configuring SSL for your database, please see:
service-1  | 2026-05-08T09:17:20.120185848Z !!!
service-1  | 2026-05-08T09:17:20.120187712Z !!!   https://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server
service-1  | 2026-05-08T09:17:20.120189726Z !!!
service-1  | 2026-05-08T09:17:20.120191429Z !!! Please refer to the Central 2026.1.0 release notes for more information on this change.
service-1  | 2026-05-08T09:17:20.120193163Z !!!
service-1  | 2026-05-08T09:17:20.120194886Z !!! ODK Central backend will not start until this issue is resolved.
service-1  | 2026-05-08T09:17:20.120197351Z !!!
service-1 exited with code 1 (restarting)
service-1  | 2026-05-08T09:18:20.387388287Z !!!
service-1  | 2026-05-08T09:18:20.387400901Z !!! You have the DB_SSL variable defined (in your .env file, probably).
service-1  | 2026-05-08T09:18:20.387404328Z !!! This variable is no longer supported from Central 2026.1 onwards.
service-1  | 2026-05-08T09:18:20.387407274Z !!! There is a new way of configuring SSL for your database, please see:
service-1  | 2026-05-08T09:18:20.387410039Z !!!
service-1  | 2026-05-08T09:18:20.387413055Z !!!   https://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server
service-1  | 2026-05-08T09:18:20.387416000Z !!!
service-1  | 2026-05-08T09:18:20.387420369Z !!! Please refer to the Central 2026.1.0 release notes for more information on this change.
service-1  | 2026-05-08T09:18:20.387423204Z !!!
service-1  | 2026-05-08T09:18:20.387425910Z !!! ODK Central backend will not start until this issue is resolved.
service-1  | 2026-05-08T09:18:20.387428705Z !!!
service-1 exited with code 1 (restarting)
service-1  | 2026-05-08T09:19:20.651362541Z !!!
service-1  | 2026-05-08T09:19:20.651374394Z !!! You have the DB_SSL variable defined (in your .env file, probably).
service-1  | 2026-05-08T09:19:20.651377179Z !!! This variable is no longer supported from Central 2026.1 onwards.
service-1  | 2026-05-08T09:19:20.651379454Z !!! There is a new way of configuring SSL for your database, please see:
service-1  | 2026-05-08T09:19:20.651381638Z !!!
service-1  | 2026-05-08T09:19:20.651383802Z !!!   https://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server
service-1  | 2026-05-08T09:19:20.651386167Z !!!
service-1  | 2026-05-08T09:19:20.651388321Z !!! Please refer to the Central 2026.1.0 release notes for more information on this change.
service-1  | 2026-05-08T09:19:20.651390545Z !!!
service-1  | 2026-05-08T09:19:20.651392709Z !!! ODK Central backend will not start until this issue is resolved.
service-1  | 2026-05-08T09:19:20.651394944Z !!!
service-1 exited with code 1 (restarting)
service-1  | 2026-05-08T09:20:20.934397647Z !!!
service-1  | 2026-05-08T09:20:20.934408808Z !!! You have the DB_SSL variable defined (in your .env file, probably).
service-1  | 2026-05-08T09:20:20.934411524Z !!! This variable is no longer supported from Central 2026.1 onwards.
service-1  | 2026-05-08T09:20:20.934413768Z !!! There is a new way of configuring SSL for your database, please see:
service-1  | 2026-05-08T09:20:20.934417155Z !!!
service-1  | 2026-05-08T09:20:20.934419359Z !!!   https://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server
service-1  | 2026-05-08T09:20:20.934421994Z !!!
service-1  | 2026-05-08T09:20:20.934424208Z !!! Please refer to the Central 2026.1.0 release notes for more information on this change.
service-1  | 2026-05-08T09:20:20.934426433Z !!!
service-1  | 2026-05-08T09:20:20.934428627Z !!! ODK Central backend will not start until this issue is resolved.
service-1  | 2026-05-08T09:20:20.934430781Z !!!
service-1 exited with code 1 (restarting)
service-1  | 2026-05-08T09:21:21.207766916Z !!!
service-1  | 2026-05-08T09:21:21.207780743Z !!! You have the DB_SSL variable defined (in your .env file, probably).
service-1  | 2026-05-08T09:21:21.207784450Z !!! This variable is no longer supported from Central 2026.1 onwards.
service-1  | 2026-05-08T09:21:21.207787266Z !!! There is a new way of configuring SSL for your database, please see:
service-1  | 2026-05-08T09:21:21.207790021Z !!!
service-1  | 2026-05-08T09:21:21.207792776Z !!!   https://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server
service-1  | 2026-05-08T09:21:21.207795752Z !!!
service-1  | 2026-05-08T09:21:21.207798507Z !!! Please refer to the Central 2026.1.0 release notes for more information on this change.
service-1  | 2026-05-08T09:21:21.207801213Z !!!
service-1  | 2026-05-08T09:21:21.207803938Z !!! ODK Central backend will not start until this issue is resolved.
```

Checking at build-time allows for faster failure and clearer feedback to sysadmins who are upgrading, and previously depended on this env var. However, the downside is that if container images are pre-built centrally, this check will be skipped.

This check will also now be skipped if the container is started with a non-standard CMD/command/COMMAND.

An alternative might be to include the check into `files/service/with-pgenvblock.pl`, which IIUC is intended to be run in any situation where env vars should be injected into a container command.

I don't think this should affect users.

I don't think so.

### DIFF
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -3,7 +3,7 @@ set -o pipefail
 shopt -s inherit_errexit
 
 # Check for illegal DB_SSL environment variable.
-if ! [[ "${DB_SSL-}" = notset ]]; then
+if ! [[ "${DB_SSL-}" = null ]]; then
   echo "!!!"
   echo "!!! You have the DB_SSL variable defined (in your .env file, probably)."
   echo "!!! This variable is no longer supported from Central 2026.1 onwards."
@@ -18,7 +18,7 @@ if ! [[ "${DB_SSL-}" = notset ]]; then
   sleep 60 # reduce resource waste from quick restart and instant failure
   exit 1
 fi
-unset DB_SSL_CHECK
+unset DB_SSL
 
 # Serialize (as a raw env block) the environment set up by docker, for later
 # availability to processes running with a reset environment (such as cronjobs).
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -2,6 +2,24 @@
 set -o pipefail
 shopt -s inherit_errexit
 
+# Check for illegal DB_SSL environment variable.
+if ! [[ "${DB_SSL-}" = notset ]]; then
+  echo "!!!"
+  echo "!!! You have the DB_SSL variable defined (in your .env file, probably)."
+  echo "!!! This variable is no longer supported from Central 2026.1 onwards."
+  echo "!!! There is a new way of configuring SSL for your database, please see:"
+  echo "!!!"
+  echo "!!!   https://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server"
+  echo "!!!"
+  echo "!!! Please refer to the Central 2026.1.0 release notes for more information on this change."
+  echo "!!!"
+  echo "!!! ODK Central backend will not start until this issue is resolved."
+  echo "!!!"
+  sleep 60 # reduce resource waste from quick restart and instant failure
+  exit 1
+fi
+unset DB_SSL_CHECK
+
 # Serialize (as a raw env block) the environment set up by docker, for later
 # availability to processes running with a reset environment (such as cronjobs).
 # See https://github.com/getodk/central/issues/1747 .
--- a/service.dockerfile
+++ b/service.dockerfile
@@ -15,8 +15,7 @@ RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODEN
     && curl https://www.postgresql.org/media/keys/ACCC4CF8.asc \
       | gpg --dearmor > /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg
 
-ARG DB_SSL
-RUN [ -z "${DB_SSL}" ] || (/bin/echo -e '\n\n\n\n\nYou have the DB_SSL variable defined (in your .env file, probably).\nThis variable is no longer supported from Central 2026.1 onwards.\nThere is a new way of configuring SSL for your database, please see:\n\nhttps://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server\n\nPlease refer to the Central 2026.1.0 release notes for more information on this change.\n\n\n\n\n'; exit 13)
+
 
 FROM node:${node_version}-slim AS intermediate
 RUN apt-get update \
--- a/test/service.spec.js
+++ b/test/service.spec.js
@@ -16,7 +16,7 @@ describe('service image', () => {
       process.stdout.on('data', appendOutput);
       process.stderr.on('data', appendOutput);
 
-      const timer = setTimeout(() => { process.kill(); }, 4_000);
+      const timer = setTimeout(() => { process.kill(); }, 8_000);
 
       process.on('close', (code, signal) => {
         clearTimeout(timer);
@@ -38,7 +38,7 @@ describe('service image', () => {
     });
 
     it('should reject DB_SSL=true', async function() {
-      this.timeout(5000);
+      this.timeout(10_000);
 
       // when
       const { stdcombi } = await runService('--env', 'DB_SSL=true');
@@ -49,7 +49,7 @@ describe('service image', () => {
     });
 
     it('should start OK if DB_SSL is not set', async function() {
-      this.timeout(5000);
+      this.timeout(10_000);
 
       // when
       const { stdcombi } = await runService();
--- a/test/service.spec.js
+++ b/test/service.spec.js
@@ -28,6 +28,7 @@ describe('service image', () => {
     before(function () {
       this.timeout(120_000);
       log('Building "service" docker image...');
+      execSync('touch .env', { cwd:'..', stdio:['ignore', 'inherit', 'inherit'] });
       execSync('docker compose build service', { cwd:'..', stdio:['ignore', 'inherit', 'inherit'] });
       log('"service" docker image built OK.');
     });
--- a/test/service.spec.js
+++ b/test/service.spec.js
@@ -8,20 +8,20 @@ const log = (...args) => console.log(logPrefix, ...args);
 describe('service image', () => {
   describe('DB_SSL handling', () => {
     const runService = (...args) => new Promise(resolve => {
-      const stdout = [];
-      const stderr = [];
+      const stdcombi = [];
 
       const process = spawn('docker', [ 'compose', 'run', ...args, 'service' ], { cwd:'..' });
 
-      process.stdout.on('data', data => stdout.push(data.toString()));
-      process.stderr.on('data', data => stderr.push(data.toString()));
+      const appendOutput = data => stdcombi.push(data.toString());
+      process.stdout.on('data', appendOutput);
+      process.stderr.on('data', appendOutput);
 
       const timer = setTimeout(() => { process.kill(); }, 2_000);
 
       process.on('close', (code, signal) => {
         clearTimeout(timer);
         const asLines = datas => datas.join('').split('\n');
-        resolve({ code, signal, stdout:asLines(stdout), stderr:asLines(stderr) });
+        resolve({ code, signal, stdcombi:asLines(stdcombi) });
       });
     });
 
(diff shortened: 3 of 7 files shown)


garethbowen (approved):

files/service/scripts/start-odk.sh
```diff
@@ -2,6 +2,24 @@
 set -o pipefail
 shopt -s inherit_errexit
 
+# Check for illegal DB_SSL environment variable.
+if ! [[ "${DB_SSL-}" = notset ]]; then
```
matthew-white:
```suggestion
if [[ -n "${DB_SSL-}" ]]; then
```

I don't quite understand the intention behind `notset` as the default value of `DB_SSL`. I don't recall us using that pattern elsewhere. Could we make the default value an empty string, then check here for any non-empty string?

files/service/scripts/start-odk.sh
```diff
@@ -2,6 +2,24 @@
...
+  echo "!!!   https://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server"
+  echo "!!!"
+  echo "!!! Please refer to the Central 2026.1.0 release notes for more information on this change."
+  echo "!!!"
+  echo "!!! ODK Central backend will not start until this issue is resolved."
+  echo "!!!"
+  sleep 60 # reduce resource waste from quick restart and instant failure
+  exit 1
+fi
+unset DB_SSL_CHECK
```
matthew-white:
```suggestion
unset DB_SSL
```

Should this be `DB_SSL`? I don't see any other reference in this PR to `DB_SSL_CHECK`.

matthew-white (approved):
I'll leave an additional comment soon about some of the background context here, but it makes sense to me that this needs to become a runtime check. :+1:

matthew-white:
Here's my understanding of the history here:

- Users used to use `DB_SSL` to configure SSL for their external database.
- In the last release, we removed/replaced `DB_SSL`.
- We didn't want users to think they had SSL configured when they didn't, or for users to specify conflicting configuration, so we decided to error out when `DB_SSL` was set.
- I think Wicher initially checked just for `DB_SSL=true`, but I suggested that we check for any `DB_SSL` value. Going forward, it's not expected that `DB_SSL` would ever be set.
- I don't recall exactly why we decided to check it at build time rather than runtime. But given that the variable no longer drives app behavior, I don't think it matters too much.
- Here are some potential small advantages to checking at build time:
  - As @alxndrsn said, "Checking at build-time allows for faster failure and clearer feedback to sysadmins who are upgrading"
  - Not all sysadmin interactions with `service` necessarily run through start-odk.sh. E.g., if you `docker compose exec service ...`, the check in start-odk.sh won't trip. As @alxndrsn wrote, the check "will now be skipped if the container is started with a non-standard CMD/command/COMMAND."
    - Though I think you can also change the .env (or much of it) without rebuilding, so I don't think the build time check was foolproof either.
  - The new runtime check happens before `cron` is started, so I don't think build time vs. runtime matters for our cron jobs / our Backend "tasks".
- Those potential small advantages are definitely outweighed by the desire to prebuild images. :+1:

files/service/scripts/start-odk.sh
alxndrsn: The intention related to the special property of `DB_SSL`: when it has an empty value, that implies something different than if it's not set.  I'll double-check this and add an explanatory commend and/or appropriate tests.

files/service/scripts/start-odk.sh
alxndrsn: :facepalm: good catch - this was from a previous iteration.

Another sign this might benefit from a test or two.

files/service/scripts/start-odk.sh
alxndrsn: Original code at coerced unset values to `null`:

https://github.com/getodk/central/pull/1647/changes#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3L60 

I've reintroduced this code.  I've also added tests for different possible values of `DB_SSL` - the only acceptable values should now be `"null"` or completely unset.  Other values would either enable SSL, or would break `getodk/central-backend`.

files/service/scripts/start-odk.sh
alxndrsn: Fixed.

service.dockerfile
```diff
@@ -15,8 +15,7 @@ RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(grep -oP 'VERSION_CODEN
     && curl https://www.postgresql.org/media/keys/ACCC4CF8.asc \
       | gpg --dearmor > /etc/apt/trusted.gpg.d/apt.postgresql.org.gpg
 
-ARG DB_SSL
-RUN [ -z "${DB_SSL}" ] || (/bin/echo -e '\n\n\n\n\nYou have the DB_SSL variable defined (in your .env file, probably).\nThis variable is no longer supported from Central 2026.1 onwards.\nThere is a new way of configuring SSL for your database, please see:\n\nhttps://docs.getodk.org/central-install-digital-ocean/#using-a-custom-database-server\n\nPlease refer to the Central 2026.1.0 release notes for more information on this change.\n\n\n\n\n'; exit 13)
+
```
alxndrsn, line 18: Whitespace reintroduced to separate image definitions.

files/service/scripts/start-odk.sh
matthew-white: > the only acceptable values should now be `"null"` or completely unset. Other values would either enable SSL, or would break `getodk/central-backend`.

This strategy of using `null` sounds reasonable to me. :+1: But just to clarify, `DB_SSL` really doesn't drive app behavior in any part of the stack. Nothing in ops is enabling or disabling SSL based on `DB_SSL`. The Backend config JSON used to have an `ssl` property within the `database` object, but that's no longer the case. Backend no longer has special logic to enable database SSL. That's all done now via `PG*` environment variables, e.g., `PGSSLMODE`. The only reason we're checking `DB_SSL` is that the variable used to drive app behavior. We want to make sure that users aren't setting `DB_SSL` with the intention of enabling SSL when in reality is has no effect.

> The intention related to the special property of `DB_SSL`: when it has an empty value, that implies something different than if it's not set.

Back when `DB_SSL` drove behavior, I think empty value and not-set had the same effect. That's because `DB_SSL=${DB_SSL:-null}` would set `DB_SSL` to `'null'` when it was either empty or not-set.

For some of the `PG*` environment variables, the difference between empty and not-set can be significant. That's why we use `-` instead of `:-` with the `PG*` environment variables, e.g., `PGPASSWORD=${PGPASSWORD-${DB_PASSWORD:-odk}}`. But that's due to Postgres behavior, so I don't think that applies to `DB_SSL`.

test/service.spec.js
```diff
@@ -0,0 +1,118 @@
...
+
+      log('Building "service" docker image...');
+      exec('touch .env');
+      exec('docker compose pull --include-deps service');
+      exec('docker compose build --with-dependencies service');
+      log('"service" docker image built OK.');
+    });
+
+    [
+      '', // e.g. if passed via `docker compose run --env DB_SSL=`
```
matthew-white, line 68: Ooh interesting, so an empty value of `DB_SSL` is still possible via `--env`, despite the use of `:-` in `DB_SSL=${DB_SSL:-null}`. That sounds reasonable to me to error out in that case.

matthew-white:
I need to take a closer look at the new tests, but the approach here is looking good to me. :+1: I'll take a look at those tests ASAP.

test/service.spec.js
```diff
@@ -0,0 +1,118 @@
...
+        clearTimeout(timer);
+        const asLines = datas => datas.join('').split('\n');
+        resolve({ code, signal, stdcombi:asLines(stdcombi) });
+      });
+    });
+
+    before(function () {
+      this.timeout(120_000);
+
+      const exec = cmd => execSync(cmd, { cwd:'..', stdio:['ignore', 'inherit', 'inherit'] });
```
matthew-white, line 58: Just for my own understanding, why is `stdio:['ignore', 'inherit', 'inherit']` needed? Why does the default `stdio` option not work?

matthew-white (approved):
I love the tests!

files/service/scripts/start-odk.sh
matthew-white: > Other values would either enable SSL, or would break `getodk/central-backend`.

Oh I see, you're saying that it worked that way in the past. Sorry, for some reason I thought you were saying that it was the current behavior.

test/service.spec.js
alxndrsn: `'inherit'` means that the outputs of these commands are seen in the tests' `stdout` rather than sent to a NodeJS buffer.  Without extra code, this allows:

1. progress to be observed when running slow docker commands, and
2. errors to be seen

PR #1893, opened 2026-05-14
Fixes: expected docker context is increased due to WF merge

sadiqkhoja:
Size of git db (objects) has increase quite a bit after merging web-forms code into central-frontend, therefore increasing expected file size of docker context.

I am wondering if there is a way not to include git objects in the docker context but we rely on them to generate version.txt. Maybe once we start publishing built images we can get rid of them.

_PS: Found this issue while working on a separate issue. To understand the root cause I rerun GH Actions on `next` and `master` branches even those were failing with everything (OS, runner, git version, docker version) being same e.g. https://github.com/getodk/central/actions/runs/25595119221/job/76010520926. Only difference I could see was `git/objects/pack/pack*`._ 

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,7 @@ jobs:
         fetch-depth: 0
         fetch-tags: true
         submodules: recursive
-    - run: ./test/check-docker-context.sh --min-size 50000 --max-size 60000 --min-count 1500 --max-count 1700
+    - run: ./test/check-docker-context.sh --min-size 50000 --max-size 100000 --min-count 1500 --max-count 1700
     - run: ./test/test-images.sh
     - if: always()
       run: docker compose logs


lognaturel (approved):
Approving to fix CI in the short term and because I agree we should review tagging and version generation as part of what @alxndrsn and @garethbowen are working on to pre-build images.

An alternate small fix would be to set the fetch depth to something like 50 instead of 0 (unlimited). That should pull the tagged commits we need without increasing the download size so much.

lognaturel:
We've talked about this plenty but it's another example of how our installs are not exactly reproducible. CC @yanokwa

alxndrsn:
Seems like the new file count window is very big.

sadiqkhoja:
Talked with @alxndrsn on Slack, and he is suggesting that we increase the min-size value as well so that expected size window is not too large. I'll create a new PR to do that.

PR #1925, opened 2026-05-31
dev/docker-compose: restrict open ports to local machine

alxndrsn:
Restrict TCP ports to the local machine.  This will prevent exposing these dev services on the local network or wider.

- [x] tested locally

Should not change dev workflows, while resolving an issue.

No effect - just dev.

No.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -11,7 +11,7 @@ services:
     profiles:
       - central
     ports:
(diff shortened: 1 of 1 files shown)


sadiqkhoja (approved):
Looks good to me. 

Verified by running this branch locally, and I don't see any disruption in my workflow 🙏

PR #1927, opened 2026-05-31
test/nginx/docker-compose: restrict open ports to local machine

alxndrsn:
Restrict TCP ports to the local machine.  This will prevent exposing these dev services on the local network or wider.

- [x] CI
- [x] tested locally

Should not change dev workflows, while resolving an issue.

Ideally the `sentry-mock` port would also be bound to `127.0.0.1`.  Unfortunately the related code is quite fiddly, so instead of resolving the issue a `FIXME` has been added.

No effect - just dev.

No.


test/nginx/nginx.test.docker-compose.yml
```diff
@@ -3,21 +3,22 @@ services:
...
-      - "8383:8383"
+      - 127.0.0.1:8383:8383
     environment:
       - PORT=8383
   sentry-mock:
     build:
       dockerfile: mock-sentry.dockerfile
     ports:
       # Sentry port is not currently configurable in nginx config, so use default HTTPS port
+      # FIXME in line with other .ports, make this port binding more restrictive
```
matthew-white, line 21: You wrote in the PR description:

> Ideally the `sentry-mock` port would also be bound to `127.0.0.1`.  Unfortunately the related code is quite fiddly, so instead of resolving the issue a `FIXME` has been added.

Adding a `FIXME` for this one while making progress on other ports sounds good to me. What do you think makes sense in terms of resolving the `FIXME`? Are you thinking to try to resolve it soon? Or should we file an issue to keep track of it?

test/nginx/nginx.test.docker-compose.yml
```diff
@@ -30,14 +31,18 @@ services:
     environment:
       - SSL_TYPE=selfsign
     ports:
-      - "9000:80"
-      - "9001:443"
+      - "127.0.0.1:9000:80"
+      - "[::1]:9000:80"
```
matthew-white, line 35: Why are both 127.0.0.1 and [::1] needed here, whereas above only 127.0.0.1 was needed?

alxndrsn: There are explicit tests for the nginx config to ensure it serves over IPv6.  `service` and `enketo` are not tested in this way.

test/nginx/nginx.test.docker-compose.yml
alxndrsn: I wasn't planning to prioritise fixing it.  Maybe `FIXME` is too strong.

matthew-white: Is it just not an issue to expose the `sentry-mock` on a wider network? Is that because it's pretty minimal / doesn't do much?

> Unfortunately the related code is quite fiddly

I think I remember some of that fiddliness, so I definitely see the appeal of avoiding a revisit to some of those tests.

alxndrsn: > Is it just not an issue to expose the `sentry-mock` on a wider network?

It's an issue, but it already exists, and the fix will be more complicated than for all the other exposed services.  This PR improves on the status quo for very little effort, and defers more complicated work which is independent of these changes.

matthew-white (approved):

PR #1937, opened 2026-06-03
nginx: specify nginx version

### LINKED ISSUE #1936: [nginx] docker base image can transparently change nginx version
A confusing detail of work on https://github.com/getodk/central/pull/1737 is that the current "pinned" nginx base docker image can actually change nginx version over time.

Currently `nginx.dockerfile` pins the nginx base image as:

https://github.com/getodk/central/blob/6b1ac3b8a8f4b21c74eda22d4773a332a4c724e3/nginx.dockerfile#L16-L18

But per https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/cd4a0c825ecd605b568e8a1103c2cb6418a161b1/docs/dockerhub_tags.md?plain=1#L24-L26 this can pull in any nginx version from:

* `1.29.5`
* `1.29.6`
* `1.29.7`

Building from a more specific base image would provide less random surprises, e.g.

```Dockerfile
FROM jonasal/nginx-certbot:6.1.0-nginx1.29.7
```

alxndrsn:
Closes #1936

✅ CI

Reduces nginx changes silently affecting production behaviour.

Reduces drift in nginx versions between same ODK Central versions.  This should reduce risk.

I don't think so.

### DIFF
--- a/nginx.dockerfile
+++ b/nginx.dockerfile
@@ -13,9 +13,14 @@ RUN files/prebuild/build-frontend.sh
 
 
 
-# when upgrading, look for upstream changes to redirector.conf
-# also, confirm setup-odk.sh strips out HTTP-01 ACME challenge location
-FROM jonasal/nginx-certbot:6.1.0
+# When upgrading:
+#
+# 1. Use full-length tag, including nginx version.  See:
+#    * https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/master/docs/dockerhub_tags.md
+#    * https://hub.docker.com/r/jonasal/nginx-certbot/tags
+# 2. Look for upstream changes to redirector.conf
+# 3. Confirm setup-odk.sh strips out HTTP-01 ACME challenge location.
+FROM jonasal/nginx-certbot:6.1.0-nginx1.29.7
 
 EXPOSE 80
 EXPOSE 443


matthew-white:
I'm thinking that it makes sense for @sadiqkhoja to review this one, since he's the one who typically updates nginx (e.g., as part of #1805).

nginx.dockerfile
```diff
@@ -13,9 +13,14 @@ RUN files/prebuild/build-frontend.sh
...
-# also, confirm setup-odk.sh strips out HTTP-01 ACME challenge location
-FROM jonasal/nginx-certbot:6.1.0
+# When upgrading:
+#
+# 1. Use full-length tag, including nginx version.  See:
+#    * https://github.com/JonasAlfredsson/docker-nginx-certbot/blob/master/docs/dockerhub_tags.md
+#    * https://hub.docker.com/r/jonasal/nginx-certbot/tags
+# 2. Look for upstream changes to redirector.conf
+# 3. Confirm setup-odk.sh strips out HTTP-01 ACME challenge location.
+FROM jonasal/nginx-certbot:6.1.0-nginx1.29.7
```
lognaturel, line 23: ~Servers that have rebuilt recently would have been running `-nginx1.31.1`. I think we should go straight to that tag since some production users already have it.~ 

Whoops, I see 6.2.0 straddles nginx 1.29 and 1.31 but 6.1.0 does not. So we could stay on 6.1.0 for now and this would be the correct pin. Then we could look at actually upgrading later based on what the latest version is at that time.

lognaturel:
@sadiqkhoja I added some notes at https://github.com/getodk/central/issues/1936#issuecomment-4615383935 that you may find to be helpful context.

sadiqkhoja (approved):

PR #1947, opened 2026-06-04
nginx: reject form previews with unexpected query params

alxndrsn:

New tests, existing tests.

`$args` matching seems like a blunt implement, but nginx config is also quite limiting.

No expected effect.

No.

### DIFF
--- a/test/nginx/src/mocha/nginx.spec.js
+++ b/test/nginx/src/mocha/nginx.spec.js
@@ -1,4 +1,3 @@
-const { isIPv6 } = require('node:net');
 const tls = require('node:tls');
 const { Readable } = require('stream');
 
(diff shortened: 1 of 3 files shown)


test/nginx/src/mocha/nginx.spec.js
```diff
@@ -1,3 +1,4 @@
+const { isIPv6 } = require('node:net');
```
alxndrsn: revert

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -1129,8 +1231,7 @@ function request(url, { body, ...options }={}) {
   });
 }
 
-function getProtocolImplFrom(url) {
-  const { protocol } = new URL(url);
+function getProtocolImplFrom({ protocol }) {
```
alxndrsn: revert

lognaturel (approved):

PR #1949, opened 2026-06-04
test/nginx: prevent request() from normalising paths

alxndrsn:
This will be useful for testing nginx config at some point, and very misleading if trying to test path normalisation differences and `request()` collapses path traversals.

* existing tests pass
* checked with new tests relating to collapsed path-traversal

While not immediately used, this will be useful for testing nginx config at some point, and very misleading if trying to test path normalisation differences and `request()` collapses path traversals.

No effect.

No.

### DIFF
--- a/test/nginx/src/mocha/nginx.spec.js
+++ b/test/nginx/src/mocha/nginx.spec.js
@@ -1,3 +1,4 @@
+const { isIPv6 } = require('node:net');
 const tls = require('node:tls');
 const { Readable } = require('stream');
 
@@ -1087,13 +1088,20 @@ async function resetMock(port) {
 //
 // 1. do not follow redirects
 // 2. allow overriding of fetch's "forbidden" headers: https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name
-function request(url, { body, ...options }={}) {
+function request(urlString, { body, ...options }={}) {
   if(!options.headers) options.headers = {};
   if(!options.headers.host) options.headers.host = 'odk-nginx.example.test';
 
+  const url = new URL(urlString);
+  if(url.username || url.password) throw new Error('Basic auth creds not yet supported.');
+
+  options.host = safeIpv6(url.hostname);
+  options.port = url.port;
+  options.path = urlString.replace(/^http(s?):\/\/[^/]*/, '') || '/';
+
   return new Promise((resolve, reject) => {
     try {
-      const req = getProtocolImplFrom(url).request(url, options, res => {
+      const req = getProtocolImplFrom(url).request(options, res => {
         res.on('error', reject);
 
         const body = new Readable({ read:() => {} });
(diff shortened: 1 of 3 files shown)


test/nginx/src/mocha/nginx.spec.js
```diff
@@ -1087,13 +1088,20 @@ async function resetMock(port) {
...
+function request(urlString, { body, ...options }={}) {
   if(!options.headers) options.headers = {};
   if(!options.headers.host) options.headers.host = 'odk-nginx.example.test';
 
+  const url = new URL(urlString);
+  if(url.username || url.password) throw new Error('Basic auth creds not yet supported.');
+
+  options.host = safeIpv6(url.hostname);
+  options.port = url.port;
+  options.path = urlString.replace(/^http(s?):\/\/[^/]*/, '') || '/';
```
matthew-white: What do you think about using `url.pathname` here instead? Mostly just thinking for code clarity. Or does `pathname` do its own path normalization?

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -1087,13 +1088,20 @@ async function resetMock(port) {
...
 // 2. allow overriding of fetch's "forbidden" headers: https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name
-function request(url, { body, ...options }={}) {
+function request(urlString, { body, ...options }={}) {
   if(!options.headers) options.headers = {};
   if(!options.headers.host) options.headers.host = 'odk-nginx.example.test';
 
+  const url = new URL(urlString);
+  if(url.username || url.password) throw new Error('Basic auth creds not yet supported.');
+
+  options.host = safeIpv6(url.hostname);
```
matthew-white: How about adding a comment here explaining that the goal in adding these properties to `options` is to prevent path normalization?

matthew-white (approved):

test/nginx/src/mocha/nginx.spec.js
alxndrsn:
```js
> new URL('http://example.test/a/../b').pathname
'/b'
```

But I think the wider point is that this code should be tested!  Adding...

alxndrsn: Tests for `request()` added at https://github.com/getodk/central/pull/1955.

After that's merged, this PR can be reconsidered.

alxndrsn: Additionally:

```js
> new URL('http://example.test/?x=1').pathname
'/'
```

So `.pathname` is unsuitable because:

* it normalises the path
* it strips query params

test/nginx/src/mocha/nginx.spec.js
alxndrsn: Extracted to a separate function and documented.

PR #1977, opened 2026-06-15
test/nginx: remove comment re: web-forms paths

alxndrsn:
The linked PR is very long, and doesn't give clear guidance/explanation of what's going on.

Ultimately this test should be the source of truth for the paths which serve web-forms content, and more context can be added here if required.

It's just a comment update.

Ultimately this test should be the source of truth for the paths which serve web-forms content, and more context can be added here if required.

No effect.

No.


lognaturel (approved):
Agreed.

PR #1980, opened 2026-06-15
nginx/web-forms: assert correct HTML is served

alxndrsn:

- [x] tested locally
- [x] tested in CI

Nginx's behaviour in this context is very opaque, and more accurate testing is very helpful to prevent surprises in production.

No effect expected.

No.


latin-panda (approved):
Approving: Low risk. CI tests are passing. Follows the same line of testing central (id=app).

PR #1998, opened 2026-06-16
client: replace submodule with specific release

alxndrsn:
Closes https://github.com/getodk/central/issues/1996

#### What has been done to verify that this works as intended?

- [x] tested locally
- [x] tested in CI
- [x] ~tested with dev builds~ (not relevant)
- [x] tested to ensure submodule is not _deleted_ when switching branches
```sh
$ git checkout next && git submodule update --init --recursive && ls client && git checkout no-more-submodule && git submodule update --init --recursive && ls client
M	server
Switched to branch 'next'
Your branch is up to date with 'upstream/next'.
Submodule path 'client': checked out '56b6bbdb4cac59aa850e3ddd1cd19ab9f2be1af3'
Submodule path 'server': checked out '142c1c85501d90d225116862ba79dd914c5fcfc1'
bin              docs       icomoon.json  jsconfig.json  LICENSE     NOTICE        package-lock.json  public     SECURITY.md  test       vite.config.js
CONTRIBUTING.md  e2e-tests  index.html    karma.conf.js  nginx-conf  package.json  Procfile           README.md  src          transifex  vue.config.js
warning: unable to rmdir 'client': Directory not empty
M	server
Switched to branch 'no-more-submodule'
Your branch is up to date with 'origin/no-more-submodule'.
Submodule path 'server': checked out 'd36403996af4abca303db4d4e2dbc7b4425f972d'
bin              docs       icomoon.json  jsconfig.json  LICENSE     NOTICE        package-lock.json  public     SECURITY.md  test       vite.config.js
CONTRIBUTING.md  e2e-tests  index.html    karma.conf.js  nginx-conf  package.json  Procfile           README.md  src          transifex  vue.config.js
```

#### Why is this the best possible solution? Were any other approaches considered?

* :+1: very fast, low overhead container builds
* :+1: allows local edits 
* :+1: doesn't delete local changes in the `client/` directory
* :+1: allows arbitrary treeish refs

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have accidentally been affected by code changes. In other words, what are the regression risks?

* :+1: should help 3rd-party deployers who don't edit `client`/`central-frontend` code
* :+1: should not confuse 3rd-party deployers who do edit `client`/`central-frontend` code, although they will need to set a new var in `docker-compose.yml` when updating

#### Does this change require updates to documentation? If so, please file an issue [here](https://github.com/getodk/docs/issues/new) and include the link below.

[...]

PR #2005, opened 2026-06-18
chore: pass frontend Sentry DSN

latin-panda:
Follow-up from this work: https://github.com/getodk/web-forms/issues/803 ([PR#1610](https://github.com/getodk/central-frontend/pull/1610))

Unit tests are passing. Docker is built with the correct client-config.json

It follows the same pattern as other variables are passed.

Documented in the README file.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -44,7 +44,6 @@ HTTPS_PORT=443
 # SENTRY_TRACE_RATE=
 
 # Optional: configure frontend error reporting
-# Set to your Sentry DSN before `docker compose build` to embed it in the frontend bundle.
 # ODK_CENTRAL_FRONTEND_SENTRY_DSN=
 
 # Optional: configure S3-compatible storage for binary files
--- a/.env.template
+++ b/.env.template
@@ -43,6 +43,10 @@ HTTPS_PORT=443
 # SENTRY_PROJECT=
 # SENTRY_TRACE_RATE=
 
+# Optional: configure frontend error reporting
+# Set to your Sentry DSN before `docker compose build` to embed it in the frontend bundle.
+# ODK_CENTRAL_FRONTEND_SENTRY_DSN=
+
 # Optional: configure S3-compatible storage for binary files
 # S3_SERVER=
 # S3_ACCESS_KEY=
--- a/test/nginx/lib.docker-compose.yml
+++ b/test/nginx/lib.docker-compose.yml
@@ -17,6 +17,7 @@ services:
       - SENTRY_ORG_SUBDOMAIN=o-fake-dsn
       - SENTRY_PROJECT=example-sentry-project
       - OIDC_ENABLED=false
+      - ODK_CENTRAL_FRONTEND_SENTRY_DSN=
     volumes:
       - ../../files/nginx/odk.conf.template:/usr/share/odk/nginx/odk.conf.template:ro
       - ../../files/nginx/client-config.json.template:/usr/share/odk/nginx/client-config.json.template:ro
--- a/test/nginx/lib.docker-compose.yml
+++ b/test/nginx/lib.docker-compose.yml
@@ -17,7 +17,7 @@ services:
       - SENTRY_ORG_SUBDOMAIN=o-fake-dsn
       - SENTRY_PROJECT=example-sentry-project
       - OIDC_ENABLED=false
-      - ODK_CENTRAL_FRONTEND_SENTRY_DSN=https://fake-dns.fake-sentry
+      - ODK_CENTRAL_FRONTEND_SENTRY_DSN=https://fake-dsn.fake-sentry
     volumes:
       - ../../files/nginx/odk.conf.template:/usr/share/odk/nginx/odk.conf.template:ro
       - ../../files/nginx/client-config.json.template:/usr/share/odk/nginx/client-config.json.template:ro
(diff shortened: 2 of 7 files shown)


.env.template
```diff
@@ -43,6 +43,10 @@ HTTPS_PORT=443
 # SENTRY_PROJECT=
 # SENTRY_TRACE_RATE=
 
+# Optional: configure frontend error reporting
+# Set to your Sentry DSN before `docker compose build` to embed it in the frontend bundle.
```
lognaturel: We can put the second line in docs instead of here! 

I haven’t looked deeply at how things are wired but wanted to flag that we’re working to switch to using a pre-built bundle in this release. So no problem if the var needs to be built into the image but it won’t be accessible for bundling.

lognaturel (commented):
Want to make sure you’ve see https://github.com/getodk/central/pull/1998 in case it intersects!

.env.template
latin-panda: As far as I understand, if https://github.com/getodk/central/pull/1998 lands, we won't be able to pass the DNS during build time, which is ideal, especially if we prefer not to track self-hosted instances. 
@alxndrsn do you have a suggestion?

latin-panda: Is #1998 planned for this release?

lognaturel: #1998 is currently planned for this release because it addresses the significant increase in build memory needed after the Web Forms merger.

I think what would make the most sense for now is to include the DSN in `/client-config.json` to match the OIDC config.

lognaturel: Reopened a [related Slack conversation](https://getodk.slack.com/archives/C051L6AKG4F/p1780613745083259?thread_ts=1780597625.941729&cid=C051L6AKG4F)!

latin-panda:
@alxndrsn Could you please review when you have time?

[PR to pass the DSN to the central-frontend](https://github.com/getodk/central-frontend/pull/1649)
[Issue for adding a proxy in the future ](https://github.com/getodk/central/issues/2006)

test/nginx/lib.docker-compose.yml
```diff
@@ -17,6 +17,7 @@ services:
       - SENTRY_ORG_SUBDOMAIN=o-fake-dsn
       - SENTRY_PROJECT=example-sentry-project
       - OIDC_ENABLED=false
+      - ODK_CENTRAL_FRONTEND_SENTRY_DSN=
```
alxndrsn: Is this line required?

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -452,7 +452,7 @@ function standardTestSuite({ fetchHttp, fetchHttp6, apiFetch, apiFetch6, forward
 
     // then
     assert.equal(res.status, 200);
-    assert.deepEqual(await res.json(), { oidcEnabled: false });
+    assert.deepEqual(await res.json(), { oidcEnabled: false, sentryDsn: '' });
```
alxndrsn: Is there any value in testing a realistic DSN here, or is the intention to test the default-to-empty-string handling?

alxndrsn (commented):
Comments in-line.

test/nginx/lib.docker-compose.yml
latin-panda: When ODK_CENTRAL_FRONTEND_SENTRY_DSN is empty, everything should still build correctly. However, I've added a placeholder string to test whether it's passed around like the other Sentry variables (L17, L18), which is more useful.

test/nginx/src/mocha/nginx.spec.js
latin-panda: I've added a placeholder string to test whether it's passed around like the other Sentry variables

latin-panda:
@alxndrsn this is ready for another look! Thanks!

test/nginx/lib.docker-compose.yml
```diff
@@ -17,7 +17,7 @@ services:
       - SENTRY_ORG_SUBDOMAIN=o-fake-dsn
       - SENTRY_PROJECT=example-sentry-project
       - OIDC_ENABLED=false
-      - ODK_CENTRAL_FRONTEND_SENTRY_DSN=
+      - ODK_CENTRAL_FRONTEND_SENTRY_DSN=https://fake-dns.fake-sentry
```
alxndrsn: `s/dns/dsn/`?

alxndrsn: (not sure why, but "suggestions" aren't working)

alxndrsn (changes requested):
Looks like one typo (comment in-line).

test/nginx/lib.docker-compose.yml
latin-panda: I'm sorry, I didn't quite understand the suggestion. Why use `s/dns/dns/`? 

Since Sentry DNS uses HTTPS, these tests currently accept any string and simply check if it's passed along. I was thinking of adding a string that starts with HTTPS to give a clearer idea of what kind of value is expected.

alxndrsn: I meant I think you've written "DNS" when it should be "DSN" :upside_down_face:

latin-panda: 🤦🏽‍♀️ I completely missed that typo! Thank you!

alxndrsn (approved):

PR #2033, opened 2026-07-02
nginx: remove ssl_trusted_certificate

### LINKED ISSUE #831: nginx: Remove SSL client certificate support config
Following up on [this discussion comment](https://github.com/getodk/central/pull/809#discussion_r1873624419), we may want to cut down on useless config noise by removing `ssl_trusted_certificate` directives from the nginx config.

alxndrsn:
This setting is only relevant if either `ssl_verify_client` or `ssl_stapling` is enabled.  Neither setting is.

See: https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_trusted_certificate

Any deployer who does require client certificate verification is likely sophisticated enough to re-enable themselves.

Closes #831

- [x] ci

This setting is only relevant if either `ssl_verify_client` or `ssl_stapling` is enabled.  Neither setting is.

No effect likely; self-hosters who use self-signed server certificates + client SSL certificates might be affected, but it seems like an unlikely combination.

No obvious related discussion in either of

* forums: https://forum.getodk.org/search?q=selfsign%20client
* issues: https://github.com/getodk/central/issues?q=is%3Aissue%20state%3Aopen%20selfsign%20client

No.


yanokwa:
@alxndrsn How does this change perform when testing with [ssllabs.com](https://www.ssllabs.com/ssltest/)

alxndrsn:
> @alxndrsn How does this change perform when testing with [ssllabs.com](https://www.ssllabs.com/ssltest/)

It has no visible effect on ssllabs.com scoring.

I would not have expected any effect anyway, as the setting only affects client certificate validation, which:

* I don't think ssllabs.com tests for, and
* I don't think is enabled anywhere in ODK Central's nginx config.

yanokwa (approved):

Issue #2029, opened 2026-06-30
[central-backend] ensure slow client reads are buffered by nginx


alxndrsn:
Currently it looks like slow client reads of large streamed data from postgres via `odk-central-backend` are held by NodeJS streams, which means they also hold PostgreSQL connections open.

Nginx should be able to buffer these responses more efficiently.

1. demonstrate the issue in a test in `test/nginx`
2. fix, likely by removing:

https://github.com/getodk/central/blob/d30c049dea9feabcb0e3f29ab0f872d120efbc54/files/nginx/odk.conf.template#L214

## Related

* https://github.com/getodk/central/issues/1901
* https://github.com/getodk/central/issues/2002

matthew-white:
Closed by #2036.

PR #2036, opened 2026-07-02
nginx: buffer responses in nginx, not application layer

alxndrsn:
Closes https://github.com/getodk/central/issues/2029

- [x] added a new test
- [x] deployed to a test server and demonstrated improved performance and cleaner failures for concurrent requests for large responses

- [ ] nginx disk RAM + space should be less expensive than postgres connection
- [ ] ~to avoid resource lock-up for responses over 100MB, they could by rejected by nginx~

- [ ] This should be a big win for performance, but what happens when nginx runs out of disk space?

No.


files/nginx/odk.conf.template
```diff
@@ -211,8 +211,14 @@ server {
...
     proxy_request_buffering on;
-    proxy_buffering off;
+    proxy_buffering on;
     proxy_read_timeout 2m;
+    proxy_temp_path /tmp;
+    proxy_max_temp_file_size 256m;
+    proxy_temp_file_write_size 256m;
+    proxy_buffers 64 64k;
+    proxy_buffer_size 64k;
+    proxy_busy_buffers_size 128k;
```
alxndrsn: test if all of these are necessary

test/nginx/mock-http-server/index.js
```diff
@@ -29,9 +33,58 @@ app.get('/health',      (req, res) => res.send('OK'));
...
   res.json('OK');
 });
 
+app.get('/v1/100MB.csv', (req, res) => {
+  const csvSizeBytes = 100_000_000;
+
+  // TODO confirm from IRL what headers should be set, e.g.
+  // content-type
+  // content-length
+  // transfer-encoding
```
alxndrsn: resolve this TODO

test/nginx/mock-http-server/index.js
```diff
@@ -29,9 +33,58 @@ app.get('/health',      (req, res) => res.send('OK'));
...
+
+  // TODO up this to 100MiB
+  const randomStream = Readable.from(generateCsv(csvSizeBytes));
+  randomStream.pipe(res);
+  req.on('close', () => {
+    randomStream.destroy();
+    --openProcessorCount;
+  });
+});
+app.get('/open-processor-count', (req, res) => {
```
alxndrsn: prefix these endpoints with `/direct-access/` or something?

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -436,6 +437,67 @@ function standardTestSuite({ fetchHttp, fetchHttp6, apiFetch, apiFetch6, forward
     });
   });
 
+  describe('response buffering', () => {
+    it('should buffer responses in nginx, not backend services', async function() {
+      // NOTE the final check should pass before the test timeout.  If it's taking longer,
+      // especially significantly longer, that implies that there is back-pressure on the
```
```suggestion
      // there may be back-pressure on the
```

test/nginx/src/mocha/request.js
```diff
@@ -16,29 +15,11 @@ function request(url, { body, ...options }={}) {
       const req = getProtocolImplFrom(url).request({ ...options, ...preserve(url) }, res => {
         res.on('error', reject);
 
-        const body = new Readable({ read:() => {} });
-        res.on('error', err => body.destroy(err));
-        res.on('data', data => body.push(data));
-        res.on('end', () => body.push(null));
-
-        const text = () => new Promise((resolve, reject) => {
```
alxndrsn: All this change should come from #2045

test/nginx/mock-http-server/index.js
alxndrsn: See: https://github.com/getodk/central/pull/2046

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -436,6 +437,67 @@ function standardTestSuite({ fetchHttp, fetchHttp6, apiFetch, apiFetch6, forward
     });
   });
 
+  describe('response buffering', () => {
+    it('should buffer responses in nginx, not backend services', async function() {
+      // NOTE the final check should pass before the test times out.  If timeout
+      // is exceeded, there may be back-pressure on the NodeJS server.  This
```
```suggestion
      // is reached, there may be back-pressure on the NodeJS server.  This
```

test/nginx/mock-http-server/index.js
alxndrsn: Expect:

### `/projects/1/forms/consumeDatasets/attachments/people.csv`

```http
content-disposition: attachment; filename="people.csv"; filename*=UTF-8''people.csv
content-type: text/csv; charset=utf-8
connection: close
transfer-encoding: chunked
...
```

### `/projects/:projectId/datasets/:name/entities.csv`

```http
content-disposition': attachment; filename="people.csv"; filename*=UTF-8''people.csv
content-type: text/csv; charset=utf-8
connection: close
transfer-encoding: chunked
...
```

### `/projects/:projectId/trash/datasets/:datasetId/entities.csv`

```http
content-disposition: attachment; filename="people-deleted.csv"; filename*=UTF-8''people-deleted.csv
content-type: text/csv; charset=utf-8
connection: close
transfer-encoding: chunked
...
```

### `/projects/:projectId/forms/:xmlFormId/submissions.csv`

```http
content-disposition: attachment; filename="simple.csv"; filename*=UTF-8''simple.csv
content-type: text/csv; charset=utf-8
connection: close
transfer-encoding: chunked
...
```

### `/projects/:projectId/forms/:xmlFormId/submissions.csv.zip`

```http
content-disposition: attachment; filename="simple.zip"; filename*=UTF-8''simple.zip
content-type: application/zip
connection: close
transfer-encoding: chunked
...
```

### `/projects/:projectId/forms/:xmlFormId/draft/submissions.csv`

```http
content-disposition: attachment; filename="simple.csv"; filename*=UTF-8''simple.csv
content-type: text/csv; charset=utf-8
connection: close
transfer-encoding: chunked
...
```

### `/projects/:projectId/forms/:xmlFormId/draft/submissions.csv.zip`

```http
content-disposition: attachment; filename="simple.zip"; filename*=UTF-8''simple.zip
content-type: application/zip
connection: close
transfer-encoding: chunked
...
```

files/nginx/odk.conf.template
```diff
@@ -211,7 +211,7 @@ server {
 
     # buffer requests, but not responses, so streaming out works.
```
lognaturel: This comment needs to be removed.

lognaturel:
> what happens when nginx runs out of disk space

My understanding is that if nginx runs out of disk space while trying to buffer a request to disk then that request aborts ([ref](https://trac.nginx.org/nginx/ticket/1950#comment:2)).

We could consider setting `proxy_max_temp_file_size` low, maybe something like in the 10mb range. We expect most requests to be small. Once that limit is exceeded for a single request, backpressure will go back to the application as it is now. That limit is per-request so with many concurrent clients it would still be possible to exhaust temp space.

Or we keep `proxy_max_temp_file_size` to its default of 1Gb and accept that on servers with many concurrent clients and low disk space some connections will be aborted.

I don't have a big preference between the two above. Maybe a slight preference for the second because it sticks to defaults, and we can adjust as we learn more about how that behaves.

Pverall, his change looks like a good one to me. I really don't understand why the default was overridden in the first place. In the case of a client/connection that can keep up with the application server, nginx will just stream the response out, right? Maybe there's a small difference in time to first byte since it looks like nginx does fill an in-memory buffer before sending: https://mailman.nginx.org/pipermail/nginx/2011-October/029972.html Maybe `proxy_buffering` was interpreted as the whole response being buffered instead of just the first 4KB.

I do think it would be nice to do a sanity check on e.g. `dev` before merging. The nice thing would be to simulate a client with a slow connection because I imagine that's where we'll see the biggest difference. I imagine there will be very little change for fast clients.

lognaturel:
> to avoid resource lock-up for responses over 100MB, they could by rejected by nginx

Can you expand on this? Because many of the responses are streamed, I don't think any part of the system would know that a response would be over 100MB ahead of time.

lognaturel:
@garethbowen when you're back could you please finish this one with @alxndrsn? No blockers from my end, just a couple of suggestions! Can't wait to see the impact it has.

alxndrsn:
> Can you expand on this? Because many of the responses are streamed, I don't think any part of the system would know that a response would be over 100MB ahead of time.

I think you're correct!  I've crossed out that suggestion in the PR description.

alxndrsn:
> We could consider setting `proxy_max_temp_file_size` low, maybe something like in the 10mb range. We expect most requests to be small.

I'm happy with the 1GB default - I think big responses are the main cause of problems when proxy-buffering is disabled.

files/nginx/odk.conf.template
```diff
@@ -209,9 +209,8 @@ server {
     proxy_pass http://service:8383;
     proxy_redirect off;
 
-    # buffer requests, but not responses, so streaming out works.
```
garethbowen, line 212: I tried to track down why this was added in the first place. I got down to [this commit](https://github.com/getodk/central/commit/5539a0c0d332421811ebca32640bef81289f7266) but couldn't find any issue or more info.

I suspect the key change was about turning request buffering on and leaving proxy buffering off, but it would be interesting to know if there was some reason that proxy buffering needed to be off.

Do you have any idea?

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -436,6 +437,67 @@ function standardTestSuite({ fetchHttp, fetchHttp6, apiFetch, apiFetch6, forward
     });
   });
 
+  describe('response buffering', () => {
+    it('should buffer responses in nginx, not backend services', async function() {
+      // NOTE the final check should pass before the test times out.  If timeout
+      // is reached, there may be back-pressure on the NodeJS server.  This
+      // would imply that nginx buffering is not working correctly.
+      // !!! ONLY CHANGE THIS TIMEOUT IF THE ABOVE COMMENT IS WELL UNDERSTOOD !!!
+      const testTimeout = 5_000;
```
garethbowen, line 443: The first time I read this comment I thought the test was just checking that it completed within 5 seconds, ie: the old way was slow, and the new way is fast. However upon closer inspection, if nginx buffering is broken, then because you only `reader.read()` once, this test will never complete regardless of what the timeout is set to, right?

If that's correct then this comment is not needed, and I'd argue, setting the test timeout is not needed - the default timeout will sort it out. It's not wrong, per se, but it threw me down the wrong track.

Furthermore, I think we can speed this test up by using a smaller file - it only needs to the smallest file above the buffer size to ensure the buffering is happening in the right service.

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -436,6 +437,67 @@ function standardTestSuite({ fetchHttp, fetchHttp6, apiFetch, apiFetch6, forward
...
+        const res = await apiFetch('/v1/projects/123/forms/some_form_id/attachments/100MB.csv', { signal });
+        const reader = res.body.getReader();
+        await reader.read();
+
+        // then
+        assert.equal(res.status, 200);
+        // and
+        assert.deepEqual(await getOpenProcessorCount(), { openProcessorCount:1, completedProcessorCount:0 });
+
+        await untilOpenProcessorCountIs({ timeout:testTimeout, openProcessorCount:0, completedProcessorCount:1 });
```
garethbowen, line 470: To test that the buffering is actually working, it would be good to `await reader.read()` again, and assert that the buffering is continuing even though the processor count is 0.

garethbowen (changes requested):
The nginx changes are great, more like best practice, and potentially great for performance.

I've got a couple of suggestions for the test that would give me more confidence that we don't regress later...

files/nginx/odk.conf.template
alxndrsn: From that commit:

> bug: set nginx proxy buffering settings to enable response streaming.

```
# buffer requests, but not responses, so streaming out works.
```

I think it's just a different trade-off was made - the original focus was propagating backpressure to NodeJS, vs. this PR which aims to relieve pressure on NodeJS (and consequently Postgres).

test/nginx/src/mocha/nginx.spec.js
alxndrsn: > If that's correct then this comment is not needed, and I'd argue, setting the test timeout is not needed - the default timeout will sort it out.

:+1: I'll look at tidying this up.

> Furthermore, I think we can speed this test up by using a smaller file - it only needs to the smallest file above the buffer size to ensure the buffering is happening in the right service.

This goes against my experience, but happy to be shown a faster alternative.

I think there are various buffers involved:

* NodeJS internal buffers
* TCP buffers in each container involved
* nginx buffers

On my machine, the 100MB file needs to be at least 11.7MB, and I'd rather leave a comfortable margin.  100MB is also realistic for production data that's currently causing issues.

alxndrsn: I think the test timeout is needed: with a file-size of 100MB, I'm seeing test times of ~2.2 seconds.  Default timeout is 2 seconds, so an increase is required.  I haven't tested on a slower laptop, but 5 seconds gives some headroom.

test/nginx/src/mocha/nginx.spec.js
alxndrsn: I've updated the tests to read the full 100MB buffer, and check that the expected amount of data has been read.

test/nginx/src/mocha/nginx.spec.js
alxndrsn: @garethbowen I think there's still value to the original comment, but it sounds like it's intention is unclear.

I'm aiming to communicate that a test timeout can indicate either:

1. nginx is propagating back-pressure to NodeJS, which is what we want to avoid, or
2. `testTimeout` is not big enough, as the test inherently relies on a race

Does that make sense?  If so, would a more helpful comment be something like:

```js
// If this test times out, it can indicate either:
// 1. nginx is propagating back-pressure to NodeJS, which is what we want to avoid, or
// 2. `testTimeout` is not big enough, as the test inherently relies on a race
```

?

garethbowen: > Default timeout is 2 seconds, so an increase is required.

Ahh I thought you were decreasing the timeout in an attempt to catch the bad behaviour. Increasing it makes sense, thanks.

> If so, would a more helpful comment be something like:

The comment is there to warn people against increasing the timeout right? Given the all caps and exclamation marks, I take this to mean you don't want people to _increase_ the timeout because that may hide the backpressure.

However from my read through the test is written better than that and if you can increase the timeout as much as you like but it'll never pass the test because it only ever reads the first page, right? This is the key point for me which I think you haven't answered yet - if the test is relying on the timeout being big enough but not too big then it's not deterministic (ie: faster hardware may get a false positive).

So, what happens if you revert the nginx fix and set the test timeout to very large - does the test pass despite the backpressure, or does it infinite loop waiting for the server to finish?

test/nginx/src/mocha/nginx.spec.js
garethbowen: Beautiful! Thanks

garethbowen (commented):
Comment inline. Mostly I just need to confirm my understanding of how the test works!

test/nginx/src/mocha/nginx.spec.js
alxndrsn: > However from my read through the test is written better than that and if you can increase the timeout as much as you like but it'll never pass the test because it only ever reads the first page, right?

:+1: I'll confirm this, and update the comment accordingly.

alxndrsn: Confirmed; I've removed the comment.

garethbowen (approved):
Beautiful!

alxndrsn:
# Testing 50 requests:

## Method

```sh
clear; \
echo "sleeping for 5 minutes to let server cool off a bit..." ;\
sleep 300; \
\
rm *.csv; \
\
for i in {1..50}; do (
  time curl \
      --limit-rate 250k \
      --silent \
      -H "Cookie: $COOKIE" \
      'https://dev.getodk.cloud/v1/projects/93/datasets/entities_100k/entities.csv' \
  >downloaded-$RANDOM.csv
) & sleep 2; done; \
wait; \
\
echo "all jobs done"; \
ls -halt *.csv; \
ls *.csv | xargs md5sum | cut -d' ' -f1; ls *.csv | xargs md5sum | cut -d' ' -f1 | sort | uniq -c;
```

## Results

### Possible file contents

```
45b50626f41c0e78d7e513b4540c4725 # {"message":"Internal Server Error"}
a3618ce2089f12053434f40dc5a9f33d # full CSV
ffa9475f398844d16f4a91e02f030583 # gateway timeout
```

### Without response buffering:

```
     12 45b50626f41c0e78d7e513b4540c4725
     21 a3618ce2089f12053434f40dc5a9f33d
     17 ffa9475f398844d16f4a91e02f030583
```


### With response buffering

```
      9 45b50626f41c0e78d7e513b4540c4725
     23 a3618ce2089f12053434f40dc5a9f33d
     18 ffa9475f398844d16f4a91e02f030583
```

## Conclusions

* This change does not appear to have a significant affect in this particular test.
* Enabling response buffering still feels like the correct call.

PR #2038, opened 2026-07-03
Update news for v2026.2

matthew-white:
This PR updates news.html for v2026.2.

Nothing, I just followed the existing pattern.

CI is failing, but it's also failing on the `master` branch.

### DIFF
--- a/docs/news.html
+++ b/docs/news.html
@@ -7,15 +7,15 @@
   </head>
   <body>
     <div class="news-item">
-      <span class="item-date">2026/05/04</span>
-      <a href="https://forum.getodk.org/t/57925/6" target="_blank">
-        ODK Central v2026.1.2
+      <span class="item-date">2026/07/03</span>
+      <a href="https://forum.getodk.org/t/58254" target="_blank">
+        ODK Central v2026.2
       </a>
     </div>
     <div class="news-item">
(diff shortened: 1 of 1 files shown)


matthew-white:
I'm intentionally targeting the `master` branch, since typically we merge the news PR after tagging the release.

docs/news.html
```diff
@@ -7,15 +7,15 @@
   </head>
   <body>
     <div class="news-item">
-      <span class="item-date">2026/05/04</span>
-      <a href="https://forum.getodk.org/t/57925/6" target="_blank">
-        ODK Central v2026.1.2
+      <span class="item-date">2026/07/03</span>
+      <a href="https://forum.getodk.org/t/58254" target="_blank">
+        ODK Central v2026.2
```
lognaturel: I think we usually add the last .0

lognaturel (approved):
Small suggestion!

docs/news.html
matthew-white: I was wondering about that and thought I remembered that we usually don't. I thought we only added the patch number when it's relevant / once we have a patch release. In any case, I think .0 is quite reasonable, so I'll go ahead and make that change. :+1:

matthew-white:
```suggestion
        ODK Central v2026.2.0
```

PR #2042, opened 2026-07-04
ci/build-push-image: fix nginx FRONTEND_BUILD_MODE error

### LINKED ISSUE #2043: [central] ci/build-push-image/nginx: fails due to missing build args
https://github.com/getodk/central/actions/runs/28684448271/job/85074813744

```
------
 > [linux/amd64 intermediate 4/5] RUN files/prebuild/write-version.sh:
0.073 files/prebuild/write-version.sh: line 26: FRONTEND_BUILD_MODE: unbound variable
------
nginx.dockerfile:15
--------------------
  13 |     COPY ./ ./
  14 |     
  15 | >>> RUN files/prebuild/write-version.sh
  16 |     RUN files/prebuild/build-frontend.sh
  17 |     
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c files/prebuild/write-version.sh" did not complete successfully: exit code: 1
```

alxndrsn:
Closes #2043

Ran it at: https://github.com/getodk/central/actions/runs/28703543011/job/85125477082

Still failing, but with a different error:

```
#39 [auth] getodk/central-nginx:pull,push token for ghcr.io
#39 DONE 0.0s
#38 exporting to image
#38 pushing layers 0.7s done
#38 ERROR: failed to push ghcr.io/getodk/central-nginx:pr-2042: denied: installation not allowed to Write organization package
------
 > exporting to image:
------
ERROR: failed to build: failed to solve: failed to push ghcr.io/getodk/central-nginx:pr-2042: denied: installation not allowed to Write organization package
```

I think this error is correct - this build was not run on a release git tag, so it's probably correct that it doesn't have permission to upload docker images for a release docker tag.

Likely a better long-term solution is to change actions from `docker/build-push-action` to `docker/bake-action`.  However, this close to release it's prudent to keep changes minimal by not changing action.  A ticket has been filed for changing action in the future at #2044. 

No impact.

No.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,137 +1,18 @@
-name: Test, Build, Publish
+name: Test build-push-image
 
 on:
   push:
   pull_request:
-  workflow_dispatch:
-    inputs:
-      publish_image:
-        description: 'Publish image to registry?'
-        required: true
-        type: boolean
-        default: false
 
 jobs:
-  test-misc: # quick, simple checks
-    timeout-minutes: 2
-    runs-on: ubuntu-latest
-    steps:
-    - run: docker --version
-    - run: docker compose version
-    - uses: actions/checkout@v5
-    - run: ./test/check-submodules.sh
-    - run: ./test/test-with-pgenvblock.sh
-    - run: sudo apt-get install shellcheck
-    - run: ./test/check-scripts.sh
-    - run: ./test/check-for-large-files.sh
-    - run: ./test/check-dockerfiles.sh
-    - run: cd test && npm clean-install
-    - run: cd test && npm run lint
-    - run: cd test && npm run test:github-actions
-  test-envsub:
-    timeout-minutes: 2
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v5
-    - run: cd test/envsub && ./run-tests.sh
-  test-nginx:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -162,6 +162,13 @@ jobs:
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v4
 
+    - name: Extract FRONTEND_VERSION
+      run: |
+           touch .env
+           echo "FRONTEND_VERSION=$(
+             docker compose config --format json | jq -r .services.nginx.build.args.FRONTEND_VERSION
```
lognaturel, line 169: How about making this `jq -er` so the whole step fails if somehow the docker compose file or specific constant are missing?

lognaturel (approved):
Thank you! I think this can be merged now. But it won't give us tagged v2026.2.0 images, right? Can we somehow force this action version to run on the tagged commit? Or temporarily set the frontend version in Github variables and rerun the old action?

.github/workflows/main.yml
alxndrsn: very helpful :+1:

alxndrsn: ...but doesn't work out-of-the-box in this case.

I think the assignment would have to be extracted from the `echo` call.

PR #2054, opened 2026-07-07
service/start-odk: expose postgres connection errors & allow failure

alxndrsn:
* replace `psql` with `pg_isready`
* don't suppress useful output

##### ☑️ if `pg_isready` is missing, it fails with:

```
service-1   | waiting for PostgreSQL to become connectable to...
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | sleeping 1 second waiting for a database connection
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | sleeping 1 second waiting for a database connection
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | sleeping 1 second waiting for a database connection
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | sleeping 1 second waiting for a database connection
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | sleeping 1 second waiting for a database connection
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | sleeping 1 second waiting for a database connection
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | sleeping 1 second waiting for a database connection
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | sleeping 1 second waiting for a database connection
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | sleeping 1 second waiting for a database connection
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | sleeping 1 second waiting for a database connection
service-1   | ./start-odk.sh: line 48: pg_isreadyxxx: command not found
service-1   | Postgres not available after 10 attempts.
service-1 exited with code 1 (restarting)
```

##### ☑️ If host is incorrect if fails with:

```
service-1  | waiting for PostgreSQL to become connectable to...
service-1  | postgres.bad.example.test:5432:5432 - no response
service-1  | sleeping 1 second waiting for a database connection
service-1  | postgres.bad.example.test:5432:5432 - no response
service-1  | sleeping 1 second waiting for a database connection
service-1  | postgres.bad.example.test:5432:5432 - no response
service-1  | sleeping 1 second waiting for a database connection
service-1  | postgres.bad.example.test:5432:5432 - no response
service-1  | sleeping 1 second waiting for a database connection
service-1  | postgres.bad.example.test:5432:5432 - no response
service-1  | sleeping 1 second waiting for a database connection
service-1  | postgres.bad.example.test:5432:5432 - no response
service-1  | sleeping 1 second waiting for a database connection
service-1  | postgres.bad.example.test:5432:5432 - no response
service-1  | sleeping 1 second waiting for a database connection
service-1  | postgres.bad.example.test:5432:5432 - no response
service-1  | sleeping 1 second waiting for a database connection
service-1  | postgres.bad.example.test:5432:5432 - no response
service-1  | sleeping 1 second waiting for a database connection
service-1  | postgres.bad.example.test:5432:5432 - no response
service-1  | sleeping 1 second waiting for a database connection
service-1  | Postgres not available after 10 attempts.
service-1 exited with code 1 (restarting)
```

##### ☑️ If database name is incorrect if fails with:

```
service-1  | waiting for PostgreSQL to become connectable to...
service-1  | postgres14:5432 - accepting connections
service-1  | running migrations..
service-1  | Error: database "bad_db_name" does not exist
service-1 exited with code 1 (restarting)
```

##### ☑️ If username is incorrect if fails with:

```
service-1  | waiting for PostgreSQL to become connectable to...
service-1  | postgres14:5432 - accepting connections
service-1  | running migrations..
service-1  | Error: password authentication failed for user "not_a_real_user"
service-1 exited with code 1 (restarting)
```

##### ☑️ If password is incorrect if fails with:

```
service-1  | waiting for PostgreSQL to become connectable to...
service-1  | postgres14:5432 - accepting connections
service-1  | running migrations..
service-1  | Error: password authentication failed for user "odk"
```

* specifically written for this job
* provided by postgres team
* idiomatic

See: https://www.postgresql.org/docs/current/app-pg-isready.html

Low chance of regression, but perhaps the current approach is doing something non-obvious.

No.

### DIFF
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -43,7 +43,10 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
 export SENTRY_TAGS
 
 echo "waiting for PostgreSQL to become connectable to..."
-while ! (psql --no-password --quiet --command "" > /dev/null 2>&1 || (echo "sleeping 1 second waiting for a database connection"; false)); do sleep 1; done
+while ! pg_isready --quiet; do
+  echo "sleeping 1 second waiting for a database connection"
+  sleep 1
+done
 
 echo "running migrations.."
 node ./lib/bin/run-migrations
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -43,10 +43,18 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
 export SENTRY_TAGS
 
 echo "waiting for PostgreSQL to become connectable to..."
-while ! pg_isready --quiet; do
+maxRetries=10
+retries=$maxRetries
+while ! pg_isready --timeout 10 && [[ $retries -gt 0 ]]; do
   echo "sleeping 1 second waiting for a database connection"
+  retries=$((retries-1))
   sleep 1
 done
+if [[ "$retries" = 0 ]]; then
+  echo "Postgres not available after $maxRetries attempts."
+  sleep 1 # reduce thrashing
+  exit 1
+fi
 
 echo "running migrations.."
 node ./lib/bin/run-migrations
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -43,16 +43,16 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
 export SENTRY_TAGS
 
 echo "waiting for PostgreSQL to become connectable to..."
-maxRetries=15
-retries=$maxRetries
+maxTries=15
+retries=$((maxTries-1))
 while ! pg_isready; do
-  echo "sleeping 1 second waiting for a database connection"
-  retries=$((retries-1))
-  sleep 1
-  if [[ "$retries" -lt 0 ]]; then
-    echo "PostgreSQL not available after $maxRetries attempts."
+  if [[ "$retries" = 0 ]]; then
+    echo "PostgreSQL not available after $maxTries attempts."
     exit 1
   fi
+  echo "PostgreSQL not yet available; sleeping 1 second..."
+  sleep 1
+  retries=$((retries-1))
 done
 
 echo "running migrations.."
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -43,13 +43,13 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
 export SENTRY_TAGS
 
 echo "waiting for PostgreSQL to become connectable to..."
-maxRetries=10
+maxRetries=15
 retries=$maxRetries
-while ! pg_isready --timeout 10; do
+while ! pg_isready; do
   echo "sleeping 1 second waiting for a database connection"
   retries=$((retries-1))
   sleep 1
-  if [[ "$retries" = 0 ]]; then
+  if [[ "$retries" -lt 0 ]]; then
     echo "PostgreSQL not available after $maxRetries attempts."
     exit 1
   fi
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -50,7 +50,7 @@ while ! pg_isready --timeout 10; do
   retries=$((retries-1))
   sleep 1
   if [[ "$retries" = 0 ]]; then
-    echo "Postgres not available after $maxRetries attempts."
+    echo "PostgreSQL not available after $maxRetries attempts."
     exit 1
   fi
 done
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -45,16 +45,15 @@ export SENTRY_TAGS
 echo "waiting for PostgreSQL to become connectable to..."
 maxRetries=10
 retries=$maxRetries
-while ! pg_isready --timeout 10 && [[ $retries -gt 0 ]]; do
+while ! pg_isready --timeout 10; do
   echo "sleeping 1 second waiting for a database connection"
   retries=$((retries-1))
   sleep 1
+  if [[ "$retries" = 0 ]]; then
+    echo "Postgres not available after $maxRetries attempts."
+    exit 1
+  fi
 done
-if [[ "$retries" = 0 ]]; then
-  echo "Postgres not available after $maxRetries attempts."
-  sleep 1 # reduce thrashing
-  exit 1
-fi
 
 echo "running migrations.."
 node ./lib/bin/run-migrations


matthew-white:
> Low chance of regression, but perhaps the current approach is doing something non-obvious.

Wicher and I did discuss `psql` vs. `pg_isready` at https://github.com/getodk/central/pull/1671#discussion_r2897326268. My understanding is that `pg_isready` checks that the server can be connected to, validating the host and port. Whereas `psql` will also validate the database name. I guess `pg_isready` would pass successfully if someone mistyped the database name?

matthew-white:
That understanding is just based on Wicher's comment, I haven't actually confirmed it myself.

alxndrsn:
@matthew-white thanks for the explanation.  In light of that discussion, I think either:

### A. the messaging needs updating:

"waiting for PostgreSQL to become connectable to..." is accurate for `pg_isready`, but inaccurate for `psql`, or

### B. switching to `pg_isready`

As touched on in the discussion, this would be more similar to the previous behaviour with `wait-for-it`, but it changed incidentally in:

* https://github.com/getodk/central/pull/1647

It would be helpful to get a clear understanding of what failure actually looks like if postgres is available, but the db name or other details are incorrect.

matthew-white:
Not sure if this is a good idea, but what about an option C of using both? Using `pg_isready` + `sleep` in a loop to do the waiting. Then once the waiting is done, running `psql` just once to confirm that the database name is valid and to avoid the sort of boot loop that Wicher mentioned in his comment.

alxndrsn:
> My understanding is that `pg_isready` checks that the server can be connected to, validating the host and port. Whereas `psql` will also validate the database name. I guess `pg_isready` would pass successfully if someone mistyped the database name?

The current `psql` setup swallows the error if the db name is incorrect:

```
service-1  | waiting for PostgreSQL to become connectable to...
service-1  | sleeping 1 second waiting for a database connection
service-1  | sleeping 1 second waiting for a database connection
service-1  | sleeping 1 second waiting for a database connection
```

Switching to `pg_isready` does "boot loop":

```
service-1  | waiting for PostgreSQL to become connectable to...
service-1  | running migrations..
service-1  | Error: database "central_bad" does not exist
service-1 exited with code 1
service-1  | waiting for PostgreSQL to become connectable to...
service-1  | running migrations..
service-1  | Error: database "central_bad" does not exist
service-1 exited with code 1
```

> Not sure if this is a good idea, but what about an option C of using both? Using `pg_isready` + `sleep` in a loop to do the waiting. Then once the waiting is done, running `psql` just once to confirm that the database name is valid and to avoid the sort of boot loop that Wicher mentioned in his comment.

A hybrid approach could look something like:

```sh
log "waiting for PostgreSQL to become connectable to..."
while ! pg_isready --quiet; do
  log "sleeping 1 second waiting for a database connection"
  sleep 1
done
log "testing database connection..."
if ! psql --no-password --command ""; then
  sleep 1 # reduce thrashing
  exit 1
fi
```

output:

```
service-1  | [start-odk] waiting for PostgreSQL to become connectable to...
service-1  | [start-odk] testing database connection...
service-1  | psql: error: connection to server at "example.com" (123.123.123.123), port 5432 failed: FATAL:  database "central_bad" does not exist
service-1 exited with code 1
service-1  | [start-odk] waiting for PostgreSQL to become connectable to...
service-1  | [start-odk] testing database connection...
service-1  | psql: error: connection to server at "example.com" (123.123.123.123), port 5432 failed: FATAL:  database "central_bad" does not exist
service-1 exited with code 1
```

This will still restart the container after failure.

Per the comment at https://github.com/getodk/central/pull/1671#discussion_r2898154274:

> not bootlooping makes other things (such as docker exec) easier.

If avoiding restarting the container is an explicit goal, then I think this PR should be closed and replaced with one which:

1. corrects log messages to reflect what's actually happening, and
2. stops swallowing the cause of failures

matthew-white:
> Switching to `pg_isready` does "boot loop"

> A hybrid approach could look something like: &hellip; This will still restart the container after failure.

> If avoiding restarting the container is an explicit goal, then I think this PR should be closed and replaced with one which:
> 
> 1. corrects log messages to reflect what's actually happening, and
> 2. stops swallowing the cause of failures

I feel like I'm getting to the limit of my knowledge of Docker best practice. Wicher seems to have wanted to avoid restarting the container. Is that a good idea / a beneficial approach? If so, I like the idea of exploring an alternative PR that corrects log messages and stops swallowing errors.

alxndrsn:
> wanted to avoid restarting the container. Is that a good idea / a beneficial approach?

Probably not.  If the database is available, then presumably the hostname is correct.  `pg_isready` should pass.

If the connection fails due to other misconfiguration (db name, username, password) then this can be resolved by:

1. changing values in `.env` or `docker-compose.yml`, or
2. updating/creating relevant db or user on the postgres side

Any changes made via **1** will only be available once the container has restarted.

Additionally, any healthchecks looking at container liveness or lifetime won't see any problems with a container that never restarts on connection failure.

matthew-white:
Sounds good, I'm convinced that we don't need to avoid restarting the container. :+1:

I kind of like the look of your hybrid approach above. I like that it fails fast, before reaching "running migrations.." Do you have a preference between the current approach of this PR vs. a hybrid approach?

alxndrsn:
> Do you have a preference between the current approach of this PR vs. a hybrid approach?

I prefer this PR as it's less code.

Comparing the failure messages, I don't think there's much in it:

1. with `psql`:

```
service-1  | waiting for PostgreSQL to become connectable to...
service-1  | testing database connection...
service-1  | psql: error: connection to server at "example.com" (123.123.123.123), port 5432 failed: FATAL:  database "central_bad" does not exist
service-1 exited with code 1
```

2. without `psql`:

```
service-1  | waiting for PostgreSQL to become connectable to...
service-1  | running migrations..
service-1  | Error: database "central_bad" does not exist
service-1 exited with code 1
```

If the `psql` approach might be preferred, maybe we can debate + add in a separate PR?

files/service/scripts/start-odk.sh
```diff
@@ -43,7 +43,10 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
 export SENTRY_TAGS
 
 echo "waiting for PostgreSQL to become connectable to..."
-while ! (psql --no-password --quiet --command "" > /dev/null 2>&1 || (echo "sleeping 1 second waiting for a database connection"; false)); do sleep 1; done
+while ! pg_isready --quiet; do
```
alxndrsn: If the final decided approach is to allow container failure, a max-retry counter or timeout needs to be added here.

alxndrsn:
@matthew-white if the PR direction is accepted, I'll update the title to:

> service/start-odk: expose postgres connection errors & allow failure

matthew-white:
> I prefer this PR as it's less code.

Let's go with that then. :+1: I don't have a strong preference for the hybrid approach. I agree that the error messaging is pretty similar either way.

files/service/scripts/start-odk.sh
alxndrsn: Done.

files/service/scripts/start-odk.sh
```diff
@@ -43,7 +43,17 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
 export SENTRY_TAGS
 
 echo "waiting for PostgreSQL to become connectable to..."
-while ! (psql --no-password --quiet --command "" > /dev/null 2>&1 || (echo "sleeping 1 second waiting for a database connection"; false)); do sleep 1; done
+maxRetries=10
+retries=$maxRetries
+while ! pg_isready --timeout 10; do
+  echo "sleeping 1 second waiting for a database connection"
```
alxndrsn:
```suggestion
  echo "PostgreSQL not yet available; sleeping 1 second..."
```

files/service/scripts/start-odk.sh
```diff
@@ -43,7 +43,17 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
...
+  retries=$((retries-1))
+  sleep 1
+  if [[ "$retries" = 0 ]]; then
+    echo "Postgres not available after $maxRetries attempts."
```
alxndrsn:
```suggestion
    echo "PostgreSQL not available after $maxRetries attempts."
```

matthew-white (approved):

files/service/scripts/start-odk.sh
```diff
@@ -43,7 +43,17 @@ SENTRY_TAGS="{ \"version.central\": \"$(cat sentry-versions/central)\", \"versio
 export SENTRY_TAGS
 
 echo "waiting for PostgreSQL to become connectable to..."
-while ! (psql --no-password --quiet --command "" > /dev/null 2>&1 || (echo "sleeping 1 second waiting for a database connection"; false)); do sleep 1; done
+maxRetries=10
```
matthew-white: I feel like this number may be on the low side. How about making it 25 or 50 instead?

matthew-white: Though I guess you're specifying `--timeout 10` to `pg_isready`, and 10 x 10 = 100. So actually, probably better as-is!

alxndrsn: I think this is a reasonable concern.

The timeout for `pg_isready` is a bit more specific than it appears.

Here's what happens for a host without postgres:

```sh
$ PGHOST=example.com time -f %E pg_isready --timeout 10
example.com:5432 - no response
Command exited with non-zero status 2
0:20.11
```

vs a host that doesn't resolve:

```
$ PGHOST=baddomain.example.com time -f %E pg_isready --timeout 10
baddomain.example.com:5432 - no response
Command exited with non-zero status 2
0:00.03
```

The historic behaviour for this container was to call `wait-for-it` with default timeout: https://github.com/getodk/central/blob/a4908f504fca18b755887b16ffc365d983252192/docker-compose.yml#L82

The default `wait-for-it` timeout is 15 seconds:

Defined in source code: https://github.com/vishnubob/wait-for-it/blob/master/wait-for-it.sh#L139

Checked with use:

```
$ wait-for-it example.com:80 -- bash -c "echo finished"
wait-for-it: waiting 15 seconds for example.com:80
wait-for-it: example.com:80 is available after 0 seconds
finished
```

How about removing the `--timeout` arg from the `pg_isready` invocation, and implementing a 15+ second timeout loop around `pg_isready`?

matthew-white: > The timeout for `pg_isready` is a bit more specific than it appears.

> Here's what happens for a host without postgres: &hellip; `0:20.11`

I don't think I really understand what's happening here (but I'm also not sure I need to). Is the part that's surprising the fact that it's 20 seconds instead of 10? My initial concern was that we weren't retrying enough, but is this saying that `pg_isready` is waiting too long?

> How about removing the `--timeout` arg from the `pg_isready` invocation, and implementing a 15+ second timeout loop around `pg_isready`?

That sounds reasonable to me. It looks like `--timeout` defaults to 3 seconds, so you may need to specify `--timeout 0` to disable.

The main case I'm thinking about is the local/built-in Postgres. It may take it a little bit to spin up, so we want to make sure to give it enough time.

matthew-white: > vs a host that doesn't resolve: &hellip; `0:00.03`

Oh I think I see the problem you're describing now: it's the fact that this is super short. If `pg_isready` finishes almost instantly, then all we have is `sleep 1` x 10, which is probably not long enough.

alxndrsn: > It looks like `--timeout` defaults to 3 seconds, so you may need to specify `--timeout 0` to disable.

Three seconds seems reasonable.  If set to zero, a resolving host will cause the command to hang forever:

```sh
$ PGHOST=example.com pg_isready --timeout 0
```

alxndrsn: I've updated to try for a _minimum_ of 15 seconds.

files/service/scripts/start-odk.sh
alxndrsn: Merged manually.

PR #2064, opened 2026-07-10
service/start-odk: fix DB_SSL check in non docker-compose envs

### LINKED ISSUE #2063: [central] DB_SSL must be manually set to `null` in non-docker-compose environments
Currently the check for `DB_SSL` assumes that the `service` container is run by `docker compose`, which will substitute `null` for an unset var.

However, in non-compose environment, `start-odk.sh` fails to start if `DB_SSL` is not defined.  The failure message is very misleading.

See: https://forum.getodk.org/t/service-2026-2-0-will-not-start/58361

alxndrsn:
Closes #2063

New test.

It doesn't change behaviour when using docker compose, but should fix the problem for environments where it's really possible to not have any value for `DB_SSL`.

Fixes a bug.

No.

### DIFF
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -3,7 +3,7 @@ set -o pipefail
 shopt -s inherit_errexit
 
 # Check for illegal DB_SSL environment variable.
-if ! [[ "${DB_SSL-}" = null ]]; then
+if [[ -v DB_SSL ]] && ! [[ "$DB_SSL" = null ]]; then
   echo "!!!"
   echo "!!! You have the DB_SSL variable defined (in your .env file, probably)."
   echo "!!! This variable is no longer supported from Central 2026.1 onwards."
--- a/files/service/scripts/start-odk.sh
+++ b/files/service/scripts/start-odk.sh
@@ -3,7 +3,7 @@ set -o pipefail
 shopt -s inherit_errexit
 
 # Check for illegal DB_SSL environment variable.
-if [[ -v DB_SSL ]] && ! [[ "$DB_SSL" = null ]]; then
+if [[ -v DB_SSL ]] && ! [[ "$DB_SSL" = "null" ]]; then
   echo "!!!"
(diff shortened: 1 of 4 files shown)


test/service/service.spec.js
```diff
@@ -29,7 +29,7 @@ describe('service image', () => {
     const runService = (...args) => new Promise(resolve => {
       const stdcombi = [];
 
-      const process = spawn('docker', [ 'compose', 'run', ...args, 'service' ], { cwd:'..' });
+      const process = spawn('docker', [ 'compose', ...args, 'service' ], { cwd:'..' });
```
alxndrsn, line 32: The new test required a change to where `args` can be inserted.

test/package.json
```diff
@@ -6,7 +6,7 @@
     "test:nginx": "npm run test:nginx:mocha && npm run test:nginx:playwright",
     "test:nginx:mocha": "NODE_TLS_REJECT_UNAUTHORIZED=0 mocha './nginx/src/mocha/**/*.spec.js' --require ./nginx/src/mocha/mocha.setup.js",
     "test:nginx:playwright": "NODE_TLS_REJECT_UNAUTHORIZED=0 playwright test",
-    "test:service": "mocha ./service.spec.js",
+    "test:service": "mocha ./service/service.spec.js",
```
alxndrsn, line 9: This change is just to organise the `service` tests better, now that these tests span more than one file.

lognaturel (approved):

files/service/scripts/start-odk.sh
```diff
@@ -3,7 +3,7 @@ set -o pipefail
 shopt -s inherit_errexit
 
 # Check for illegal DB_SSL environment variable.
-if ! [[ "${DB_SSL-}" = null ]]; then
+if [[ -v DB_SSL ]] && ! [[ "$DB_SSL" = null ]]; then
```
alxndrsn:
```suggestion
if [[ -v DB_SSL ]] && ! [[ "$DB_SSL" = "null" ]]; then
```

Should make this more obvious for readers coming from JS.

Issue #2006, opened 2026-06-19
Consider adding a proxy for frontend Sentry events


latin-panda:
The frontend Sentry SDK supports a tunnel option that routes events through a first-party endpoint instead of directly to [sentry.io](http://sentry.io/), preventing ad-blockers from silently dropping error reports.

Would require:
- A new nginx location (e.g. /sentry-tunnel) proxying to the frontend Sentry project's ingest /envelope/ endpoint
- New env vars for the frontend Sentry project (separate from the existing backend SENTRY_ORG_SUBDOMAIN / SENTRY_KEY / SENTRY_PROJECT)
- tunnel: '/sentry-tunnel' added to initSentry options

PR #2066, opened 2026-07-11
nginx/csp/frontend: ensure POSTs to Sentry are not blocked

### LINKED ISSUE #2056: [central-frontend] CSP blocking sentry reports
It looks like Sentry reports for web-forms &/or other frontend apps are blocked by the CSP: https://getodk.sentry.io/issues/7600919924/

### TODO

- [ ] fully confirm
  - [ ] which frontend app(s) are affected
  - [x] if these connections are legit

### Suggested fixes

Either:

1. allow `connect-src` for .sentry.io, or the full DSN, or
2. proxy sentry requests through `nginx` (https://github.com/getodk/central/issues/2006)

alxndrsn:
Closes #2056

- [x] new tests
- [x] had a good think about it

Could make the CSP more lax (e.g. the whole of sentry.org).  This would be a smaller header, but would allow unexpected things from those domains.

No effect expected.

No.


.github/workflows/main.yml
```diff
@@ -16,24 +16,9 @@ jobs:
...
-    - run: ./test/check-dockerfiles.sh
     - run: cd test && npm clean-install
     - run: cd test && npm run lint
-    - run: cd test && npm run test:github-actions
-  test-envsub:
-    timeout-minutes: 2
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v5
-    - run: cd test/envsub && ./run-tests.sh
```
alxndrsn, line 36: revert

.github/workflows/main.yml
```diff
@@ -45,10 +30,23 @@ jobs:
     - run: cd test/nginx && npm clean-install
     - run: cd test/nginx && ./setup-tests.sh
     - run: cd test/nginx && npm run test:nginx:mocha
-    - run: cd test/nginx && ./lint-config.sh
+    #- run: cd test/nginx && ./lint-config.sh
```
alxndrsn, line 49: revert

.github/workflows/main.yml
```diff
@@ -45,10 +30,23 @@ jobs:
...
 
     - run: cd test/nginx && npx playwright install --with-deps chromium-headless-shell
-    - run: cd test/nginx && npm run test:nginx:playwright
+    - id: playwright-tests
+      run: cd test/nginx && npm run test:nginx:playwright
+    - run: ls -al ./test/.playwright-report
+      if: always()
```
alxndrsn: remove

.github/workflows/main.yml
```diff
@@ -45,10 +30,23 @@ jobs:
...
+    - name: Archive playwright result                                                                                   
+      if: always() && steps.playwright-tests.outcome == 'failure'
+      uses: actions/upload-artifact@v7
+      with:
+        name: Playwright Artifacts
+        path: test/.playwright-report/
+        include-hidden-files: true
```
alxndrsn: separate PR

.github/workflows/main.yml
```diff
@@ -45,10 +30,23 @@ jobs:
...
+
+    - if: always()
+      run: find . -type d
```
alxndrsn: remove

.github/workflows/main.yml
alxndrsn: or just remove - console logging is good already

.github/workflows/main.yml
```diff
@@ -60,113 +58,3 @@ jobs:
...
-      with:
-        file: ${{ matrix.image }}.dockerfile
-        context: .
-        push: true
-        tags: ${{ steps.meta.outputs.tags }}
-        labels: ${{ steps.meta.outputs.labels }}
-        platforms: 'linux/amd64,linux/arm64'
-        build-args: |
-          FRONTEND_BUILD_MODE=fetch
-          FRONTEND_VERSION=${{ env.FRONTEND_VERSION }}
```
alxndrsn, line 172: revert

test/nginx/mock-sentry/index.js
```diff
@@ -49,6 +49,15 @@ app.use('/api', (req, res, next) => {
 
   next();
 });
+app.use('/api/:projectId/envelope/', (req, res, next) => {
+  res.setHeader('Access-Control-Allow-Origin', 'https://odk-nginx.example.test:9001');
+  res.setHeader('Access-Control-Allow-Methods', 'POST, OPTIONS');
```
alxndrsn:
```suggestion
  res.setHeader('Access-Control-Allow-Methods', 'OPTIONS, POST');
```

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -442,7 +443,7 @@ function standardTestSuite({ fetchHttp, fetchHttp6, apiFetch, apiFetch6, forward
 
     // then
     assert.equal(res.status, 200);
-    assert.deepEqual(await res.json(), { oidcEnabled: false, sentryDsn: 'https://fake-dsn.fake-sentry' });
+    assert.deepEqual(await res.json(), { oidcEnabled:false, sentryDsn:'https://abcdef0123456789abcdef0123456789@o-fake-dsn.ingest.sentry.io/1234567890123456' });
```
alxndrsn, line 446: Separate PR.

test/nginx/src/mocha/nginx.spec.js
```diff
@@ -452,7 +453,7 @@ function standardTestSuite({ fetchHttp, fetchHttp6, apiFetch, apiFetch6, forward
...
```
alxndrsn, line 456: Separate PR.

test/nginx/lib.docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
       - SENTRY_ORG_SUBDOMAIN=o-fake-dsn
       - SENTRY_PROJECT=example-sentry-project
       - OIDC_ENABLED=false
-      - SENTRY_DSN_FRONTEND=https://fake-dsn.fake-sentry
+      - SENTRY_DSN_FRONTEND=https://abcdef0123456789abcdef0123456789@o-fake-dsn.ingest.sentry.io/1234567890123456
```
alxndrsn: Separate PR.

test/nginx/mock-sentry/index.js
```diff
@@ -32,6 +32,7 @@ app.get('/__mock_sentry/reset',       (req, res) => {
   events.length = 0;
   res.json('OK');
 });
+
```
alxndrsn, line 35: This newline separates API routes from others.

files/nginx/setup-odk.sh
```diff
@@ -45,6 +45,7 @@ echo "writing fresh nginx templates..."
   > /etc/nginx/conf.d/redirector.conf
 
 CERT_DOMAIN=$( [ "$SSL_TYPE" = "customssl" ] && echo "local" || echo "$DOMAIN") \
+SENTRY_DSN_FRONTEND_ROOT="$(sed -E -e 's:/[a-f0-9]*@:/:' -e 's:/[0-9]*$::' <<<"$SENTRY_DSN_FRONTEND")" \
```
matthew-white:
```suggestion
SENTRY_DSN_FRONTEND_ROOT="$(sed -E -e 's:/[a-f0-9]+@:/:' -e 's:/[0-9]+$::' <<<"$SENTRY_DSN_FRONTEND")" \
```

Just thinking that `*`→`+` would make the regex slightly more robust.

matthew-white: Initially I was wondering what would happen if Sentry were disabled on the server side. But it looks like `SENTRY_DSN_FRONTEND_ROOT` would end up being blank. It's still inserted/replaced in odk.conf.template, but since it's blank, it's not really part of the CSP.

matthew-white: If someone specified a `SENTRY_DSN_FRONTEND` that didn't match this pattern, then it would be inserted in odk.conf.template as-is without any replacement. Does that sound potentially problematic, or not a big deal? I'm wondering whether we should check that `SENTRY_DSN_FRONTEND` matches this pattern and error out if not. (Could probably be convinced either way, mostly just wondering at this point.)

test/nginx/src/playwright/csp.spec.js
```diff
@@ -13,6 +15,52 @@ test.beforeEach(async ({ page }) => {
   });
 });
 
+test.describe('API calls', () => {
+  [
+    '/',
+  ].forEach(path => {
```
matthew-white, line 25: Do you expect to add more paths? Just wondering whether this test structure could be simplified.

matthew-white (commented):
The approach here seems very reasonable to me. I was also a fan of #2006, but this seems straightforward as well.

Just leaving a few specific questions

test/nginx/src/playwright/csp.spec.js
```diff
@@ -13,6 +15,52 @@ test.beforeEach(async ({ page }) => {
   });
 });
 
+test.describe('API calls', () => {
```
alxndrsn: Which API?

test/nginx/src/playwright/csp.spec.js
alxndrsn: Yes, more paths added.

files/nginx/setup-odk.sh
alxndrsn: Or less robust?

alxndrsn: I've added a test for this situation.

alxndrsn: I think just stripping it from the CSP would be reasonable; I'll look at adding a test case for this.

alxndrsn: I've updated to strip the sentry host from the CSP if the format looks incorrect.

files/nginx/setup-odk.sh
```diff
@@ -15,12 +15,15 @@ fi
...
-    -days 365000
+BADHOST_DH_PATH=/etc/nginx/ssl/nginx.default
+if ! [ -s "$BADHOST_DH_PATH.key" ] || ! [ -s "$BADHOST_DH_PATH.crt" ]; then
+  mkdir -p /etc/nginx/ssl
+  openssl req -x509 -nodes -newkey rsa:2048 \
+      -subj "/CN=invalid.local" \
+      -keyout "$BADHOST_DH_PATH.key" \
+      -out    "$BADHOST_DH_PATH.crt" \
+      -days 365000
+fi
```
alxndrsn: Revert this change:

1. it's to make startup faster, so not directly relevant to this PR, and
2. this command doesn't take long anyway

alxndrsn: Change separated at https://github.com/getodk/central/pull/2096

alxndrsn: reverted.

files/nginx/setup-odk.sh
matthew-white: Don't we always expect there to be at least one of these characters? To me, `+` seems like a better fit for that. If there are zero characters and it doesn't match the expected pattern, then maybe we just should strip it from the CSP along the lines of the other comment thread below.

I don't have strong feelings about this, I doubt it'll matter in practice.

matthew-white: > Don't we always expect there to be at least one of these characters?

Or am I wrong about this and there are cases where it's normal for those values to be missing?

matthew-white: Of the two regular expressions / substitutions, the second one more so catches my eye. With `*` instead of `+`, it implies, "if the URL doesn't end with a number, but it does end with a slash, then trim the slash." Which I guess isn't a bad thing, but also might not be the obvious outcome. I'd expect the substitution only to be considered successful if it trimmed a project ID.

test/nginx/mock-sentry/index.js
matthew-white, line 1: Just making a note that I didn't review this file.

test/nginx/nginx.test.docker-compose.yml
matthew-white, line 1: Why are these changes needed for this PR out of curiosity?

matthew-white (approved):
I didn't take a close look at everything in test/, but I saw that there were some good new tests. The changes to files/ make sense to me. :+1: I can take a closer look at test/ later, but I don't feel like that needs to block merge. I think this PR can be merged at any point for inclusion in the upcoming patch release.

test/nginx/nginx.test.docker-compose.yml
alxndrsn: * generating SSL certs is very slow
* the existing `nginx-ssl-selfsign` is explicitly _testing the generation of SSL certs_
* the new tests start new containers repeatedly, but existence of SSL certs is only incidental to those tests

So either the tests can be really slow, or use pre-generated certs.

files/nginx/setup-odk.sh
alxndrsn: I'll add more test cases.

files/nginx/setup-odk.sh
```diff
@@ -45,6 +45,7 @@ echo "writing fresh nginx templates..."
   > /etc/nginx/conf.d/redirector.conf
 
 CERT_DOMAIN=$( [ "$SSL_TYPE" = "customssl" ] && echo "local" || echo "$DOMAIN") \
+SENTRY_DSN_FRONTEND_ROOT="$(awk 'sub(/:\/\/[a-z0-9]*@/, "://") && sub(/\/[0-9]*$/, "")' <<<"$SENTRY_DSN_FRONTEND")" \
```
latin-panda: @alxndrsn what about DSN with a trailing slash? that could easily bite back the system admin

latin-panda (approved):

files/nginx/setup-odk.sh
alxndrsn: :+1: Please suggest a new test case.

files/nginx/setup-odk.sh
alxndrsn: Added.

files/nginx/setup-odk.sh
latin-panda: If an admin accidentally includes a trailing slash in SENTRY_DSN_FRONTEND, for example: `https://.....ingest.sentry.io/1234567/`

alxndrsn: Hmm yes I can see that surprising an admin.  But I don't think it's a valid DSN per https://docs.sentry.io/concepts/key-terms/dsn-explainer/#the-parts-of-the-data-source-name-dsn, so I'd rather just treat it as invalid.

alxndrsn: Would it be helpful somewhere to validate the supplied DSN, and warn if it looks invalid?

latin-panda: OK! If that's already not working in Sentry, agree to a reject and treat it as invalid

PR #2067, opened 2026-07-11
test/nginx: make SENTRY_DSN_FRONTEND value realistic

alxndrsn:
This realism will become more important in future test scenarios and nginx configuration.

- [x] tested locally
- [x] tested in ci

Current test value for `SENTRY_DSN_FRONTEND` is unrealistic, which may affect tested behaviours.

No impact - just a test change.

No.


latin-panda (approved):
Thank you!

PR #2096, opened 2026-07-30
nginx: don't regenerate bad-host SSL cert

alxndrsn:
* consistent with surrounding code: other keys are only generated if they don't exist
* generation is fast, but repeating every startup is still a waste of effort

It originally span out of https://github.com/getodk/central/pull/2066 when I was re-running the tests a lot and trying to speed things up.

Consistent with surrounding code - other keys are only generated if they don't exist.

Speeds up nginx startup a tiny bit.

No.


matthew-white:
This isn't a user-facing change, right? I.e., we don't expect users to be changing/replacing those files? It makes sense to me as a way to speed up tests.

matthew-white (approved):

PR #2110, opened 2026-08-10
chore: split release issue template 

latin-panda:
> [!WARNING]
> Branch off and target `next`, not `master`. The `master` is stable and used in production (exception: documentation/infrastructure-only changes).

Followed the pattern of the other issue templates and double-checked that no steps are missing

Initially, it was Gareth's feedback, but after the recent release, I often read the steps and wondered whether they apply to patch or major releases. I believe this makes it easier to follow with less cognitive load.

No

No

### DIFF
--- a/.github/ISSUE_TEMPLATE/patch_release.md
+++ b/.github/ISSUE_TEMPLATE/patch_release.md
@@ -7,7 +7,7 @@ labels: ''
 assignees: ''
 ---
 
-Complete the steps below to release a patch version of ODK Central.
+Complete the steps below to release a patch version of ODK Central. For a major release, use the **Release** issue template instead.
 
 > **Legend**
 > - 🔎 Requires the second person (reviewer)
--- a/.github/ISSUE_TEMPLATE/patch_release.md
+++ b/.github/ISSUE_TEMPLATE/patch_release.md
@@ -55,14 +55,15 @@ Releasing requires two people: one person to push PRs and complete other tasks a
   - Set as the **latest release**. Don't set as a pre-release.
   - Body: a single line pointing to the upcoming `central` release, e.g., `See release notes at https://github.com/getodk/central/releases/tag/vX.Y.Z`.
 
-### Update submodules
+### Update versions
 
-- [ ] For each of the server and client submodules, `cd` into the directory and run:
+- [ ] If the patch includes `central-backend` changes, `cd server` and run:
   - `git fetch`
   - `git switch -d origin/master` or `git checkout origin/master`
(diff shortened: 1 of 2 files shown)


latin-panda:
@garethbowen @matthew-white 
The PR's base is `master` since it's just a GitHub template change. 

The split was originally suggested by Gareth some weeks ago, and based on my last experience helping with the patch release, I believe it reduces noise during release. I tried to keep the changes to a minimum, just extracting text to the other template.

.github/ISSUE_TEMPLATE/release.md
```diff
@@ -1,42 +1,37 @@
 ---
 name: Release
-about: Checklist for releasing a new version of ODK Central
-title: 'Release vXXXX.X.Y'
+about: Checklist for releasing a new major version of ODK Central
```
garethbowen: Initially I had a comment about how it's major OR minor, but I see our existing language just talks about major vs patch. I think that's ok as long as nobody is confused as to why there's no minor checklist...

.github/ISSUE_TEMPLATE/patch_release.md
```diff
@@ -0,0 +1,121 @@
...
+> Optional section: skip it if no `central-backend` changes are included in the patch.
+
+- [ ] Create a GitHub release on the latest `master` commit of `central-backend`.
+  - Tag: `v*.*.*` (no pre-release suffix).
+  - Set as the **latest release**. Don't set as a pre-release.
+  - Body: a single line pointing to the upcoming `central` release, e.g., `See release notes at https://github.com/getodk/central/releases/tag/vX.Y.Z`.
+
+### Update submodules
+
+- [ ] For each of the server and client submodules, `cd` into the directory and run:
```
garethbowen: There is no client submodule any more, right?

garethbowen (changes requested):

.github/ISSUE_TEMPLATE/release.md
latin-panda: I was wondering the same, but I've heard the Central team call it `major`. 

In this format, what would be considered a major release for a format that has the year (2026.2.0)? Would it be the first release of a given year (2027.1.0)? 😅 

I'm going to remove the word `major` and leave it plain.

.github/ISSUE_TEMPLATE/patch_release.md
latin-panda: I've updated the step

garethbowen (approved):
Nice!

PR #2167, opened 2026-08-27
Upgrade Enketo to 7.6.4

### LINKED ISSUE #1885: enketo forms: OSM map tiles blocked in Firefox?
From https://github.com/enketo/enketo/issues/1512

Likely caused by a combination of:

1. https://wiki.openstreetmap.org/wiki/Blocked_tiles
2. https://github.com/getodk/central/blob/af64f1da31d04ae68b4e01aed45646f88468be7b/files/nginx/common-headers.conf#L9

Could be fixed by:

* https://github.com/Leaflet/Leaflet/issues/10156
* https://github.com/enketo/enketo/issues/1513

Track:

* enketo
  * https://github.com/enketo/enketo/issues/1513
* leaflet
  * https://github.com/Leaflet/Leaflet/pull/9883
  * https://github.com/Leaflet/Leaflet/issues/10156
  * https://github.com/Leaflet/Leaflet/pull/9897

### LINKED ISSUE #2071: Enketo 7.6.2 doesn't load Google Maps tiles
#### Problem description

When loading a geopoint question in Enketo configured with Google Maps, the tiles don't load at all.

@alxndrsn using Firefox reports...
* suspicious log in the browser console: `Empty string passed to getElementById()`
* suspicious error in the debugger: `TypeError: L.GridLayer.GoogleMutant is not a constructor`
* another possible clue in the logs: `Google Maps JavaScript API has been loaded directly without loading=async. This can result in suboptimal performance. For best-practice loading patterns please see https://goo.gle/js-api-loading`

@yanokwa using Chrome reports...
* 7.5.1 works, 7.6.1 works, 7.6.2 and later do not work.
* When Google maps and api config is removed from Enketo, OSM tiles work in 7.6.2 
* When only Google maps config is removed, tiles don't load but we do get a NoApiKeys warning in the console.
* https://github.com/enketo/enketo/compare/7.6.1...7.6.2?diff=unified&w=1 shows that leaflet.gridlayer.googlemutant went from 0.13.5 to 0.16.0 between 7.6.1 and 7.6.2.

lognaturel:
Closes https://github.com/getodk/central/issues/1885 Closes https://github.com/getodk/central/issues/2071

Because the Enketo default map config does not set `referrerPolicy`, we need to add a map config block.

Upgraded to 7.6.4 on dev. Verified that Google Maps works. Verified that OSM with explicit referrerPolicy configuration works. 

While doing this I noticed that OSM tiles don't work with an /x/ "offlineable" form. I see a `NS_ERROR_INTERCEPTION_FAILED` error. This happens with 7.6.1 as well so I don't think we need to do anything about it. Cloud is unaffected because it uses Google Maps.

I considered seeing whether we should add `referrerPolicy` to the default Enketo maps config but I think this is ok.

I looked at the Enketo release notes and associated code changes and I think they're overall low-risk. The dynamic read-only doesn't appear to be complete (e.g. dates don't become read-only) but that's no worse than it was before.

No. Enketo docs have been updated for self-hosters who manage their own config.

### DIFF
--- a/enketo.dockerfile
+++ b/enketo.dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/enketo/enketo:7.6.1
+FROM ghcr.io/enketo/enketo:7.6.4
 
 ENV ENKETO_SRC_DIR=/srv/src/enketo/packages/enketo-express
 WORKDIR ${ENKETO_SRC_DIR}
--- a/files/enketo/config.json.template
+++ b/files/enketo/config.json.template
@@ -34,6 +34,12 @@
     "support": {
         "email": "support@getodk.org"
     },
+    "maps": [ {
+        "name": "street",
+        "tiles": [ "https://tile.openstreetmap.org/{z}/{x}/{y}.png" ],
+        "attribution": "Map data © <a href=\"http://openstreetmap.org\">OpenStreetMap</a> contributors",
+        "referrerPolicy": "origin"
+    } ],
     "text field character limit": 1000000,
     "exclude non-relevant": true,
     "hide powered by": true


lognaturel:
Added @sadiqkhoja in case @alxndrsn doesn't have a chance to review. Would be great to get this on staging asap.

alxndrsn:
Useful comparisons:

* github: https://github.com/enketo/enketo/compare/7.6.1...7.6.4
* git cli (full changes):
```sh
git diff 7.6.1 7.6.4
```
* git cli (commits):
```sh
git log 7.6.1..7.6.4
```

sadiqkhoja (approved):

sadiqkhoja:
Discussed in the meeting to merge this and get this to staging so that this can be test.

PR #2168, opened 2026-08-28
chore: update patch release template with issue checklist

garethbowen:
> [!WARNING]
> Branch off and target `next`, not `master`. The `master` is stable and used in production (exception: documentation/infrastructure-only changes).

Closes #

### DIFF
--- a/.github/ISSUE_TEMPLATE/patch_release.md
+++ b/.github/ISSUE_TEMPLATE/patch_release.md
@@ -20,7 +20,7 @@ Releasing requires two people: one person to push PRs and complete other tasks a
 
 - [ ] Decide the release version (`vXXXX.X.Y`). The same version is used for `central`, `central-backend`, and `central-frontend`, so the `central` release URL (`https://github.com/getodk/central/releases/tag/vXXXX.X.Y`) can be referenced from the `central-frontend` and `central-backend` release bodies.
 - [ ] List all the issues in `central` and `web-forms` repos to be included in this release, and tick them off when they have been cherry-picked into the release branch
-  - [ ] <!-- first issue -->
+  - [ ] {{ first issue }}
 
 ### Get the repository
 
--- a/.github/ISSUE_TEMPLATE/patch_release.md
+++ b/.github/ISSUE_TEMPLATE/patch_release.md
@@ -19,7 +19,8 @@ Releasing requires two people: one person to push PRs and complete other tasks a
 ## Steps
 
 - [ ] Decide the release version (`vXXXX.X.Y`). The same version is used for `central`, `central-backend`, and `central-frontend`, so the `central` release URL (`https://github.com/getodk/central/releases/tag/vXXXX.X.Y`) can be referenced from the `central-frontend` and `central-backend` release bodies.
-- [ ] Write an announcement about the release for the forum.
+- [ ] List all the issues in `central` and `web-forms` repos to be included in this release, and tick them off when they have been cherry-picked into the release branch
+  - [ ] <!-- first issue -->
 
 ### Get the repository
 


.github/ISSUE_TEMPLATE/patch_release.md
```diff
@@ -19,7 +19,8 @@ Releasing requires two people: one person to push PRs and complete other tasks a
 ## Steps
 
 - [ ] Decide the release version (`vXXXX.X.Y`). The same version is used for `central`, `central-backend`, and `central-frontend`, so the `central` release URL (`https://github.com/getodk/central/releases/tag/vXXXX.X.Y`) can be referenced from the `central-frontend` and `central-backend` release bodies.
-- [ ] Write an announcement about the release for the forum.
```
garethbowen, line 22: I don't think you need to do this up front for a patch release...

.github/ISSUE_TEMPLATE/patch_release.md
```diff
@@ -19,7 +19,8 @@ Releasing requires two people: one person to push PRs and complete other tasks a
...
+- [ ] List all the issues in `central` and `web-forms` repos to be included in this release, and tick them off when they have been cherry-picked into the release branch
```
latin-panda, line 22: If it's not obvious, you could also mention that it can be included as a comment in "this issue".

.github/ISSUE_TEMPLATE/patch_release.md
```diff
@@ -19,7 +19,8 @@ Releasing requires two people: one person to push PRs and complete other tasks a
...
+  - [ ] <!-- first issue -->
```
latin-panda: This is not rendering

Image

Maybe wrap it in a code block

latin-panda (approved):
Thank you!

.github/ISSUE_TEMPLATE/patch_release.md
latin-panda: What about moving it to the `Update news` section below

.github/ISSUE_TEMPLATE/patch_release.md
garethbowen: That was by design, so when you're creating/editing the issue you would see it and remove it, but it wouldn't show up if you accidentally left it in. Do you prefer something like `{{ first issue }}`?

.github/ISSUE_TEMPLATE/patch_release.md
garethbowen: I was hoping we'd end up with a single list inline in the issue checklist. The problem with adding them as one or more comments is you have to scroll up and down to make sure everything's included. By including them inline people can keep editing the issue, and then when it comes time to publish the release you have a single list to work through.

.github/ISSUE_TEMPLATE/patch_release.md
latin-panda: Sorry I missed this notification earlier. I think developers won't read the template before creating the issue and will likely miss this `` - [ ] <!-- first issue -->``. Yes, I think this is better:


- [ ]  {{ first issue }}

.github/ISSUE_TEMPLATE/patch_release.md
latin-panda: @garethbowen what do you think about moving this to the "Update news" section below?

.github/ISSUE_TEMPLATE/patch_release.md
garethbowen: Done.

.github/ISSUE_TEMPLATE/patch_release.md
garethbowen: Hah. Sorry I missed _this_ comment! 

We already have a checklist item for "Reply to the topic for the latest release.", which is the same thing right?

latin-panda (approved):
Ready to merge!

Issue #2080, opened 2026-07-21
There is no error message informing user about entering wrong password when downloading submissions from encrypted project


srujner:
#### Problem description
The issue was first reported in the comments below: https://github.com/getodk/central/issues/1011 and it was fixed there so it is a regression. 

There is no error message informing user about entering wrong password when downloading submissions from encrypted project

#### Steps to reproduce the problem
1. Open encrypted project with some submissions in it
2. Navigate to the Submission tab and click on the Download button 
3. Provide wrong password and click on the button next to the **All data and Attachments**
4. The window will automatically close and you'll see message: "Data download should begin soon. Once it begins, you can close this message. If it hasn’t started in 20 seconds, please try again." But nothing happens

#### Expected behavior
There should be error message after providing the wrong password 

#### Central version shown in version.txt

ee0e4207ab6fb628247f4d5b5b576325782a5a05 (v2026.2.0)
 0000000000000000000000000000000000000000 client (v2026.2.0)
 bee93414a2215bd73d7f371113c6640c66d90c95 server (v2026.2.0)

#### Browser version
Firefox 136.0 (64-bit)
Google Chrome 130.0.6723.69

debkarma:
Hi @srujner @matthew-white  I would love to take this up. I have experience with JavaScript frontend development, and since the reproduction steps are so clear, I'd like to investigate why this regression happened. 

Could you assign this to me? I'll start looking into the UI password validation logic. cheers!

matthew-white:
Hi @debkarma, thank you for your interest! This issue is a regression scheduled to be fixed for the next release. Given that it's time-constrained, it's probably not the best first issue. That said, if you have experience with frontend JS and an interest in contributing, we certainly have other small, reproducible issues for which we'd welcome a PR.

Note that we generally don't assign issues outside the ODK team. If you find an issue you'd like to work on, just leave a comment on the issue telling us that you're interested and push a PR once you're ready.

matthew-white:
I took a quick first look at this issue. Let me CC @alxndrsn here because I think it's possible that this is related to recent changes to our CSP. When I follow the repro steps, I see that a request is sent to /csp-report. I also see the following error in the browser console:

> Framing 'https://staging.getodk.cloud/' violates the following Content Security Policy directive: "frame-ancestors 'none'". The request has been blocked.

[...]

PR #2197, opened 2026-09-04
Adjust form-action of Backend CSP

matthew-white:
This PR addresses the latest aspect of #2080.

Nothing much yet. It's not an error for me locally, since I don't believe the development nginx enforces a CSP. I updated a test of the CSP. I'm planning to try the repro steps again on staging once this PR is merged.

I've added thoughts about this approach at https://github.com/getodk/central/issues/2080#issuecomment-5536795517. Ultimately, I would prefer to adjust the implementation in Frontend; I can file an issue about that. However, that'll be a little harder than the one-line change here. I don't think it's important enough to spend additional cycles on it before regression testing.

Users can start submitting HTML forms from their Backend endpoints like we are? Hopefully users don't need to take such an approach. 😅

### DIFF
--- a/files/nginx/odk.conf.template
+++ b/files/nginx/odk.conf.template
@@ -110,7 +110,7 @@ map $spa_name $spa_csp {
 
 map $upstream_http_content_security_policy $central_backend_csp {
   # pass through any Content-Security-Policy received from upstream services (central-backend, enketo)
-  ""      "default-src 'report-sample' 'none'; form-action 'none'; frame-ancestors 'self'; img-src http://${DOMAIN}/favicon.ico; report-uri /csp-report";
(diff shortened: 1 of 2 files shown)


matthew-white:
> How does this change impact users?

We're basically reverting the CSP's enforcement of form submission behavior to what it was a couple of releases ago, before the CSP enforced it as strictly.

lognaturel (approved):
I've reviewed the thread and agree this is a good temporary fix.

If I understand correctly, in the next release we will:
* have frontend go back to /blank.html as soon as it's retrieved the Backend response from document.body. (https://github.com/getodk/central/issues/2080#issuecomment-5536795517)
* revert this change
* add e2e testing about the download of encrypted submission data

Issue #2203, opened 2026-09-06
[central] with-pgenvblock.sh ignores env vars containing linebreaks


alxndrsn:
Noted while extracting tests from https://github.com/getodk/central/pull/2119

PR #2119, opened 2026-08-11
service: support NODE_EXTRA_CA_CERTS in cron jobs

alxndrsn:
### Alternative to

* https://github.com/getodk/central-backend/pull/1910

https://github.com/getodk/central-backend/pull/1910 currently preferred over this PR.

---

Closes https://github.com/getodk/central/issues/2115

#### What has been done to verify that this works as intended?

- [x] new test
- [x] passing CI

#### Why is this the best possible solution? Were any other approaches considered?

It's not - https://github.com/getodk/central-backend/pull/1910 is probably neater.

#### How does this change impact users? Describe intentional behavior changes from code updates. What are the regression risks?

Make cron jobs work again.

#### Does this change require updates to documentation? If so, please file an issue [here](https://github.com/getodk/docs/issues/new) and include the link below.

No.

Issue #1793, opened 2026-04-10
Pass more environment variables to cron jobs


matthew-white:
In #1747, we saw that the environment variables specified in docker-compose.yml aren't passed to cron jobs. For #1747, we decided just to pass `PG*` variables, but we probably want to pass all variables. This issue is to do so.

Example: `NODE_OPTIONS` would probably be a good thing to pass to cron jobs.

- [ ] Check the list of environment variables in docker-compose.yml. Do we want to pass them all to cron jobs?
- [ ] Pass all appropriate variables (probably all variables) to cron jobs.

matthew-white:
CC @brontolosone

lognaturel:
We've gone back and forth on this. My recollection is that we're currently leaning toward continuing to pass only the variables that are needed or meaningful for cron jobs.

We initially considered passing everything because that's the behavior we previously expected. But upon reflection, passing seemingly unrelated variables could have unintended effects that we haven't considered.

PR #2204, opened 2026-09-06
test/with-pgenvblock: add additional test

alxndrsn:
# Example outputs

### Passing

```sh
$ ./test/test-with-pgenvblock.sh 
Checking whether with-pgenvblock.pl works… Yes
```

### Failing

```sh
$ ./test/test-with-pgenvblock.sh 
Checking whether with-pgenvblock.pl works… No, it doesn't
```

### Passing

```sh
$ ./test/test-with-pgenvblock.sh 
[test-with-pgenvblock] Testing...
[test-with-pgenvblock]   Testing: specific postgres-related variables...
[test-with-pgenvblock]     Passed OK.
[test-with-pgenvblock]   Testing: local env with injected var...
[test-with-pgenvblock]     Passed OK.
[test-with-pgenvblock] All tests passed OK.
```

### Failing first test

```sh
$ ./test/test-with-pgenvblock.sh 
[test-with-pgenvblock] Testing...
[test-with-pgenvblock]   Testing: specific postgres-related variables...
2c2
< PGDATBASE=4
---
> PGDATABASE=4
[test-with-pgenvblock] !!!
[test-with-pgenvblock] !!! Test failed; see above for differences between '< actual' and '> expected'
[test-with-pgenvblock] !!!
```

### Failing second test

```sh
$ ./test/test-with-pgenvblock.sh 
[test-with-pgenvblock] Testing...
[test-with-pgenvblock]   Testing: specific postgres-related variables...
[test-with-pgenvblock]     Passed OK.
[test-with-pgenvblock]   Testing: local env with injected var...
[test-with-pgenvblock]     !!! Test failed.
```

---

Extracted from discarded PR https://github.com/getodk/central/pull/2119

Should help with https://github.com/getodk/central/issues/1793

It's a new test - the test was run.

Also it allowed identification of https://github.com/getodk/central/issues/2203.

The logging adds clarity around execution of the existing test, and adds an additional test which makes specific test cases easier to add.

No change - just tests.

No.

### DIFF
--- a/test/test-with-pgenvblock.sh
+++ b/test/test-with-pgenvblock.sh
@@ -34,7 +34,7 @@ log "  Testing: local env with injected var..."
 # note that the environment setting (with `export`) runs in a subshell, and as such doesn't touch our own environment,
 # and as such thus also not the environment with-pgenvblock.pl's is launched with; and thus the PGBLA environment variable
 # that the `env` invocation sees comes from `with-pgenvblock.pl`'s reading of the env block file and nowhere else.
-if files/service/with-pgenvblock.pl <(export PGBLA=hurray; cat /proc/self/environ) env | grep --quiet '^PGBLA=huray$'; then
+if files/service/with-pgenvblock.pl <(export PGBLA=hurray; cat /proc/self/environ) env | grep --quiet '^PGBLA=hurray$'; then
   log "    Passed OK."
 else
   log "    !!! Test failed."
--- a/test/test-with-pgenvblock.sh
+++ b/test/test-with-pgenvblock.sh
@@ -6,15 +6,7 @@ log() { echo >&2 "[test-with-pgenvblock] $*"; }
 
 log "Testing..."
 
-log "Testing: local env with injected var..."
-# for the skeptical reader, on the below:
-# note that the environment setting (with `export`) runs in a subshell, and as such doesn't touch our own environment,
-# and as such thus also not the environment with-pgenvblock.pl's is launched with; and thus the PGBLA environment variable
-# that the `env` invocation sees comes from `with-pgenvblock.pl`'s reading of the env block file and nowhere else.
-files/service/with-pgenvblock.pl <(export PGBLA=hurray; cat /proc/self/environ) env | grep --quiet '^PGBLA=hurray$' || (printf >&2 "No, it doesn't\n"; false) && printf >&2 "Yes\n"
-log "  Passed OK."
-
-log "Testing: specific postgres-related variables..."
+log "  Testing: specific postgres-related variables..."
 if diff \
     <(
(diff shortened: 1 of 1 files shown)


matthew-white:
I went ahead and tagged @lognaturel, since IIRC she reviewed the original version of this code.

alxndrsn:
It looks like the old test relied on being the last-executed line, so no longer fails the script if the test fails.

test/test-with-pgenvblock.sh
```diff
@@ -2,9 +2,43 @@
...
+  log "!!!"
+  exit 1
+fi
+
+log "  Testing: local env + an injected var..."
 # for the skeptical reader, on the below:
 # note that the environment setting (with `export`) runs in a subshell, and as such doesn't touch our own environment,
 # and as such thus also not the environment with-pgenvblock.pl's is launched with; and thus the PGBLA environment variable
 # that the `env` invocation sees comes from `with-pgenvblock.pl`'s reading of the env block file and nowhere else.
-files/service/with-pgenvblock.pl <(export PGBLA=hurray; cat /proc/self/environ) env | grep --quiet '^PGBLA=hurray$' || (printf >&2 "No, it doesn't\n"; false) && printf >&2 "Yes\n"
```
alxndrsn, line 10: This structure had to change because it wasn't triggering `set -e` handling due to its structure:

> The shell does not exit if the command that fails is part of ... any command executed in a `&&` or `||` list except the command following the final `&&` or `||`
> -_https://doc.guix.gnu.org/bash/latest/en/html_node/The-Set-Builtin.html_

test/test-with-pgenvblock.sh
```diff
@@ -2,9 +2,43 @@
...
+log() { echo >&2 "[test-with-pgenvblock] $*"; }
+
+log "Testing..."
+
+log "  Testing: specific postgres-related variables..."
+if diff \
+    <(
+      env --ignore-environment \
+          files/service/with-pgenvblock.pl \
+          <(printf "A=1\0PGSSLMODE=2\0C=3\0PGDATABASE=4\0E=5\0NODE_EXTRA_CA_CERTS=5\0\G=7") \
```
lognaturel: Looks like `\0\G` should just be `\0G`

lognaturel (approved):

test/test-with-pgenvblock.sh
alxndrsn: Good catch - thanks!

alxndrsn:
```suggestion
          <(printf "A=1\0PGSSLMODE=2\0C=3\0PGDATABASE=4\0E=5\0NODE_EXTRA_CA_CERTS=5\0G=7") \
```

PR #2215, opened 2026-09-07
upgrade-postgres: clarify data cleanup link

alxndrsn:
In the original implementation (https://github.com/getodk/central/pull/343), old postgres data was cleaned up automatically if a specific flag file was present.

This behaviour was changed (prior to release?) in a follow-up PR (https://github.com/getodk/central/pull/395) to link to documentation instead.

Over time, it's become less clear which part of the linked documentation is relevant.  This PR aims to link more directly to the relevant documentation.

- [x] clicked the link
- [x] ci
- [x] not much else

It took me a while to understand which part of the original link (https://docs.getodk.org/central-upgrade/) was relevant to upgrading postgres.  The new link is https://docs.getodk.org/central-upgrade/#upgrading-to-central-v2023-2

Hopefully gets them where they need to be quicker.  Low risk of regression.

No.

### DIFF
--- a/files/postgres/upgrade-postgres.sh
+++ b/files/postgres/upgrade-postgres.sh
@@ -28,7 +28,7 @@ if [[ -f "$flag_upgradeCompletedOk" ]]; then
     log "!!!"
     log "!!! This is taking up disk space: $(du -hs "$PGDATAOLD" 2>/dev/null | cut -f1)B"
     log "!!!"
-    log "!!! Continue with the instructions at https://docs.getodk.org/central-upgrade/"
+    log "!!! Continue with the instructions at https://docs.getodk.org/central-upgrade/#upgrading-to-central-v2023-2
     log "!!!"
   fi
 else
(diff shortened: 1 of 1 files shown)


files/postgres/upgrade-postgres.sh
```diff
@@ -28,7 +28,7 @@ if [[ -f "$flag_upgradeCompletedOk" ]]; then
     log "!!!"
     log "!!! This is taking up disk space: $(du -hs "$PGDATAOLD" 2>/dev/null | cut -f1)B"
     log "!!!"
-    log "!!! Continue with the instructions at https://docs.getodk.org/central-upgrade/"
+    log "!!! Continue with the instructions at https://docs.getodk.org/central-upgrade/#upgrading-to-central-v2023-2
```
alxndrsn:
```suggestion
    log "!!! Continue with the instructions at https://docs.getodk.org/central-upgrade/#upgrading-to-central-v2023-2"
```

lognaturel (approved):

PR #2228, opened 2026-09-11
updated server submodule and FRONTEND_VERSION

sadiqkhoja:


matthew-white (approved):

matthew-white:
It looks like check-docker-context.sh failed. I wonder if we should the bump the `--max-count`

```
[check-docker-context.sh] File count:  709
[check-docker-context.sh] Total size: 8.516 MB
[check-docker-context.sh] !!!
[check-docker-context.sh] !!! This is a surprising number of files - expected between 600 and 700 !!!
```

PR #2229, opened 2026-09-11
Increased expected file count to 650 to 750

sadiqkhoja:


matthew-white:
CC @alxndrsn in case this is of interest. The file count increased from 695 to 709 in #2228. We added new Frontend components in this release, so I'm assuming it's due to that.

matthew-white (approved):

Issue #1385, opened 2025-09-25
Update postgres


lognaturel:
Postgres 14 goes out of LTS Nov 12, 2026. I don't think this will be a hugely involved update but we should schedule it ahead of time.

We may also want to introduce postgis for geo-aware filtering at that point.

alxndrsn:
I'd aim as early as possible - ideally there's time for:

* testing
* self-hosted rollout before v14 support ends
* any surprises

Are we aiming for PostgreSQL v18?

* it's currently the latest version
* supported into 2030: https://www.postgresql.org/support/versioning/
* available on RDS: https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-versions.html#postgresql-version18
* v19 won't be released until late 2026, which is cutting it close

From https://www.postgresql.org/support/versioning/:

> You can upgrade from one major version to another without upgrading to intervening versions, but we recommend reading the [release notes](https://www.postgresql.org/docs/release/) of all intervening major versions prior to doing so.

This would mean reviewing:

* [v15 release notes](https://www.postgresql.org/docs/release/15.0/)
* [v16 release notes](https://www.postgresql.org/docs/release/16.0/)
* [v17 release notes](https://www.postgresql.org/docs/release/17.0/)
* [v18 release notes](https://www.postgresql.org/docs/release/18.0/)

If there's any reason to aim for an older version than v18, this should become clear during release note review & testing.

lognaturel:
@alxndrsn can we put this on your plate for .1? I agree with your process.

matthew-white:
Notes from team meeting today. @brontolosone and @lognaturel think that it'd be reasonable to upgrade directly to v19. It's acceptable for there to be a short period of time in which our version of Postgres in use is out of LTS. The postgres container should not be exposed to the universe, so that limits the surface area.

lognaturel:
Thanks for the recommendation, @spwoodcock!

PR #2248, opened 2026-09-15
docker-compose: add explicit dependency between postgres containers

alxndrsn:
This PR contains work foundational to:

* https://github.com/getodk/central/issues/1385

-----

This seems to prevent a race condition related to volume creation order:

	Container central-postgres14-1  Error response from daemon: failed to mkdir /var/lib/docker/volumes/central_postgres14/_data/data: mkdir /var/lib/docker/volumes/central_postgres14/_data/data: file exists

This error was seen ~3% of the time when running:

* `postgres` container based on `tianon/postgres-upgrade:9.6-to-14`
* `postgres14` container based on `pgautoupgrade/pgautoupgrade:17.11-trixie`

Given the vague nature of both the error itself, and reproduction circumstances, it seems prudent to introduce this dependency.

The dependency itself seems logical, as the `postgres14` container should be waiting for the `postgres` container to create a flag file before PostgreSQL itself starts up.

See:

* https://github.com/moby/moby/issues/47021
* https://forums.docker.com/t/error-while-creating-mount-source-path-after-docker-desktop-update-to-v4-35/144676

- [x] ran CI 200 times and didn't see the error repeat

Could just ignore it, and hope it never happens to a self-hoster.

Should prevent container failure.  It's unknown if the failure is temporary, or permanent.

#### Is this change user-facing or otherwise noteworthy to users? If so, please add an entry for it in CHANGELOG.md.

There's no evidence a user has ever experienced this, so it's unclear if it's helpful to add to the CHANGELOG.

Probably not.


.github/workflows/main.yml
```diff
@@ -77,6 +77,7 @@ jobs:
       with:
         node-version-file: test/package.json
     - run: cd test/nginx && npm clean-install --no-audit
+    - run: touch ./files/allow-postgres14-upgrade
```
alxndrsn, line 80: This is now required as starting the `service` container triggers starting the `postgres14` container, which now triggers starting this `postgres` container.

lognaturel (approved):

